PR #42, opened 2018-09-06
minor fixes to commenting

vdubya:


ikeough:
Thanks for following along @vdubya. I'm closing this because I've already alerted @hummingburd that closing bracket comments won't fly in this code base. They will be removed. We pulled in a bunch of his stuff that still requires some refactoring.

PR #58, opened 2018-09-12
Added color constants and new Color constructor to use them

hummingburd:


csharp/src/sdk/Geometry/Color.cs
```diff
@@ -6,7 +6,35 @@ namespace Hypar.Geometry
     /// Color represents an RGBA color.
     /// </summary>
     public class Color
-    {   
+    {
+        public static readonly float[] Aqua = { 0.3f, 0.7f, 0.7f, 0.0f };
```
ikeough: I'd like to do this how `System.Drawing.Color` works in .net. Convert all of these to static properties like:
```c#
public static Color Red
{
  get{ return new Color(1.0f, 0.0f, 0.0f, 0.0f);}
}
```
The difference would be this:
as proposed:
```
var red = new Color(Color.Red);
```
updated:
```
var red = Color.Red
```

csharp/src/sdk/Geometry/Color.cs
```diff
@@ -56,6 +84,33 @@ public Color(float red, float green, float blue, float alpha)
             this.Alpha = alpha;
         }
 
+        /// <summary>
+        /// Construct a color from an red, green, blue, and alpha value array.
+        /// </summary>
+        /// <param name="rgba">An array of the red, blue, green, and alpha color components between 0.0 and 1.0.</param>
+        public Color(float[] rgba)
```
ikeough: Pending the changes above, this constructor is no longer needed.

csharp/src/sdk/Geometry/Color.cs
hummingburd: Ah okay. That's more elegant, thanks.

csharp/src/sdk/Geometry/Color.cs
hummingburd: Right, saw that, deleted.

PR #59, opened 2018-09-12
Added initial Space object

hummingburd:
Configured Space as a Polyline and a Height.
Added a new Message to notify of Height <= 0.
Added relevant tests, a subset of the Mass tests.

### DIFF
--- a/csharp/src/sdk/Elements/Space.cs
+++ b/csharp/src/sdk/Elements/Space.cs
@@ -69,6 +69,7 @@ namespace Hypar.Elements
 
             this.Location = bottom;
             this.BottomElevation = bottomElevation;
+            this.Height = height;
             this.Material = BuiltInMaterials.Default;
         }
 
--- a/csharp/src/sdk/Elements/Space.cs
+++ b/csharp/src/sdk/Elements/Space.cs
@@ -8,157 +8,103 @@ using Hypar.Geometry;
 namespace Hypar.Elements
 {
     /// <summary>
-    /// A space represents an extruded boundary an occupiable region within a building.
+    /// A space represents the extruded boundary of an occupiable region.
     /// </summary>
-    public class Space : Element, ILocateable<Polyline>, ITessellate<Mesh>, IMaterialize
+    public class Space : Element, ILocateable<Polyline>, ITessellate<Mesh>, ITransformable, IMaterialize
     {
-        private List<Polyline> _sides = new List<Polyline>();
+        /// <summary>
+        /// The elevation of the lower Space perimeter.
+        /// </summary>
+        [JsonProperty("elevation")]
+        public double Elevation { get; }
 
         /// <summary>
-        /// The bottom perimeter of the space.
+        /// The height of the Space above its perimeter elevation.
+        /// </summary>
+        [JsonProperty("height")]
+        public double Height { get; }
+
+        /// <summary>
+        /// The lower perimeter of the Space.
         /// </summary>
         /// <returns></returns>
         [JsonProperty("location")]
         public Polyline Location{get;}
 
         /// <summary>
-        /// The elevation of the bottom perimeter.
-        /// </summary>
-        [JsonProperty("bottom_elevation")]
-        public double BottomElevation{get;}
-
-        /// <summary>
-        /// The height of the space above its bottom elevation.
-        /// </summary>
-        [JsonProperty("height")]
-        public double Height{get;}
-
-        /// <summary>
-        /// The material of the Mass.
+        /// The material of the Space.
         /// </summary>
         /// <value></value>
         [JsonIgnore]
-        public Material Material{get;set;}
+        public Material Material { get; set; }
 
         /// <summary>
-        /// Construct a default space.
+        /// The transform of the Space.
+        /// </summary>
+        /// <value></value>
+        [JsonProperty("transform")]
+        public Transform Transform { get; }
+
+        /// <summary>
+        /// The voids within the Space.
+        /// </summary>
+        /// <value></value>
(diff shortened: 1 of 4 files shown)


csharp/src/sdk/Elements/Space.cs
```diff
@@ -0,0 +1,163 @@
...
+            this.Location = bottom;
+            this.BottomElevation = bottomElevation;
+            this.Material = BuiltInMaterials.Default;
+        }
+
+        /// <summary>
+        /// A collection of curves representing the vertical edges of the space.
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<Line> VerticalEdges()
```
ikeough: Is this functionality we need right now? Same question for `Faces()` and `VerticalEdges()`. If not, remove these.

csharp/src/sdk/Elements/Space.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Hypar.Geometry;
+
+namespace Hypar.Elements
+{
+    /// <summary>
+    /// A space represents an extruded boundary an occupiable region within a building.
+    /// </summary>
+    public class Space : Element, ILocateable<Polyline>, ITessellate<Mesh>, IMaterialize
```
ikeough: I'd like to make this `ITransformable`, and create a `Transform` with its origin at the elevation set for `BottomElevation`. You can see [how this is done](https://github.com/hypar-io/sdk/blob/master/csharp/src/sdk/Elements/Floor.cs#L81) in `Floor`.

csharp/src/sdk/Elements/Space.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <returns>A mesh representing the tessellated mass.</returns>
+        public Mesh Tessellate()
+        {
+            var mesh = new Mesh();
+            foreach (var s in Faces())
+            {
+                mesh.AddQuad(s.ToArray());
+            }
+
+            mesh.AddTesselatedFace(new[] { this.Location }, this.BottomElevation);
```
ikeough: If `Space` becomes `ITransformable`, the you won't need to use `this.BottomElevation`. Everything will be made as if it's sitting on the XY plane, then the transform on the element will be saved to the glTF so that the rendering system knows how to properly transform this element.

csharp/src/sdk/Elements/Space.cs
```diff
@@ -0,0 +1,163 @@
...
+            mesh.AddTesselatedFace(new[] { this.Location }, this.BottomElevation + this.Height, true);
```
ikeough: Following on the comment above, this would just be `mesh.AddTesselatedFace(new[] { this.Location }, this.Height, true);`

ikeough:
The transform stuff seems like a pain in the neck, and perhaps not the most straightforward thing, but it will pay dividends in the future as we will be able to make elements sub-elements of other elements and the renderer will be able to use a transform stack to properly display everything.

csharp/src/sdk/Elements/Space.cs
hummingburd: Nope, left them in just in case you thought they should be there.

csharp/src/sdk/Elements/Space.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        /// <summary>
+        /// Construct a space from a perimeter and a height.
+        /// </summary>
+        /// <param name="bottom">The bottom perimeter of the space.</param>
+        /// <param name="bottomElevation">The elevation of the bottom perimeter.</param>
+        /// <param name="height">The height of the space above the bottom elevation.</param>
+        /// <returns></returns>
+        public Space(Polyline bottom, double bottomElevation, double height)
```
ikeough: Given that we talked about having `voids`, this should be:
`public Space(Polyline perimeter IEnumerable<Polyline> voids, double elevation, double height)`

Without the `bottom` input, the `bottomElevation` makes less sense.

csharp/src/sdk/Elements/Space.cs
hummingburd: Okay, I'll look at that.

csharp/src/sdk/Elements/Space.cs
hummingburd: Oh right, nice.

csharp/src/sdk/Elements/Space.cs
hummingburd: Right.

csharp/src/sdk/Elements/Space.cs
hummingburd: Removed HorizonalEdges, VerticalEdges, and Faces.

ikeough:
This was dumb of me. I hadn’t seen that you implemented this in the Messages.cs. The github review interface sometimes tricks you into thinking you’re in the same file as the previous code chunk. Ignore these comments. Leave it as is.

csharp/src/sdk/Elements/Space.cs
hummingburd: Made this change. Added a TODO comment to test that voids are within the perimeter. If we implement the Geometry operations list I supplied, we could use those functions to confirm acceptable voids.

csharp/src/sdk/Elements/Space.cs
hummingburd: Used the way Floor handles it instead.

csharp/src/sdk/Elements/Space.cs
hummingburd: I think I did this right following the example in floor.

PR #63, opened 2018-09-18
Added Clipper functionality and derivative Polygon comparisons to PolygonExtensions

hummingburd:
Added assembly visibility of internals to Hypar.SDK.Tests to allow unit testing and added tests.


csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -200,10 +210,153 @@ internal static List<IntPoint> ToClipperPath(this Polygon p)
...
             var scale = 1024.0;
             return new Polygon(p.Select(v=>new Vector3(v.X/scale, v.Y/scale)));
         }
+
+        /// <summary>
+        /// Returns true if the interior of the first polygon is coincident with every interior and boundary point of the second polygon.
+        /// </summary>
+        /// <param name="p1">The polygon boundary.</param>
+        /// <param name="p2">The polygon tested for containment within the boundary.</param>
+        /// <param name="cvr">If true, polygon is contained if boundary points are coincident.</param>
```
ikeough: Where possible use `Perimeter` in place of `Boundary` and `Vertex` or `Vertices` in place of `Point`. This is the nomenclature used elsewhere in `Polygon`.

csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -322,41 +224,190 @@ internal static bool Touches(Polygon p1, Polygon p2)
         }
 
         /// <summary>
-        /// Performs difference,intersection, union, or XOR operations on two polygons.
+        /// Returns a list of Polygons representing the difference between this Polygon and the supplied Polygon.
```
ikeough: See my comment above on using the `<returns>` tag. In this instance, you'd also want the returns tag to specify what happens when there is no difference result. Do we get an empty list, null, does it throw an exception? In this case, the returns would say: "Returns a collection of Polygons representing the difference between this Polygon and the supplied polygon. If the provided Polygons do not overlap, this returns an empty collection."

ikeough: The returns tag can also be valuable to hint at whether the method changes the internal state of something or returns a new thing. This doesn't apply in this instance, but may in others.

csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -322,41 +224,190 @@ internal static bool Touches(Polygon p1, Polygon p2)
...
+                polygons.Add(PolygonExtensions.ToPolygon(path));
             }
             return polygons;
         }
+
+        /// <summary>
+        /// Offset this polyline by the specified amount.
+        /// </summary>
+        /// <param name="offset">The amount to offset.</param>
+        /// <returns>A new polyline offset by offset.</returns>
```
ikeough: I never updated this to Polygon, but see what this does? `A new polygon..` It tells us we're getting a new Polygon *not* just updating the state of the target.

csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -161,81 +111,36 @@ public Vector3 PointAt(double u)
...
-        /// <param name="p"></param>
+        /// <param name="polygon">The Polygon to compare.</param>
         /// <returns></returns>
-        internal static List<IntPoint> ToClipperPath(this Polygon p)
+        public bool Contains(Polygon polygon)
         {
-            var scale = 1024.0;
-            var path = new List<IntPoint>();
-            foreach(var v in p.Vertices)
+            var thisPath = PolygonExtensions.ToClipperPath(this);
```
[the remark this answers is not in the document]
hummingburd: Oh. Duh. Okay.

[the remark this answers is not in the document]
hummingburd: Seems like if I want to send the supplied Polygon ToClipperPath I need to have the PolygonExtensions. prefix. Otherwise VS tells me there's no overload that accepts one argument.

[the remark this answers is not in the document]
hummingburd: The this. syntax works fine for the Polygon itself.

csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -130,14 +131,16 @@ public bool Contains(Polygon polygon)
...
-        /// <returns></returns>
+        /// <returns>
+        /// Returns true if every vertex of the supplied Polygon falls within or on the perimeter of this Polygon.
+        /// </returns>
         public bool Covers(Polygon polygon)
         {
-            var thisPath = PolygonExtensions.ToClipperPath(this);
-            var polyPath = PolygonExtensions.ToClipperPath(polygon);
+            var thisPath = this.ToClipperPath();
+            var polyPath = polygon.ToClipperPath();
```
ikeough: Nice!

csharp/src/sdk/Geometry/Polygon.cs
```diff
@@ -268,36 +283,41 @@ public IEnumerable<Polygon> Intersection(Polygon polygon)
...
             Clipper clipper = new Clipper();
             clipper.AddPath(thisPath, PolyType.ptSubject, true);
             clipper.AddPath(polyPath, PolyType.ptClip, true);
             var solution = new List<List<IntPoint>>();
             clipper.Execute(ClipType.ctUnion, solution);
-            var polygons = new List<Polygon>();
-            foreach (List<IntPoint> path in solution)
+            if (solution.Count > 1)
```
ikeough: Is this really true? Does clipper returns a list with the inputs if the union fails?

hummingburd: Yeah, it surprised me too. I sent it two disjoint Polygons through my unit test and got 2 Polygons back.

hummingburd: There was a clue in the Clipper docs that made me think I'd better test it.

PR #81, opened 2018-10-05
Added multiple Polygon inputs for Polygon Difference and Union operations

hummingburd:


hummingburd:
Also added new PointWithin method and associated test to Polygon

hummingburd:
Too many changes since, will make a different pull request that is more granular.

PR #90, opened 2018-11-29
Polygon Fixes and Enhancements

hummingburd:
1) Fixed false positives on "Touches" when Polygons actually intersect.
2) Added methods for relationships between Vector3 point(s) and Polygons.
3) Multiple Polygon and Vector3 relationship methods are freestanding from single entity methods.
4) Simplified "Intersects" method.
5) Added Distinct operator to returned Clipper list on "Union" operation to avoid sending duplicate points to "ToPolygon".
6) Updated test suite to cover new methods.

### DIFF
--- a/csharp/src/elements/Geometry/Polygon.cs
+++ b/csharp/src/elements/Geometry/Polygon.cs
@@ -73,20 +73,20 @@ namespace Hypar.Geometry
         public Polygon(IList<Vector3> vertices) : base(vertices){}
 
         /// <summary>
-        /// Tests if the supplied Vector3 point is within this Polygon without coincidence with an edge when compared on a shared plane.
+        /// Tests if the supplied Vector3 is within this Polygon without coincidence with an edge when compared on a shared plane.
         /// </summary>
-        /// <param name="point">The Vector3 point to compare to this Polygon.</param>
+        /// <param name="vector">The Vector3 to compare to this Polygon.</param>
         /// <returns>
-        /// Returns true if the supplied Vector3 point is within this Polygon when compared on a shared plane. Returns false if the Vector3 point is outside this Polygon or if the supplied Vector3 point is null.
+        /// Returns true if the supplied Vector3 is within this Polygon when compared on a shared plane. Returns false if the Vector3 is outside this Polygon or if the supplied Vector3 is null.
         /// </returns>
-        public bool Contains(Vector3 point)
+        public bool Contains(Vector3 vector)
         {
-            if (point == null)
+            if (vector == null)
             {
                 return false;
             }
             var thisPath = this.ToClipperPath();
-            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
+            var intPoint = new IntPoint(vector.X * scale, vector.Y * scale);
             if (Clipper.PointInPolygon(intPoint, thisPath) != 1)
             {
                 return false;
@@ -120,20 +120,20 @@ namespace Hypar.Geometry
         }
 
         /// <summary>
-        /// Tests if the supplied Vector3 point is within this Polygon or coincident with an edge when compared on a shared plane.
+        /// Tests if the supplied Vector3 is within this Polygon or coincident with an edge when compared on a shared plane.
         /// </summary>
-        /// <param name="point">The Vector3 point to compare to this Polygon.</param>
+        /// <param name="vector">The Vector3 to compare to this Polygon.</param>
         /// <returns>
-        /// Returns true if the supplied Vector3 point is within this Polygon or coincident with an edge when compared on a shared plane. Returns false if the supplied point is outside this Polygon, or if the supplied Vector3 point is null.
+        /// Returns true if the supplied Vector3 is within this Polygon or coincident with an edge when compared on a shared plane. Returns false if the supplied Vector3 is outside this Polygon, or if the supplied Vector3 is null.
         /// </returns>
-        public bool Covers(Vector3 point)
+        public bool Covers(Vector3 vector)
         {
-            if (point == null)
+            if (vector == null)
             {
                 return false;
             }
             var thisPath = this.ToClipperPath();
-            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
+            var intPoint = new IntPoint(vector.X * scale, vector.Y * scale);
             if (Clipper.PointInPolygon(intPoint, thisPath) == 0)
             {
                 return false;
@@ -171,21 +171,21 @@ namespace Hypar.Geometry
             return true;
         }
 
-         /// <summary>
-        /// Tests if the supplied Vector3 point is outside this Polygon when compared on a shared plane.
+        /// <summary>
+        /// Tests if the supplied Vector3 is outside this Polygon when compared on a shared plane.
         /// </summary>
-        /// <param name="point">The Vector3 point to compare to this Polygon.</param>
+        /// <param name="vector">The Vector3 to compare to this Polygon.</param>
         /// <returns>
-        /// Returns true if the supplied Vector3 point is outside this Polygon when compared on a shared plane or if the supplied Vector3 point is null.
+        /// Returns true if the supplied Vector3 is outside this Polygon when compared on a shared plane or if the supplied Vector3 is null.
         /// </returns>
-        public bool Disjoint(Vector3 point)
+        public bool Disjoint(Vector3 vector)
         {
-            if (point == null)
+            if (vector == null)
             {
                 return true;
             }
             var thisPath = this.ToClipperPath();
-            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
+            var intPoint = new IntPoint(vector.X * scale, vector.Y * scale);
             if (Clipper.PointInPolygon(intPoint, thisPath) != 0)
             {
                 return false;
@@ -247,20 +247,20 @@ namespace Hypar.Geometry
         }
 
         /// <summary>
-        /// Tests if the supplied Vector3 point is coincident with an edge of this Polygon when compared on a shared plane.
+        /// Tests if the supplied Vector3 is coincident with an edge of this Polygon when compared on a shared plane.
         /// </summary>
-        /// <param name="point">The Vector3 point to compare to this Polygon.</param>
+        /// <param name="vector">The Vector3 to compare to this Polygon.</param>
         /// <returns>
-        /// Returns true if the supplied Vector3 point coincides with an edge of this Polygon when compared on a shared plane. Returns false if the supplied Vector3 point is not coincident with an edge of this Polygon, or if the supplied Vector3 point is null.
+        /// Returns true if the supplied Vector3 coincides with an edge of this Polygon when compared on a shared plane. Returns false if the supplied Vector3 is not coincident with an edge of this Polygon, or if the supplied Vector3 is null.
         /// </returns>
-        public bool Touches(Vector3 point)
+        public bool Touches(Vector3 vector)
         {
-            if (point == null)
+            if (vector == null)
             {
                 return false;
             }
             var thisPath = this.ToClipperPath();
-            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
+            var intPoint = new IntPoint(vector.X * scale, vector.Y * scale);
             if (Clipper.PointInPolygon(intPoint, thisPath) != -1)
             {
                 return false;
--- a/csharp/src/elements/Geometry/Polygon.cs
+++ b/csharp/src/elements/Geometry/Polygon.cs
@@ -95,29 +95,6 @@ namespace Hypar.Geometry
         }
 
         /// <summary>
-        /// Tests if all the supplied Vector3 points are within this Polygon without coincidence with an edge when compared on a shared plane.
-        /// </summary>
-        /// <param name="points">The list of Vector3 points to compare to this Polygon.</param>
-        /// <returns>
-        /// Returns true if all the supplied Vector3 points are within this Polygon when compared on a shared plane. Returns false if any of the Vector3 points are outside this Polygon or if the supplied Vector3 point list is null.
-        /// </returns>
-        public bool Contains(IList<Vector3> points)
-        {
-            if (points == null)
-            {
-                return false;
-            }
-            foreach (Vector3 point in points)
-            {
-                if (!this.Contains(point))
-                {
-                    return false;
-                }
-            }
-            return true;
-        }
-
-        /// <summary>
         /// Tests if the supplied Polygon is within this Polygon without coincident edges when compared on a shared plane.
         /// </summary>
         /// <param name="polygon">The Polygon to compare to this Polygon.</param>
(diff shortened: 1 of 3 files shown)


csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -91,19 +132,163 @@ public bool Contains(Polygon polygon)
...
+                    {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Tests if the supplied 2D Vector3 point is within or touches the perimeter of this Polygon.
```
ikeough: There's no such thing as a "2D Vector3". Do you mean to say "Tests whether a Vector3 in the same plane as the Polygon is within or touches..."?

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -69,6 +69,47 @@ public Vector3 Centroid
...
+        }
+
+        /// <summary>
+        /// Tests if all the supplied 2D Vector3 points fall within the perimeter of this Polygon.
+        /// </summary>
+        /// <param name="points">The collection of 2D Vector3 points to compare.</param>
+        /// <returns>
+        /// Returns true if the supplied Vector3 point is inside this Polygon.
+        /// </returns>
+        public bool Contains(IList<Vector3> points)
```
ikeough: If we're going to keep these methods in here, than they should just call the singular version of the same. Ex:
```
Contains(IList<Vector3> points)
{
  foreach(var p in points) 
  {
    if(!Contains(p))
    {
      return false;
    }
  }
}
```

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -91,19 +132,163 @@ public bool Contains(Polygon polygon)
...
+        /// <summary>
+        /// Tests if this Polygon covers the supplied Polygon. This Polygon covers the supplied Polygon if all points of the supplied Polygon fall within or on the perimeter of this Polygon.
+        /// </summary>
+        /// <param name="polygon">The Polygon to test for coverage by this Polygon.</param>
+        /// <returns>
+        /// Returns true if the supplied Polygon is contained by or coincides with the perimeter of the supplied Polygon.
         /// </returns>
         public bool Covers(Polygon polygon)
         {
+            var tolerance = 0.00001;
```
ikeough: This is a magic number. If we always use this tolerance, it should be a `const` at the top of the `Polygon` class.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -91,19 +132,163 @@ public bool Contains(Polygon polygon)
...
+            var clipper = new Clipper();
+            var solution = new List<List<IntPoint>>();
+            clipper.AddPath(this.ToClipperPath(), PolyType.ptSubject, true);
+            clipper.AddPath(polygon.ToClipperPath(), PolyType.ptClip, true);
+            clipper.Execute(ClipType.ctIntersection, solution);
+            if (solution.Count != 1)
+            {
+                return false;
+            }
+            var testPolygon = solution.First().ToPolygon();
```
ikeough: Where is `testPolygon` used?

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -91,19 +132,163 @@ public bool Contains(Polygon polygon)
...
+        }
+
+        /// <summary>
+        /// Tests if this Polygon covers all of the supplied Polygons. This Polygon covers the supplied Polygons if all points of the supplied Polygons fall within or on the perimeter of this Polygon.
+        /// </summary>
+        /// <param name="polygons">The list of Polygons to compare.</param>
+        /// <returns>
+        /// Returns true if any of the supplied Polygons intersects with this Polygon.
+        /// </returns>
+        public bool Covers(IList<Polygon> polygons)
```
ikeough: As mentioned above, we shouldn't duplicate code here. All plural versions of a method should call the singular version.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -91,19 +132,163 @@ public bool Contains(Polygon polygon)
...
+        /// <returns>
+        /// Returns true if the supplied 2D Vector3 point falls outside the perimeter of this Polygon.
+        /// </returns>
+        public bool Disjoint(Vector3 point)
+        {
+            var scale = 1024.0;
             var thisPath = this.ToClipperPath();
-            var polyPath = polygon.ToClipperPath();
-            foreach (IntPoint vertex in polyPath)
+            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
```
ikeough: Isn't this section doing a `Contains`? Can we use our new method here?

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -140,31 +325,124 @@ public bool Disjoint(Polygon polygon)
...
+        }
+
+        /// <summary>
+        /// Tests if any of the supplied Polygons share areas with this Polygon.
+        /// </summary>
+        /// <param name="polygons">The list of Polygons to compare.</param>
+        /// <returns>
+        /// Returns true if any of the supplied Polygons intersects with this Polygon.
+        /// </returns>
+        public bool Intersects(IList<Polygon> polygons)
```
ikeough: As above, use singular `Interesects` in this method.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -140,31 +325,124 @@ public bool Disjoint(Polygon polygon)
...
+        /// Tests if any of the supplied Polygons are coincident with this Polygon.
         /// </summary>
-        /// <param name="polygon">The Polygon to compare.</param>
+        /// <param name="polygons">The collection of Polygons to compare.</param>
         /// <returns>
-        /// Returns true if any vertex of either Polygon is within the perimeter of the other.
+        /// Returns true if any of the supplied Polygons do not intersect or touch.
         /// </returns>
-        public bool Intersects(Polygon polygon)
+        public bool Disjoint(IList<Polygon> polygons)
```
ikeough: As above, use singular `Disjoint` in this method.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -140,31 +325,124 @@ public bool Disjoint(Polygon polygon)
...
+        }
+
+        /// <summary>
+        /// Tests if all the 2D Vector3 points in the supplied collection are on the perimeter of this Polygon.
+        /// </summary>
+        /// <param name="points">The 2D Vector3 point to compare.</param>
+        /// <returns>
+        /// Returns true if all the supplied 2D Vector3 points coincide with the perimeter of this Polygon.
+        /// </returns>
+        public bool Touches(IList<Vector3> points)
```
ikeough: As above, use singular `Touches` in this method.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -69,6 +69,47 @@ public Vector3 Centroid
         /// <exception cref="System.ArgumentException">Thrown when coincident vertices are provided.</exception>
         public Polygon(IList<Vector3> vertices) : base(vertices){}
 
+        /// <summary>
+        /// Tests if the supplied 2D point is within the perimeter of this Polygon.
```
ikeough: See my note below on the use of "2D point".

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -69,6 +69,47 @@ public Vector3 Centroid
...
+        /// </summary>
+        /// <param name="point">The 2D Vector3 point to compare.</param>
```
ikeough: See not below about "2D Vector3".

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -69,6 +69,47 @@ public Vector3 Centroid
...
+        /// <returns>
+        /// Returns true if the supplied Vector3 point is inside this Polygon.
```
ikeough: Can we extend this comment to say "Returns true if the supplied Vector3 is inside this Polygon, and false if the supplied Vector3 is outside this Polygon or coincident with it."? The edge condition is when the point is on the polygon. The comment should clarify what happens in that situation.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -69,6 +69,47 @@ public Vector3 Centroid
...
+            var intPoint = new IntPoint(point.X * scale, point.Y * scale);
+            if (Clipper.PointInPolygon(intPoint, thisPath) != 1)
+            {
+                return false;
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Tests if all the supplied 2D Vector3 points fall within the perimeter of this Polygon.
```
ikeough: "Tests if all of the supplied Vector3 fall within..."

ikeough:
You'll also need to bump the patch number of the library to include these changes.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Yup it's magic. Should we do the same with the 1024.0 scaling factor needed by Clipper?

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Bleah, I'll check. Might be a leftover of another idea.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: As above, originally how I had it. I changed it because perhaps I misunderstood this Slack comment:
ikeo [5:08 PM]
In looking through the changes my first instinct is to remove all the methods that just contain a loop calling the method.
This will reduce the size of the PR. It may be that we want to add these back in the future. But if we do this, I want it to be applied across the library.

Anthony Hauck [5:09 PM]
I figured that was a pattern you wouldn't like much -- I can self-contain those, but it's pretty annoying to have external loops everywhere to call the base method. (edited)

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Right-o

csharp/src/elements/Geometry/Polygon.cs
hummingburd: As above.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Acknowledged.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Absolutely.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Aye aye.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Used in "if" statement immediately following.

csharp/src/elements/Geometry/Polygon.cs
ikeough: > ...my first instinct is to remove all the **methods** that just...

That comment was meant to suggest that we should remove them from this PR entirely, and allow the user to call the function in a loop however they'd like.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done for both tolerance (now "areaTolerance" and scale)

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done

csharp/src/elements/Geometry/Polygon.cs
hummingburd: Done, I believe.

csharp/src/elements/Geometry/Polygon.cs
hummingburd: I see you point, and that we should make it a larger pattern if we go here. I'll pull the plural methods out of the base library, stand by, should be done by the time you're starting your day.

hummingburd: Done.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -94,29 +94,6 @@ public bool Contains(Vector3 point)
...
         }
 
-        /// <summary>
-        /// Tests if all the supplied Vector3 points are within this Polygon without coincidence with an edge when compared on a shared plane.
-        /// </summary>
-        /// <param name="points">The list of Vector3 points to compare to this Polygon.</param>
-        /// <returns>
-        /// Returns true if all the supplied Vector3 points are within this Polygon when compared on a shared plane. Returns false if any of the Vector3 points are outside this Polygon or if the supplied Vector3 point list is null.
-        /// </returns>
-        public bool Contains(IList<Vector3> points)
```
ikeough: Thank you! There's a better than even chance that we'll put these back in some form one day. But I want to be sure when we do that it's a consistent design.

hummingburd: Yeah, I realized you were right, this should be a library-wide approach if / when we do it. I only need one of the plural functions for now, so I'll instantiate it in my HyparSpaces library as a Polygon extension. However, I need all of the single inputs in various bit of HyparSpaces.

hummingburd:
Not sure why the tests are failing -- local tests check out.

csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -70,15 +73,18 @@ public Vector3 Centroid
...
 
         /// <summary>
-        /// Tests if the supplied 2D point is within the perimeter of this Polygon.
+        /// Tests if the supplied Vector3 point is within this Polygon without coincidence with an edge when compared on a shared plane.
         /// </summary>
-        /// <param name="point">The 2D Vector3 point to compare.</param>
+        /// <param name="point">The Vector3 point to compare to this Polygon.</param>
         /// <returns>
-        /// Returns true if the supplied Vector3 point is inside this Polygon.
+        /// Returns true if the supplied Vector3 point is within this Polygon when compared on a shared plane. Returns false if the Vector3 point is outside this Polygon or if the supplied Vector3 point is null.
```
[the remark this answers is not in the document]
hummingburd: Pushed changes.

PR #92, opened 2018-12-05
Fixed bug in "Covers" which would give false negatives.

hummingburd:
1) Area compare method of Intersection was sometimes faulty due to integer / double conversion. This improved method is exact, deletes a few lines of code, and eliminates the "magic number" areaTolerance.

2) Bumped version number.

### DIFF
--- a/csharp/src/elements/Geometry/Polygon.cs
+++ b/csharp/src/elements/Geometry/Polygon.cs
@@ -346,7 +346,7 @@ namespace Hypar.Geometry
             var polygons = new List<Polygon>();
             foreach (List<IntPoint> path in solution)
             {
-                polygons.Add(PolygonExtensions.ToPolygon(path));
+                polygons.Add(PolygonExtensions.ToPolygon(path.Distinct().ToList()));
             }
             return polygons;
         }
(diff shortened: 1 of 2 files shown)


csharp/src/elements/Geometry/Polygon.cs
```diff
@@ -163,12 +162,7 @@ public bool Covers(Polygon polygon)
             {
                 return false;
             }
-            var testPolygon = solution.First().ToPolygon();
-            if (Math.Abs(polygon.Area - testPolygon.Area) > areaTolerance)
-            {
-                return false;
-            }
-            return true;
+            return solution.First().ToPolygon().Area == polygon.ToClipperPath().ToPolygon().Area;
```
ikeough: `First` will throw an exception if `solution` is empty or null. Is there a situation in which that might occur?

hummingburd: No, that's handled just above:
  if (solution.Count != 1)
            {
                return false;
            }

ikeough: Got it. Didn't see the code above the fold.

PR #104, opened 2019-02-16
Search for IElements in all assemblies 

wynged:
rather than the just the Hypar Elements library


ikeough:
Seems like this is failing because number of tests related to structural framing are failing during deserialization of json. In the future I recommend doing `dotnet test` and it'll show you this on your end before you create the PR. The tests are failing with the following:
```
Could not load type 'Microsoft.VisualStudio.TestPlatform.ObjectModel.Navigation.DiaSourceClass' from assembly 'Microsoft.VisualStudio.TestPlatform.ObjectModel, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```
This happens when you try to reflect on an assembly that can't be loaded for reflection. So our logic for loading is too broad.

ikeough:
We may be able to limit this using [`Assembly.GetEntryAssembly`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=netframework-4.7.2). You would have to use that plus `Assembly.GetExecutingAssembly` to get both the built-in `Element` types and your types.

ikeough:
What is the expectation with this work? Is the expectation that you'll be able to serialize a model containing your `Element` types to JSON and have that be deserialized by another instance of Elements? If so, the deserializing instance would have to have access to your type. How we allow arbitrary elements to be serialized and deserialized into a `Model` requires some design.

wynged:
It makes sense that we would need access to the types if we are going to
Deserialize.  Currently we are creating a "core" library that all of our
separate functions reference, and our personalized Elements will live in
there.  This means get executing assembly wouldn't cover it.  One thought
that might work would be to use the Fody wrapper Costura to pull all of the
dll into one... But I don't know if that would make .GetExecutingAssembly()
to work off the top of my head.  If that was the answer, maybe that Costura
step could be built into the 'hypar publish' step?


When I run dotnet test locally on the elements library I don't get any
errors... I even confirmed I think that the dll in your error is indeed
being loaded.  Microsoft.VisualStudio.TestPlatform.ObjectModel is among the
Assemblies inspected when I run the tests locally.  And 97/98 tests
succeed.  Elements.Tests.IfcTests.IFC is skipped for some reason.

Final thought.  Would there be a way to make iteration over all Assemblies
protected against this error?  Either simply try-catch, or maybe some
property that let's us check if it is safe to load types from an assembly.
I don't see the error right now, so it's hard to test these options.

ikeough:
Restating the problem: we need a way for those who write types which extend the Elements library to serialize and deserialize those types as part of a `Model`. What I believe we need to do is implement an extensions schema which allows you to specify that your types come from and extension. And the deserializing system has to support that extension or reject the request to deserialize. This is similar to what glTF has with extensions. I'm opening a new issue to track the addition of this functionality.

ikeough:
Tracking in https://github.com/hypar-io/elements/issues/new.

ikeough:
@wynged A slightly more robust fix is proposed in https://github.com/hypar-io/elements/pull/115. Does that work for you? If so, I'm going to close this and merge that.

wynged:
Looks cool to me!  Nice solution that doesn't appear to need any
intervention from the user's end, while also not loading assemblies wildly.
close away.

wynged:
so, because this is solved in the extensions pull request, I"m closing this one to tidy up.

ikeough:
Thanks. I'll merge the PR.

Issue #97, opened 2019-01-28
Proposal for the introduction of mechanical Element types.


ikeough:
The Elements library should have types which represent mechanical systems and equipment according to @wynged. What's the simplest way that we can define components in a directed mechanical system?. What's the simplest representation that we can provide of the connecting elements and the equipment within those systems?

```c#

// A component in a directed mechanical system.
// The ICurveProvider interface requires an ICurve representing the center line of the Element.
// The IProfileProvider interface requires a Profile representing the cross-section shape of the element.
public interface IFlowElement : IElement, ICurveProvider, IProfileProvider
{
 // The port from which this Duct begins. 
 Port From{get;}

 // The port to which the end of this Duct attaches.
 Port To{get;}

 // The profile of the Duct. Ex: round.
 Profile Profile{get;}
}

public class Port
{
  IFlowElement From{get;}
  IEquipment To{get;}
}

// A mechanical Duct.
public class Duct : IFlowElement{...}

// An electrical conduit.
public class Conduit : IFlowElement{...}

// A piece of mechanical equipment which will be represented
// by a box with the given dimensions and have Ports
public interface IEquipment
{
  Port[] Ports{get;}
  double Width{get;}
  double Length{get;}
  double Height{get;}
  Equipment(Port[] ports, double length, double width, double height){...}
}
public class VAV : IEquipment{...}
public class Fan : IEquipment{...}
public class Diffuser : IEquipment{...}
```

wynged:
- The connections between equipment should be directly port to port only on equipment.  Most important thing is that equipment is connected to each other.  _how_ it's connected is a drafting exercise, and I think should be handled differently.
  - Note, this is still an interesting task, how to route / draw the connections, just not relevant to the fact that they are connected. 
- Connections then should be Port to Port similar to IDirectedFlow just as you've shown.  I'd rename to IConnection for the reasons below
  - Is IDirectedFlow and IFlowElement meant to be the same?  I don't See IFlowElement defined anywhere
- Question now in my conception is how to manage traversal.  This could be done I believe by having each port have a field for the equipment it resides on, its “parent” so one would go equipment->port->connection->otherport->Parent-Equipment 
  - subtyping ports as either Inport or Outport might simplify this even further.
  - ports now also need to know about their connection…? Is this getting messy?

wynged:
[...]

PR #120, opened 2019-03-11
Model Serialization settings to help serialization of Connected System interfaces

wynged:
This contains changes to the JSON serialization settings to help with:
1) Deserializing properties that are interfaces, using Newtonsofts Type handling
2) Serializing the `Parent` property of connectors, which is necessary for traversal, but can create a serialization loop. #97

### DIFF
--- a/src/Dynamo/Class1.cs
+++ b/src/Dynamo/Class1.cs
@@ -1,0 +1,8 @@
+﻿using System;
+
+namespace Hypar.Elements.Dynamo
+{
+    public class Class1
+    {
+    }
+}
--- a/src/Dynamo/Hypar.Elements.Dynamo.csproj
+++ b/src/Dynamo/Hypar.Elements.Dynamo.csproj
@@ -1,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+</Project>
--- a/src/Dynamo/README.md
+++ b/src/Dynamo/README.md
@@ -1,0 +1,2 @@
+# Dynamo
+Elements converters for Dynamo.


src/Elements/Serialization/ModelConverter.cs
```diff
@@ -98,7 +100,9 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                         new ProfileToIdConverter(model.Profiles),
                         new SolidConverter(model.Materials)
                     },
-                NullValueHandling = NullValueHandling.Ignore
+                NullValueHandling = NullValueHandling.Ignore,
+                TypeNameHandling = TypeNameHandling.Auto,
```
ikeough: I don't think we want to add this. `TypeNameHandling` will add a field `$type` to all serialized objects, with the fully namespaced type that is the target for deserialization. We've already got a version of this with our `type` field. 

Is there a reason you need it?

test/PolygonTests.cs
```diff
@@ -501,6 +501,10 @@ public void PointAt()
             var p = new Polygon(new[]{a,b,c,d});
             Assert.Equal(4, p.Segments().Count());
             Assert.Equal(new Vector3(1.0, 1.0), p.PointAt(0.5));
+
+            var r = Polygon.Rectangle(2,2);
+            Assert.Equal(new Vector3(1,1,0), r.PointAt(0.5));
```
ikeough: This seems like it should belong to #121. Do you want to include it here and close that PR?

wynged: you are right.  I'd rather include it in the other pull request, I'm just all over the place with my forked branches.  I'm inlcuding it there now....not sure how to un-include this change here.

src/Elements/Serialization/ModelConverter.cs
wynged: first, sorry, I broke this pull request trying to clean.  let's continue the convo here.
I have needed it to deserialize elements with properties where the property type is an interface, not specific classes.  This appears in my lastest versions of creating/using `IConnect` interfaces, for example.
```c#
    public interface IOneToManyConnection
    {
        IConnector Start{get;}
        IConnector[] End{get;}
    }
```
When one tries to deserialize a class that implements this interface, the Start and End properties throw errors, because whatever is stored there cannot have it's type identified.
I see now that the element type property seems like it SHOULD work... 
How about this for proceeding, are you ready to include those interfaces in Hypar.Elements?  I have some slight tweaks but I could make a different pull request that contained the serialization settings, the interfaces and some tests...  that sound like a plan, or do you want to hold off on putting those interfaces in the library?

wynged: Just double checked, and when I serialize/deserialize the first sample of json does not work, but the second does.  It's an MEPConnector object stored in a `Dictionary<string, IConnector>`  I see that a type is written in both cases, but only when the $type property is there does the deserialization work... thoughts?
```json
 "Room Air": {
        "id": 23,
        "type": "mister.core.systems.mepconnector",
        "Direction": 0,
        "Medium": {
          "UniqueName": "Room Air"
        },
```
```json
 "Room Air": {
        "$type": "Mister.Core.Systems.MEPConnector, Mister.Core",
        "id": 23,
        "type": "mister.core.systems.mepconnector",
        "Direction": 0,
        "Medium": {
          "$type": "Mister.Core.Systems.Medium, Mister.Core",
          "UniqueName": "Room Air"
        },
```

wynged: @ikeough   Just want to make sure you see this.  Not sure if notifications happen with closed pull requests.

PR #121, opened 2019-03-15
Remove the .Normalize() to fix PointAtInternal

wynged:
The Normalize method call when calculating the current vector is causing errors.  Line 166 does not want a normalized vector, it wants to compute how much of the full vector is needed to reach the desired paramter.  
Suggest adding a test in Polyline tests that looks like
``` C#
            var r = Polygon.Rectangle(2,2);
            Assert.Equal(new Vector3(1,1,0), r.PointAt(0.5));
```
This test, which uses a rectangle not of unit length will reveal the problem.

### DIFF
--- a/src/Elements/Geometry/Polyline.cs
+++ b/src/Elements/Geometry/Polyline.cs
@@ -159,7 +159,7 @@ namespace Elements.Geometry
                 var a = this._vertices[i];
                 var b = this._vertices[i + 1];
                 var currLength = a.DistanceTo(b);
-                var currVec = (b - a).Normalized();
+                var currVec = (b - a)
                 if (totalLength <= d && totalLength + currLength >= d)
                 {
                     segmentIndex = i;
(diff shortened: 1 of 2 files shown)


ikeough:
This is a great catch! Thanks @wynged.

src/Elements/Geometry/Polyline.cs
```diff
@@ -159,7 +159,7 @@ private Vector3 PointAtInternal(double u, out int segmentIndex)
                 var a = this._vertices[i];
                 var b = this._vertices[i + 1];
                 var currLength = a.DistanceTo(b);
-                var currVec = (b - a).Normalized();
+                var currVec = (b - a)
```
ikeough: Missing a `;`, which is causing the build to fail. Can you update? Then I'll merge.

wynged:
This now contains the test that was accidentally added to  #120

PR #132, opened 2019-04-21
Save the profile when space created with a profile

wynged:
Thought this was already in there, but it's missing in 0.2.2.


ikeough:
This will appear in 0.2.4.2. I'm not merging this PR as there's been a lot of reorg in that class and its easier to just include your one line change.

wynged:
Works for me!

PR #137, opened 2019-04-24
IAggregateElement sub elements should be actually added to the model

wynged:
currently the only code being called is the call to get RootElements, which will be called anyways if AddElement is called.

### DIFF
--- a/src/Elements/Model.cs
+++ b/src/Elements/Model.cs
@@ -310,7 +310,7 @@ namespace Elements
                 {
                     foreach (var esub in ae.Elements)
                     {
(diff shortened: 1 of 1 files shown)


ikeough:
This is a great catch. Thanks @wynged. Although, I'm not going to add the suggested fix as it puts the logic one level too deep. IOW, the adding of the sub elements as you have proposed is a side affect of getting other data (profiles, materials, etc.). In 0.2.4.2, I'll add this functionality at the same level as where we add a single element to make it a little more obvious what is happening.

wynged:
Makes sense.  Thanks!

PR #153, opened 2019-06-25
Test that Geometry for spaces is actually created if profile used

wynged:
Spaces currently don't have their geometry created if they are created with a profile.  This adds a test that geometry exists and also generates the geometry.

If users are meant to generate the geometry themselves from the properties of the IExtrude interface, that's fine, but then the space should probably not be an ISolid, which implies there will be readily available geometry for the user to access.

### DIFF
--- a/src/Elements/Space.cs
+++ b/src/Elements/Space.cs
@@ -110,6 +110,7 @@ namespace Elements
             this.Transform = transform != null ? transform : new Transform(new Vector3(0, 0, elevation));
             this.Material = material == null ? BuiltInMaterials.Mass : material;
             this.Height = height;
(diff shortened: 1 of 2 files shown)


ikeough:
Apologies for the lack of clarity here. We’ve removed the geometry creation methods from the constructors in favor of generating the geometry directly before visualization. This change would break that pattern. I do however understand why you’d want to have the geometry available directly after construction. So I’m clear on the use case, what do need to do with the geometry?

ikeough:
I believe we can close this now in favor of some new functionality. You can now call `ISolid.GetUpdatedSolid()` to get the solid.

PR #154, opened 2019-06-25
Don't return after adding IAggregateElement geometry

wynged:
Users (aka Eric) might want to add geometry of an element as well as the geometry of the aggregated elements, so we should not return from the GetRenderDataForElement method just because we've finished adding the aggregated geometries.


wynged:
Awesome, glad this got pulled.  
When will this change be visible in a nuget package, and is that the same moment when it is available on the Hypar platform?

ikeough:
It's available on the Hypar platform when you include it in a function. I'll deploy a new version now.

ikeough:
Hypar 0.2.15 has just been deployed. It might take a few minutes to get listed on NuGet.

PR #161, opened 2019-07-01
Fixed Polygon Centroid coordinates for all quadrants.

hummingburd:
I'm an idiot. For some reason I was moving all coordinates into the first quadrant. Got rid of Math.Abs of x and y coordinates at conclusion of Polygon Centroid calculation.


hummingburd:
Elaborated Polygon.Centroid test to trial a square in each Cartessian quadrant and a bow tie shape in Quadrants I and III.

PR #172, opened 2019-07-09
Returning absolute value of area

anthonyhauck:
It looks as if I forgot to push this. Sorry.

### DIFF
--- a/src/Elements/Geometry/Polygon.cs
+++ b/src/Elements/Geometry/Polygon.cs
@@ -567,19 +567,12 @@ namespace Elements.Geometry
         {
             var x = 0.0;
             var y = 0.0;
-            var factor = 0.0;
-            for (var i = 0; i < this._vertices.Length; i++)
+            foreach (var pnt in Vertices)
             {
-                factor =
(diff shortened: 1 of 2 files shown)


ikeough:
Sorry I'm just looping back to this @anthonyhauck. Can you investigate why the build is failing? Then I'll merge.

anthonyhauck:
Will do sometime today.

anthonyhauck:
Previous Centroid method was dependent on negative Areas for correct result in Quadrant 2. Revised Centroid method to much simpler and dependable method of minimizing distances to all Polygon vertices.

anthonyhauck:
Putting these changes into the Floor extrusion change instead.

ikeough:
@anthonyhauck We didn't merge this because I don't think the behavior is wrong. We can fix the documentation to make it clear that the area calculation returns the "signed area" where a negative area means a CCW winding. This gives us two benefits:
- The area of a set of polygons with different windings is simply the sum of its areas.
- A developer can determine the winding of a Polygon by just getting the area, or we can provide a method like `IsClockwise` which checks the sign of the area.

PR #180, opened 2019-08-15
Added constructor boolean to set floor extrusion direction along z-axis

anthonyhauck:
- Renamed constructor "profile" argument to "polygon" for clarity.
- Reversed polygon vertex order for negative z-axis extrusion to avoid making a hole.
- Added an associated test with optional model creation.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/180)

### DIFF
--- a/src/Elements/Floor.cs
+++ b/src/Elements/Floor.cs
@@ -69,7 +69,8 @@ namespace Elements
         /// <param name="elevation">The elevation of the top of the floor.</param>
         /// <param name="transform">The floor's transform. If set, this will override the floor's elevation.</param>
         /// <param name="openings">An array of openings in the floor.</param>
-        public Floor(Polygon profile, FloorType elementType, double elevation = 0.0, Transform transform = null, List<Opening> openings = null)
+        /// <param name="extrudeUp">A boolean indicating the polygon extrusion direction along either the positive or negative z-axis. Defaults to positive.</param>
+        public Floor(Polygon profile, FloorType elementType, double elevation = 0.0, Transform transform = null, List<Opening> openings = null, bool extrudeUp = true)
         {
             this.Profile = new Profile(profile);
             if(openings != null)
@@ -80,7 +81,14 @@ namespace Elements
             this.ElementType = elementType;
             var thickness = elementType.Thickness();
             this.Transform = transform != null ? transform : new Transform(new Vector3(0, 0, elevation));
-            this.ExtrudeDirection = Vector3.ZAxis;
+            if (extrudeUp)
+            {
+                this.ExtrudeDirection = Vector3.ZAxis;
+            }
+            else
+            {
+                this.ExtrudeDirection = Vector3.ZAxis * -1;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Elements/Floor.cs
```diff
@@ -64,23 +65,35 @@ public List<Opening> Openings
...
             var thickness = elementType.Thickness();
-            this.Transform = transform != null ? transform : new Transform(new Vector3(0, 0, elevation));
-            this.ExtrudeDirection = Vector3.ZAxis;
+            this.Transform = transform ?? new Transform(new Vector3(0, 0, elevation));
+            if (extrudeUp)
+            {
+                this.Profile = new Profile(polygon);
+                this.ExtrudeDirection = Vector3.ZAxis;
+            }
+            else
```
ikeough: If we don't allow the option to extrude up, this condition can just be replaced with `Polygon.Reversed()`

This whole conversation makes me wonder if we want to change the construction of floor to be plane-based. So that you provide a polygon and a plane, and it projects the polygon onto the plane and extrudes down from the plane. Then you could create many floors using the same polygon, but different planes.

src/Elements/Floor.cs
```diff
@@ -96,15 +109,14 @@ public Floor(Profile profile, Transform start, Vector3 direction, FloorType elem
         {
             this.Elevation = elevation;
             this.ElementType = elementType;
-            this.Transform = transform != null ? transform : new Transform(new Vector3(0, 0, elevation));
+            this.Transform = transform ?? new Transform(new Vector3(0, 0, elevation));
```
ikeough: This is great. I need to start using the null coalescing operator.

src/Elements/Floor.cs
anthonyhauck: I like that idea. After I revise this, I'll open another branch.

anthonyhauck:
Made requested changes and found that a previous merge involving polygon area and polygon centroid calculation revisions got reverted on at some point. Restored in this PR.

anthonyhauck:
Question before you merge -- do I need to touch [JsonConstructor] as well?

src/Elements/Floor.cs
```diff
@@ -64,55 +65,55 @@ public List<Opening> Openings
...
             {
                 this._openings = openings;
             }
             this.Elevation = elevation;
             this.ElementType = elementType;
             var thickness = elementType.Thickness();
-            this.Transform = transform != null ? transform : new Transform(new Vector3(0, 0, elevation));
-            this.ExtrudeDirection = Vector3.ZAxis;
+            this.Transform = transform ?? new Transform(new Vector3(0, 0, elevation));
+            this.Profile = new Profile(perimeter.Reversed());
```
ikeough: I've thought about this a bit and I'm not sure it works:
- Floors created in this way will now have their perimeters wound in the opposite direction of floors created using the ctor which takes a profile.

Things we can try:
- We could set the profile down and extrude "up", but this would then report the profile below the level of the slab.
- We could set the profile down and extrude "up", but report the perimeter at the correct height.

ikeough: In general, I think we need to revisit the design of these floor constructors. They were a mashup of stuff to support IFC (see the `start` parameter - which doesn't make sense to a user). We'd be better off going straight to floor planes.

PR #215, opened 2019-12-05
Produce valid empty GLTFs when no elements produce geometry.

mattc:
This matches the behavior when there are no elements.
Manual testing: tested functions with and without geometry.

Fixes hypar-io/Explore#288

### DIFF
--- a/src/Elements/Serialization/glTF/GltfExtensions.cs
+++ b/src/Elements/Serialization/glTF/GltfExtensions.cs
@@ -37,6 +37,7 @@ namespace Elements.Serialization.glTF
 
         /// <summary>
         /// Save a model to gltf.
+        /// If there is no geometry, an empty GLTF will still be produced.
         /// </summary>
         /// <param name="model">The model to serialize.</param>
         /// <param name="path">The output path.</param>
@@ -51,6 +52,7 @@ namespace Elements.Serialization.glTF
                     {
                         return;
                     }
+                    // Else fall through to produce an empty GLTF.
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/Elements/Serialization/glTF/GltfExtensions.cs
```diff
@@ -43,20 +43,25 @@ private static int GetNextId()
...
+            if(model.Elements.Count > 0)
             {
-                File.WriteAllText(path, emptyGltf);
-                return;
+                if(useBinarySerialization)
+                {
+                    if (SaveGlb(model, path))
+                    {
+                        return;
+                    }
```
ikeough: Else what? Should we throw an exception or return false here as well?

We're not consistent in this with other `To` methods. In `ToJson` if there's a problem serializing an exception will be thrown. I either want all the `To` methods to return `bool` or all of them to throw exceptions. I looked around a bit to see if there are best practices for serialization method signatures and couldn't find any rules that people follow, but If you look at something like [File.WriteAllText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.writealltext?view=netframework-4.8), you'll see that it returns `void` but with a number of very specific exceptions. So we could throw an `EmptyModelGeometryException` in `ToGlTF` and the caller would do something like:
```
try
{
  myModel.ToGlTF(...);
}
catch(EmptyModelGeometryException ex)
{
  Console.WriteLine("The model does not contain any geometry.")
}
```

mattc: Discussed offline. I pushed more comments to make the behavior explicit (producing an empty GLTF). This is consistent with system methods: if you serialize an empty model, you get an empty file, not an error.

PR #222, opened 2019-12-26
Regenerate docs with latest version of DocFx.

mattc:
This minimizes the diff from my impending followup change. I did not make any explicit changes, so these only reflect changes to DocFX and possibly any changes made to the code since docs were updated.


anthonyhauck:
Most of the questions I have are more about Elements than these documentation changes.

PR #225, opened 2020-01-03
add additional Scale overloads

andrewheumann:
Adding a few utility overloads to Transform.Scale(): 
```csharp
Transform.Scale(double amount)
Transform.Scale(double amount, Vector3 origin)
```
open to better/more efficient ways to scale about an origin without the double-translate.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/225)

### DIFF
--- a/src/Elements/Geometry/Transform.cs
+++ b/src/Elements/Geometry/Transform.cs
@@ -322,6 +322,27 @@ namespace Elements.Geometry
         {
             return new Plane(this.Origin, this.YAxis);
         }
+        /// <summary>
+        /// Scale uniformly about the origin.
+        /// </summary>
+        /// <param name="amount">The amount to scale uniformly</param>
+        public void Scale(double amount)
+        {
+            Scale(new Vector3(amount, amount, amount));
+        }
+        /// <summary>
+        /// Scale uniformly about a point
+        /// </summary>
+        /// <param name="amount">The scale factor</param>
+        /// <param name="origin">The origin of scaling</param>
+        /// /// <returns></returns>
+        public void Scale(double amount, Vector3 origin)
+        {
+            Move(origin.Negate());
+            Scale(amount);
+            Move(origin);
+
+        }
 
         /// <summary>
         /// Create a transform that is oriented along 
--- a/src/Elements/Geometry/Transform.cs
+++ b/src/Elements/Geometry/Transform.cs
@@ -43,13 +43,14 @@ namespace Elements.Geometry
         {
             this.Matrix = new Matrix();
         }
-        
+
         /// <summary>
         /// Create a transform by copying another transform.
         /// </summary>
         /// <param name="t">The transform to copy.</param>
-        public Transform(Transform t): 
-            this(new Matrix(new Vector3(t.XAxis), new Vector3(t.YAxis), new Vector3(t.ZAxis), new Vector3(t.Origin))){}
+        public Transform(Transform t) :
+            this(new Matrix(new Vector3(t.XAxis), new Vector3(t.YAxis), new Vector3(t.ZAxis), new Vector3(t.Origin)))
+        { }
 
         /// <summary>
         /// Create a transform with a translation.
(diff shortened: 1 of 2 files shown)


src/Elements/Geometry/Transform.cs
```diff
@@ -322,6 +322,27 @@ public Plane XZ()
         {
             return new Plane(this.Origin, this.YAxis);
         }
+        /// <summary>
+        /// Scale uniformly about the origin.
+        /// </summary>
+        /// <param name="amount">The amount to scale uniformly</param>
+        public void Scale(double amount)
```
mattc: Although it'd be inconsistent with the original function, I'd still call this the factor rather than the amount to make it clear this is multiplicative.

src/Elements/Geometry/Transform.cs
```diff
@@ -322,6 +322,27 @@ public Plane XZ()
...
+        {
+            Scale(new Vector3(amount, amount, amount));
+        }
+        /// <summary>
+        /// Scale uniformly about a point
+        /// </summary>
+        /// <param name="amount">The scale factor</param>
+        /// <param name="origin">The origin of scaling</param>
+        /// /// <returns></returns>
```
mattc: Remove this line.

src/Elements/Geometry/Transform.cs
```diff
@@ -322,6 +322,27 @@ public Plane XZ()
...
+        public void Scale(double amount, Vector3 origin)
+        {
+            Move(origin.Negate());
+            Scale(amount);
+            Move(origin);
+
```
mattc: Remove empty line.

src/Elements/Geometry/Transform.cs
```diff
@@ -322,6 +322,27 @@ public Plane XZ()
...
+        /// Scale uniformly about a point
+        /// </summary>
+        /// <param name="amount">The scale factor</param>
+        /// <param name="origin">The origin of scaling</param>
+        /// /// <returns></returns>
+        public void Scale(double amount, Vector3 origin)
+        {
+            Move(origin.Negate());
+            Scale(amount);
+            Move(origin);
```
mattc: I think these three lines are equivalent to:
```
Scale(amount);
Move(origin * amount);
```

but please verify.

andrewheumann:
___
*[src/Elements/Geometry/Transform.cs, line 343 at r1](https://reviewable.io/reviews/hypar-io/elements/225#-LxvuUtG7N2FjpDgshHc-r1-343:-Ly9uYnY5WVPM7xSy1Si:b7aq9o0) ([raw file](https://github.com/hypar-io/elements/blob/adcaaffd1308c1e4563473d7a68c1fee7cda158b/src/Elements/Geometry/Transform.cs#L343)):*

I think these three lines are equivalent to:
```
Scale(amount);
Move(origin * amount);
```

but please verify.
</blockquote></details>

does not appear to be the case:
scale pos=4 about origin=2 by factor=3 should be 8:
```
4 - 2 = 2
2 * 3 = 6
6 + 2 = 8
```
vs. 
```
4 * 3 = 12
12 + (2 * 3) = 18
```

To simplify I could just construct a matrix, although this might be less obvious in the code? 
```
[f,         0,         0,         0,
 0,         f,         0,         0,
 0,         0,         f,         0,
 (1-f)*o.x, (1-f)*o.y, (1-f)*o.z, 0]
```

andrewheumann:
___
*[src/Elements/Geometry/Transform.cs, line 343 at r1](https://reviewable.io/reviews/hypar-io/elements/225#-LxvuUtG7N2FjpDgshHc-r1-343:-LyA5MxN1O7NJMANi347:b8e480r) ([raw file](https://github.com/hypar-io/elements/blob/adcaaffd1308c1e4563473d7a68c1fee7cda158b/src/Elements/Geometry/Transform.cs#L343)):*

does not appear to be the case:
scale pos=4 about origin=2 by factor=3 should be 8:
```
4 - 2 = 2
2 * 3 = 6
6 + 2 = 8
```
vs. 
```
4 * 3 = 12
12 + (2 * 3) = 18
```

To simplify I could just construct a matrix, although this might be less obvious in the code? 
```
[f,         0,         0,         0,
 0,         f,         0,         0,
 0,         0,         f,         0,
 (1-f)*o.x, (1-f)*o.y, (1-f)*o.z, 0]

</blockquote></details>

in fact, the simplification is:
```
Scale(amount);
Move(origin * (1-amount));
```

👾Algebra! 👾
```
(p - o)*f + o

pf - of + o

pf + o(1 -f)
```


<!-- Sent from Reviewable.io -->

PR #233, opened 2020-01-31
1d and 2d Grids

andrewheumann:
This PR adds 1d and 2d grid functionality in the `Grid1d` and `Grid2d` classes, along with some supporting changes. 

### Supporting changes include: 
* Creation of `Elements.Math` namespace and `Elements.Math.MathUtils` utility class
* Creation of `Domain1d` struct to represent a numerical interval
*  Bug fixes for `BBox3` Polygon<> solver
* Override to `Color`'s "ToString()" method
* Addition of a `BooleanTwoSets` Method to the polygon class, which can handle two lists of polygons to Union, Difference, XOr, or Intersect

In particular I would appreciate suggestions for rounding out the unit tests and any thoughts around the consistency, idiomaticity, and general design of the Grid classes from a user's standpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/233)


andrewheumann:
___
*[src/Elements/MathUtils/DomainExtensions.cs, line 52 at r4](https://reviewable.io/reviews/hypar-io/elements/233#-LzxvvRE2Xiaf_jJ5DdQ:-M-AoaQKDjgefH99EbFB:b39mts0) ([raw file](https://github.com/hypar-io/elements/blob/a78e3f47b393eb5b5137707968041ad00b29825c/src/Elements/MathUtils/DomainExtensions.cs#L52)):*

How would you handle this in the case I've got here — where the user can override this tolerance? I can't supply a default parameter `tolerance = Vector3.Epsilon` since it has to be a compile-time constant. Make it a `double?` and set to null by default? 
</blockquote></details>

Or I suppose just a separate function for with / without default tolerance would be the cleanest.

PR #242, opened 2020-02-10
bug fix for grids with transforms

andrewheumann:
previously the Grid2d constructor which consumes a Transform had its Transforms backwards, so grids created with transforms would be aligned incorrectly.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/242)


src/Elements/Spatial/Grid2d.cs
```diff
@@ -100,8 +100,8 @@ public Grid2d(double uDimension, double vDimension)
         /// The polygon's bounding box parallel to the supplied transform will be
         /// used as the grid extents. 
         /// </summary>
-        /// <param name="boundary"></param>
-        /// <param name="t">A transform representing the </param>
+        /// <param name="boundary">The external boundary of this grid system.</param>
+        /// <param name="t">A transform representing the alignment of the grid.</param>
```
ikeough: Can we rename `t` to `transform`? Apologies that I haven't caught this in other places. I'm trying to make this library as friendly to devs getting started as possible, so I'm erring on the side of legibility.

src/Elements/Spatial/Grid2d.cs
```diff
@@ -129,9 +129,9 @@ public Grid2d(IList<Polygon> boundaries, Transform t = null)
             }
 
 
-            toGrid = new Transform(t);
-            t.Invert();
             fromGrid = new Transform(t);
+            t.Invert();
```
ikeough: Transforms are passed by reference, so here you are inverting the transform passed into this method affecting everywhere else it is used. Is this the intention?

ikeough: Perhaps you mean `fromGrid.Invert()`?

andrewheumann: Good catch. I think this is actually doing what I wanted it to, but the possibility of side effects would have been a problem — this is corrected now. The intention is to set `fromGrid` to be the existing transform `transform`, and toGrid to be its inverse.

src/Elements/Spatial/Grid2d.cs
andrewheumann: Done.

PR #246, opened 2020-02-11
Add method for extracting walls from Revit models

wynged:
BACKGROUND:
We want to export walls from Revit as well as floors.

DESCRIPTION:
This PR adds nodes to extract walls, there is a new UserElement class created to make the wall from a face extracted from the Revit wall.

TESTING:
run `dotnet build` and checkout the new nodes.
  
FUTURE WORK:
This only handles Planar walls with a line centerline.

COMMENTS:
There are some geometry issues around the intersections of walls, the joining behavior is not kept with this method of gathering the wall geometry.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/246)

### DIFF
--- a/src/Dynamo/src/HyparDynamo/HyparDynamo.csproj
+++ b/src/Dynamo/src/HyparDynamo/HyparDynamo.csproj
@@ -11,7 +11,6 @@
     <PackageReference Include="DynamoVisualProgramming.DynamoServices" Version="2.5.0.7432">
       <Private>False</Private>
     </PackageReference>
-    <PackageReference Include="Hypar.Elements" Version="0.4.4.2"/>
   </ItemGroup>
 
   <ItemGroup>
@@ -31,6 +30,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\Revit\RevitHyparTools\RevitHyparTools.csproj" />
+    <ProjectReference Include="..\..\..\Elements\Elements.csproj" />
   </ItemGroup>
 
   <ItemGroup>
@@ -38,7 +38,7 @@
   </ItemGroup>
 
 
-  <Target Name="BuildPackage" AfterTargets="Publish" >
+  <Target Name="BuildPackage" AfterTargets="Publish">
     <RemoveDir Directories="..\..\deploy\HyparDynamo\bin\" />     
     <Copy SourceFiles="@(NodeLibraries)" DestinationFolder="..\..\deploy\HyparDynamo\bin\" ContinueOnError="true" />
     <Copy SourceFiles=".\HyparDynamo_DynamoCustomization.xml" DestinationFolder="..\..\deploy\HyparDynamo\bin\" ContinueOnError="true" />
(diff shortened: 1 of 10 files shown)


src/Revit/RevitHyparTools/Create.Wall.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Autodesk.Revit.DB;
+using Elements.Geometry;
+using GeometryEx;
+using ElemGeom = Elements.Geometry;
+
+using Revit = Autodesk.Revit.DB;
+
+namespace RevitHyparTools
```
ikeough: `namespace Hypar.Revit`

src/Dynamo/src/HyparDynamo/HyparDynamo.csproj
```diff
@@ -31,14 +30,15 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\Revit\RevitHyparTools\RevitHyparTools.csproj" />
+    <ProjectReference Include="..\..\..\Elements\Elements.csproj" />
```
ikeough: This should not be a local reference right? This should point to `Elements`?

src/Revit/RevitHyparTools/Create.Wall.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public static partial class Create
```
ikeough: Remind me why this is `partial`.

src/Revit/RevitHyparTools/RevitHyparTools.csproj
```diff
@@ -16,4 +15,7 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\Elements\Elements.csproj" />
```
ikeough: No local reference as per comment above.

src/Revit/RevitHyparTools/WallByProfile.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Elements;
+using Elements.Geometry;
+using Elements.Geometry.Solids;
+
+namespace RevitHyparTools {
+    [UserElement]
+    public class WallByProfile : Wall
```
ikeough: Always add `///` comments to the class and to its methods and fields.

src/Revit/RevitHyparTools/Create.Wall.cs
wynged: ok, i'll do this here and other places

src/Dynamo/src/HyparDynamo/HyparDynamo.csproj
wynged: all this work is inside the Elements repo, and this reference points to the Elements.csproj inside the Elements repo that this project is in.... 
a bit confusing but basically I'm point to an "adjacent" elements.csproj in the same repo, which I think is right, this should always be using the latest elements no?

src/Revit/RevitHyparTools/Create.Wall.cs
wynged: the pattern I have is that there is one Create class spread across multiple files depending on what is being created.  It is more obvious in subsequent work with Floor and columns following the same pattern

src/Revit/RevitHyparTools/RevitHyparTools.csproj
wynged: same thought,  I feel like we should reference the version of Elements that is in this repo, not the nuget version.

src/Dynamo/src/HyparDynamo/HyparDynamo.csproj
ikeough: I see now. The way you’ve done it is correct.

src/Revit/RevitHyparTools/WallByProfile.cs
wynged: this is moving to be a generated type, so I'm not going to do this.

PR #250, opened 2020-02-13
Ray/Solid Intersection

andrewheumann:
BACKGROUND:
- In order to be able to calculate view analyses, a mechanism to intersect rays with context buildings was necessary. 

DESCRIPTION:
- This PR adds a host of `Intersects` overloads to the `Ray` class:
* `Intersects(SolidOperation solidOperation, out Vector3 result)`
* `Intersects(Solid solid, out Vector3 result)` 
* `Intersects(Face face, out Vector3 result)` 
* `Intersects(Plane plane, out Vector3 result)` 

A `ToPolygon()` method is also added to `Loop`. 

TESTING:
- Use any of the new methods to test ray intersection. A new test is added to `RayTests.cs` to demonstrate its use and behavior. 
  
FUTURE WORK:
- This assumes that all Solid Faces are planar. In the future if faces can be non-planar, these methods will need to be adapted to more sophisticated methods.


src/Elements/Geometry/Ray.cs
```diff
@@ -77,6 +78,102 @@ public bool Intersects(Triangle tri, out Vector3 result)
             }
         }
 
+        /// <summary>
+        /// Does this ray intersect with the provided SolidOperation?
+        /// </summary>
+        /// <param name="solidOp">The SolidOperation to intersect with.</param>
+        /// <param name="result">The intersection result.</param>
+        /// <returns>True if an intersection occurs, otherwise false. If true, check the intersection result for the location of the intersection.</returns>
+        public bool Intersects(SolidOperation solidOp, out Vector3 result)
```
ikeough: Can we make the signature `Intersects(SolidOperation solidOp, out List<Vector3> result)`? Then we can record all intersections along the ray.

src/Elements/Geometry/Ray.cs
```diff
@@ -77,6 +78,102 @@ public bool Intersects(Triangle tri, out Vector3 result)
...
+        /// <param name="solid">The Solid to intersect with.</param>
+        /// <param name="result">The intersection result.</param>
+        /// <returns>True if an intersection occurs, otherwise false. If true, check the intersection result for the location of the intersection.</returns>
+        internal bool Intersects(Solid solid, out Vector3 result)
+        {
+            var distanceToClosestResult = double.MaxValue;
+            var faces = solid.Faces;
+            var intersects = false;
+            var finalResult = default(Vector3);
+            foreach (var face in faces)
```
ikeough: As per the comment about, let's store all the intersections and return a `List<Vector3>` ordered by their proximity to the origin of the ray.

src/Elements/Geometry/Solids/Loop.cs
```diff
@@ -32,13 +33,21 @@ public Loop()
...
             this.Edges = new List<HalfEdge>();
-            foreach(var e in edges)
+            foreach (var e in edges)
             {
                 this.Edges.Add(e);
                 e.Loop = this;
             }
         }
 
+        //
```
ikeough: No empty comments on public members :)

test/Elements.Tests/RayTests.cs
```diff
@@ -1,126 +1,162 @@
...
+                    {
+                        continue;
+                    }
+                }
+            }
+        }
+
+
+        [Fact]
+        public void RayIntersectsSolidOperation()
```
ikeough: Can we break up these assertions into separate tests like. I'll suggest some names in line below.

test/Elements.Tests/RayTests.cs
```diff
@@ -1,126 +1,162 @@
...
+            var doesIntersect = ray.Intersects(extrude, out Vector3 result);
+            Assert.True(doesIntersect);
+            Assert.Equal(new Vector3(4, 3, 3), result);
+
+            //right along the edge of two faces
+            var ray2 = new Ray(new Vector3(6, 6, 6), new Vector3(1, 1, 1));
+            var doesIntersect2 = ray2.Intersects(extrude, out Vector3 result2);
+            Assert.False(doesIntersect2);
+
+            //far off
```
ikeough: `RayDoesNotIntersectWhenPointingAwayFromSolid`.

test/Elements.Tests/RayTests.cs
```diff
@@ -1,126 +1,162 @@
...
+            var ray3 = new Ray(new Vector3(6, 6, 0), new Vector3(1, 0,0));
+            var doesIntersect3 = ray3.Intersects(extrude, out Vector3 result3);
+            Assert.False(doesIntersect3);
+
+            //inside
```
ikeough: `RayIntersectsFromInsideSolid`

test/Elements.Tests/RayTests.cs
```diff
@@ -1,126 +1,162 @@
...
+                new Vector3(0,4,0)
+            });
+            var extrude = new solids.Extrude(polygon, 10, new Vector3(1, 1, 1), false);
+
+            var ray = new Ray(new Vector3(-2, 0, 3), new Vector3(2, 1, 0));
+            var doesIntersect = ray.Intersects(extrude, out Vector3 result);
+            Assert.True(doesIntersect);
+            Assert.Equal(new Vector3(4, 3, 3), result);
+
+            //right along the edge of two faces
```
ikeough: `RayIntersectsAlongEdge`

test/Elements.Tests/RayTests.cs
ikeough: There are also a couple of checks that we should add:
- Ray intersection at a vertex.
- Ray intersection when the ray origin is in the plane of a face.

src/Elements/Geometry/Ray.cs
```diff
@@ -77,6 +78,102 @@ public bool Intersects(Triangle tri, out Vector3 result)
...
+        }
+
+        /// <summary>
+        /// Does this ray intersect the provided plane?
+        /// </summary>
+        /// <param name="plane">The Plane to intersect with.</param>
+        /// <param name="result">The intersection result.</param>
+        /// <returns>True if an intersection occurs, otherwise false — this can occur if the ray is very close to parallel to the plane.
+        /// If true, check the intersection result for the location of the intersection.</returns>
+        public bool Intersects(Plane plane, out Vector3 result)
```
ikeough: This is not required for this PR, but might be nice...This code and the `Line.Intersects` code will share the same logic for intersection testing. The line intersection code simply checks if the intersection is past the end of the line as its final step. Can we use the same intersection logic in both places?

src/Elements/Geometry/Ray.cs
andrewheumann: Done.

src/Elements/Geometry/Ray.cs
andrewheumann: Done.

test/Elements.Tests/RayTests.cs
andrewheumann: What should the expected behavior be in these cases? intersection or no?

andrewheumann: I note that IntersectsAtVertex assumes that intersects = true. But if a ray lies exactly along an edge, the intersection is a line, so could not be meaningfully reported without loss of information (ie only one or two vertices). Similarly, if the ray origin is in the plane of the face, the ray might also be parallel to that face and intersect it again, along infinite points. My inclination is actually to always *exclude* vertices/ edges / parallel along faces, but there may be a good reason to want these included and a sensible "result" to return as well.

src/Elements/Geometry/Ray.cs
andrewheumann: Done!

test/Elements.Tests/RayTests.cs
ikeough: Good question. In the case of the vertex, I'd return the vertex. In the case of the ray origin in the plane of the face (an also inside the face's loop), I'd return the origin.

andrewheumann: Done.

test/Elements.Tests/RayTests.cs
```diff
@@ -1,126 +1,231 @@
...
+
+            Assert.True(ray1.Intersects(ray2, out result, true));
+        }
+
+        private static Vector3 Center(Triangle t)
+        {
+            return new Vector3[] { t.Vertices[0].Position, t.Vertices[1].Position, t.Vertices[2].Position }.Average();
+        }
+
+
```
ikeough: One small whitespace correction.

Issue #249, opened 2020-02-12
Hypar walls are not in the same place as Revit.


ikeough:
**Describe the bug**
Walls extracted from Revit appear to be extracted using one face of the wall to generate the Hypar `Wall`. In the attached image, you can see how the extracted walls do not align with the slab subtractions suggesting that either the inner face or the outer face of the Revit wall was used to create our extrusion as opposed to getting the center plane and extruding in both directions.

**Expected behavior**
Hypar walls extracted from Revit should be in exactly the same location.

**Screenshots**
![image](https://user-images.githubusercontent.com/1139788/74387654-f565ec80-4dad-11ea-85e1-d723818066b1.png)

wynged:
Should be solved by #251 

here is the new wall appearance
![image](https://user-images.githubusercontent.com/5872187/74483238-0f71fe80-4e84-11ea-8b3e-8624ea1f19ca.png)

PR #251, opened 2020-02-13
Dynamo extraction improvements

wynged:
BACKGROUND:
Needed to improve how walls were extracted

DESCRIPTION:
improve the wall extraction so that it reliably places walls in the correct location, addressing #249 

TESTING:
run `dotnet publish` and use the dynamo nodes to extract walls
  
FUTURE WORK:
Column extraction is still not complete

Here's what the walls look like now
image


src/Dynamo/src/HyparDynamo/Hypar.cs
```diff
@@ -2,55 +2,66 @@
...
         /// <summary>
-        /// gets the walls
+        /// Convert Revit walls to Elements.WallsByProfiles for use in Hypar models
         /// </summary>
-        /// <param name="wall">The walls to be exported</param>
-        /// <returns name="Hypar.Wall">The Hypar Wall element </param>
-        public static Elements.Wall[] FromRevitWall( this Revit.Elements.Wall RevitWall) 
+        /// <param name="RevitWall">The walls to be exported</param>
+        /// <returns name="WallByProfile">The Hypar walls</param>
+        public static Elements.WallByProfile[] FromRevitWall(this Revit.Elements.Wall RevitWall)
```
ikeough: Can we capitalization consistent here ex: `revitWall` and use the xml to override it?

src/Dynamo/src/HyparDynamo/Hypar.cs
```diff
@@ -2,55 +2,66 @@
...
     {
         /// <summary>
-        /// gets the walls
+        /// Convert Revit walls to Elements.WallsByProfiles for use in Hypar models
         /// </summary>
-        /// <param name="wall">The walls to be exported</param>
-        /// <returns name="Hypar.Wall">The Hypar Wall element </param>
-        public static Elements.Wall[] FromRevitWall( this Revit.Elements.Wall RevitWall) 
+        /// <param name="RevitWall">The walls to be exported</param>
+        /// <returns name="WallByProfile">The Hypar walls</param>
```
ikeough: "The Hypar walls." (period)

ikeough: Go through all these comments and add periods.

src/Dynamo/src/HyparDynamo/Hypar.cs
```diff
@@ -2,55 +2,66 @@
...
     }
 
-    public static class Floor {
-        public static Elements.Floor[] FromRevitFloor(this Revit.Elements.Floor RevitFloor) {
+    public static class Floor
+    {
+        /// <summary>
+        /// Convert Revit walls to Elements.Floors for use in Hypar models
+        /// </summary>
+        /// <param name="RevitFloor">The floors to be exported</param>
```
ikeough: This is a single floor correct?

src/Dynamo/src/HyparDynamo/Hypar.cs
```diff
@@ -64,16 +75,20 @@ public static void WriteJson(string filePath, Elements.Model model)
...
             }
-            catch (Exception ex) {
+            catch (Exception ex)
+            {
                 throw new Exception(ex.Message);
             }
         }
 
-        public static Elements.Model ModelFromElements(IList<object> elements) {
+        public static Elements.Model ModelFromElements(IList<object> elements)
```
ikeough: Would this not just be `FromElements` so that it reads as `Model.FromElements`?

src/Revit/RevitHyparTools/Create.Wall.cs
```diff
@@ -29,13 +27,18 @@ public static Elements.Wall[] WallsFromRevitWall(ADSK.Wall wall, Document doc)
...
+            var profiles = Utils.GetScaledProfilesOfFace(wallPlane);
 
             var centerline = (wall.Location as LocationCurve).Curve;
             var line = new ElemGeom.Line(centerline.GetEndPoint(0).ToVector3(), centerline.GetEndPoint(1).ToVector3());
+            line = Utils.ScaleLineFtToMeters(line);
+            
 
-            var walls = profiles.Select(p => new WallByProfile(line, p, wall.Width));
+            var walls = profiles.Select(p => new WallByProfile(Utils.ReverseProfile(p),
+                                                               wall.Width * Utils.FT_TO_METER_FACTOR,
```
ikeough: Can we use the built in `Elements.Units` class to get these scaling factors? I believe the method would be `FeetToMeters(...)`.

src/Revit/RevitHyparTools/WallByProfile.cs
```diff
@@ -1,40 +1,34 @@
...
-                                                 isElementDefinition,
-                                                 id != default(Guid) ? id : Guid.NewGuid(),
-                                                 name)
-        {
-            Profile = profile;
-            CenterLine = centerLine;
-            Thickness = thickness;
+        public WallByProfile(Profile @profile, double @thickness, Line @centerline, Transform @transform=null, Material @material=null, Representation @representation=null, bool @isElementDefinition=false)
+            : base(transform != null ? transform : new Transform(),
+                   material != null ? material : BuiltInMaterials.Default,
```
ikeough: I think our walls use `BuiltInMaterials.Concrete` by default. If not, they should.

ikeough: Why don't we just add this to Elements proper so anybody using elements can take advantage of it?

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Hypar.Revit
+{
+    public static class Utils
+    {
+        public const double FT_TO_METER_FACTOR = 0.3048;
+        public static Elements.Geometry.Profile ScaleProfileFtToMeters(Elements.Geometry.Profile profile) {
+            var transform = new Elements.Geometry.Transform();
+            transform.Scale(FT_TO_METER_FACTOR);
+            profile.Transform(transform);
```
ikeough: This transforms the profile "in place" meaning that the supplied profile will be transformed. Is that the expected behavior? It seems from the rest of the method that you want to return a new transformed profile instead. In that case, I'd do `transform.OfProfile(profile)` which will return a transformed copy of the original profile.

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,89 @@
...
+                if (!polyIsInnerLoop)
+                {
+                    polygonLoopDict.Add(polygon, new List<Polygon>());
+                }
+            }
+
+            return polygonLoopDict;
+        }
+
+        public static Polygon CurveLoopToPolygon(CurveLoop cL)
```
ikeough: This is a perfect opportunity for an extension method `CurveLoop.ToPolygon`.

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return new Polygon(cL.Select(l => l.GetEndPoint(0).ToVector3()).ToList());
+        }
+
+        public static PlanarFace[] GetMostLikelyTopFacesOfSolid(Solid solid)
+        {
+            var faces = new List<PlanarFace>();
+            foreach (PlanarFace face in solid.Faces)
+            {
+                if (face.FaceNormal.DotProduct(XYZ.BasisZ) > 0.85 && face.FaceNormal.DotProduct(XYZ.BasisZ) <= 1)
```
ikeough: Let's set this 0.85 as a default to a parameter on the method like `double threshold`. Then you can add some nice comments about what the threshold means.

src/Revit/RevitHyparTools/WallByProfile.cs
```diff
@@ -1,40 +1,34 @@
...
+            this.Centerline = @centerline;
         }
 
         public override void UpdateRepresentations() {
             this.Representation.SolidOperations.Clear();
-            // The wall is a simple extrusion of the profile
-            this.Representation.SolidOperations.Add(new Extrude(this.Profile, this.Thickness, Profile.Perimeter.Normal(), false));
+
+            // to ensure the correct direction, we find the direction form a point on the polygon to the vertical plane of the centerline
+            var point = Profile.Perimeter.Vertices.First();
```
ikeough: I will require good excuses to introduce Linq stuff like `First`, if only because Linq tends to be abused. Is there a reason we can't do `Profile.Perimeter.Vertices[0]`?

src/Dynamo/src/HyparDynamo/Hypar.cs
wynged: yes, it is one floor, lacing handled by dynamo.  I've removed the "s"

src/Revit/RevitHyparTools/Create.Wall.cs
wynged: yea, didn't know that was there.  done.

src/Revit/RevitHyparTools/WallByProfile.cs
wynged: what do you mean Linq tends to be abused?  I think it's concise, clear and a pleasure to use.  what are you looking out for?
Of course changing to getting by index will work just as well.

src/Revit/RevitHyparTools/Utils.cs
wynged: sounds good to me.  I assume I should make a new class like "CurveLoopExtensions.cs" that will contain this?  it won't be a very discoverable extension method if you have to know to do `using RevitHyparTools.Utils` to get access to that CurveLoop method.

src/Revit/RevitHyparTools/Utils.cs
wynged: Good idea.
Permission to make the threshhold parameter in degrees and do `Math.cos(threshold*PI/180)` inside the function so that users don't have to do that themselves?

src/Revit/RevitHyparTools/Utils.cs
wynged: you are right, I don't want to modify the original transform.

src/Revit/RevitHyparTools/WallByProfile.cs
ikeough: Let's leave this usage for now. A conversation lazy evaluation is outside the scope of this pr.

src/Revit/RevitHyparTools/Utils.cs
ikeough: Permission granted :)

src/Dynamo/src/HyparDynamo/Hypar.cs
wynged: I can't find a way to overide a parameter name.  only class and namespaces names.  
It appears that camelCase is the node input naming standard in dynamo, so we should be good with just fixing this one.

src/Revit/RevitHyparTools/Utils.cs
wynged: never mind, I remembered we alreayd have a RevitExtensions class, this will go in there.

src/Dynamo/src/HyparDynamo/Hypar.cs
wynged: this is the last status I could find regarding this feature in Dynamo
https://github.com/DynamoDS/Dynamo/issues/6360

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+
+        /// <summary>
+        /// Analyzes all of the PlanarFaces of a Revit solid and returns those that 
+        /// face "up" within a certain threshold.
+        /// </summary>
+        /// <param name="solid">The revit Solid</param>
+        /// <param name="verticalThreshold">The angle (in degrees) that represents the threshold for a face to be considered facing up.  A completely horizontal face will have an angle of 0.  A face that does not face up or down at all will have an angle of 90.</param>
+        public static PlanarFace[] GetMostLikelyTopFacesOfSolid(Solid solid, double verticalThreshold = 30)
```
ikeough: Extension method `Solid.GetMostLikelyTopFaces(...)`

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
...
+        public static Elements.Geometry.Profile ReverseProfile(Elements.Geometry.Profile profile)
+        {
+            var perimeter = profile.Perimeter.Reversed();
+            var voids = profile.Voids.Select(v => v.Reversed());
+
+            return new Elements.Geometry.Profile(perimeter, voids.ToList(), Guid.NewGuid(), "");
+        }
+
+        // Revit stores lengths in ft, but Hypar's standard is meters, so we need to scale the faces when we convert them to profiles
+        public static Elements.Geometry.Profile[] GetScaledProfilesOfFace(PlanarFace f)
```
ikeough: Extension method: `PlanarFace.GetProfiles(...)` The fact that the profiles are scaled is implied in the conversion to Hypar geometry.

src/Revit/RevitHyparTools/Create.Floor.cs
```diff
@@ -0,0 +1,47 @@
...
+            foreach (var profile in profiles)
+            {
+                var zMove = profile.Perimeter.Vertices.Max(v => v.Z);
+                var transform = new ElemGeom.Transform(0, 0, -zMove );
+
+                var zeroedProfile = transform.OfProfile(profile);
+
+                transform.Invert();
+                var floorThickness = thickness.HasValue ? Elements.Units.FeetToMeters(thickness.Value): Elements.Units.FeetToMeters(1);
+                // there appears to be a differen
```
ikeough: Don't leave me hanging.

src/Revit/RevitHyparTools/Create.Column.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static partial class Create {
+
+        public static Elements.Column ColumnFromRevitColumn(Revit.FamilyInstance column) {
+            if (column.Category.Id.IntegerValue != (int) BuiltInCategory.OST_StructuralColumns) {
+                throw new InvalidCastException("The incoming family instance is not a structural column");
+            }
+
+            var geom = column.get_Geometry(new Options());
+
+
```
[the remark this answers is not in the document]
wynged: done

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Autodesk.Revit.DB;
+using Elements.Geometry;
+
+namespace Hypar.Revit
+{
+    public static class Utils
```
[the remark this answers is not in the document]
wynged: makes sense, I'll distribute this class.

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        public static Elements.Geometry.Profile ScaleProfileFtToMeters(Elements.Geometry.Profile profile) {
+            var transform = new Elements.Geometry.Transform();
+            transform.Scale(Elements.Units.FeetToMeters(1));
+            var transformedProfile = transform.OfProfile(profile);
+            
+            return new Elements.Geometry.Profile(transformedProfile.Perimeter, transformedProfile.Voids, Guid.NewGuid(), transformedProfile.Name);
+        }
+        public static Elements.Geometry.Profile ReverseProfile(Elements.Geometry.Profile profile)
```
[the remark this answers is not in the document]
wynged: turns out, already exists!

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
...
+            var polygons = f.GetEdgesAsCurveLoops().Select(cL => cL.ToPolygon());
+
+            var polygonLoopDict = MatchOuterLoopPolygonsWithInnerHoles(polygons);
+
+            var profiles = polygonLoopDict.Select(kvp => new Elements.Geometry.Profile(kvp.Key, kvp.Value, Guid.NewGuid(), "Revit Profile"));
+            var scaledProfiles = profiles.Select(p => ScaleProfileFtToMeters(p));
+            return scaledProfiles.ToArray();
+        }
+
+        internal static Elements.Geometry.Line ScaleLineFtToMeters(Elements.Geometry.Line line)
```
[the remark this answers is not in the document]
wynged: I was anticipating needing to do this more than once, but it turns out it's only done it one place, so I'll just move this methods contents to that line.
Ee're scaling the centerline using a transform, could scale the individual vectors, would require a couple more lines.

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -7,5 +8,9 @@ public static class RevitExtensions {
         public static Vector3 ToVector3(this XYZ xyz) {
             return new Vector3(xyz.X, xyz.Y, xyz.Z);
         }
```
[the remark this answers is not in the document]
wynged: i've set it to format on save now.  I've got the default one with the C# tools in VSCode.  I'm also having to change my brain to prefer having lines with the just the open `{` which I got away from.  Interesting note that our TypeScript linter intentionally puts those opening `{` on the previous line.

src/Revit/RevitHyparTools/Utils.cs
wynged: great, done

src/Revit/RevitHyparTools/Create.Floor.cs
wynged: I'm sure the suspense is killing you....
story ending added.

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -7,5 +8,9 @@ public static class RevitExtensions {
         public static Vector3 ToVector3(this XYZ xyz) {
             return new Vector3(xyz.X, xyz.Y, xyz.Z);
         }       
+        public static Polygon ToPolygon(this CurveLoop cL)
```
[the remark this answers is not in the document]
wynged: done

src/Revit/RevitHyparTools/Utils.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Autodesk.Revit.DB;
+using Elements.Geometry;
+
+namespace Hypar.Revit
+{
+    public static class Utils
+    {
+        public static Elements.Geometry.Profile ScaleProfileFtToMeters(Elements.Geometry.Profile profile) {
```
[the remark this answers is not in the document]
wynged: done

src/Revit/RevitHyparTools/RevitExtensions.cs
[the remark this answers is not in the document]
ikeough: Right, Typescript follows javascript convention. C# convention has opening brackets on a new line.

src/Elements/Geometry/Profile.cs
```diff
@@ -70,6 +70,17 @@ public void Transform(Transform t)
...
+        /// <summary>
+        /// Return a new profile that is this profile scaled about the origin by the desired amount.
+        /// </summary>
+        public Elements.Geometry.Profile Scale(double amount)
+        {
+            var transform = new Elements.Geometry.Transform();
+            transform.Scale(amount);
+            var transformedProfile = transform.OfProfile(this);
+
+            return new Elements.Geometry.Profile(transformedProfile.Perimeter, transformedProfile.Voids, Guid.NewGuid(), transformedProfile.Name);
```
ikeough: Is this necessary? Can you just do `return transform.OfProfile(this)`?

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -1,11 +1,75 @@
...
-    public static class RevitExtensions {
-        public static Vector3 ToVector3(this XYZ xyz) {
+    public static class RevitExtensions
+    {
+        public static Vector3 ToVector3(this XYZ xyz)
+        {
             return new Vector3(xyz.X, xyz.Y, xyz.Z);
-        }       
+        }
+        public static Polygon ToPolygon(this CurveLoop curveLoop)
```
ikeough: Add one line of whitespace between methods.

src/Revit/RevitHyparTools/Create.Wall.cs
```diff
@@ -29,13 +26,20 @@ public static Elements.Wall[] WallsFromRevitWall(ADSK.Wall wall, Document doc)
             }
 
             var wallPlane = wallFace as PlanarFace;
-            var profiles = GetProfilesOfFace(wallPlane);
+            var profiles = wallPlane.GetProfiles();
 
             var centerline = (wall.Location as LocationCurve).Curve;
+
             var line = new ElemGeom.Line(centerline.GetEndPoint(0).ToVector3(), centerline.GetEndPoint(1).ToVector3());
```
ikeough: Can we just multiply the vectors here by the scale? Do we need the additional transform steps below? 
`var line = new ElemGeom.Line(centerline.GetEndPoint(0).ToVector3() * Units.FeetToMeters(1),...`

ikeough: I suggest this because matrix operations are expensive. We should avoid them where we can.

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -1,11 +1,75 @@
...
+            return new Polygon(curveLoop.Select(l => l.GetEndPoint(0).ToVector3()).ToList());
+        }
+
+        /// <summary>
+        /// Analyzes all of the PlanarFaces of a Revit solid and returns those that 
+        /// face "up" within a certain threshold.
+        /// </summary>
+        /// <param name="solid">The revit Solid</param>
+        /// <param name="verticalThreshold">The angle (in degrees) that represents the threshold for a face to be considered facing up.  A completely horizontal face will have an angle of 0.  A face that does not face up or down at all will have an angle of 90.</param>
+        public static PlanarFace[] GetMostLikelyTopFaces(this Solid solid, double verticalThreshold = 30)
```
ikeough: Do a review of these methods and give them the most limited access possible. For example, this method is called inside one other private method. Does it need to be public?

src/Revit/RevitHyparTools/Create.Column.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using Autodesk.Revit.DB;
+using Elements.Geometry;
+using GeometryEx;
```
ikeough: Are we still using GeometryEx?

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -1,11 +1,75 @@
...
 using Elements.Geometry;
 
-namespace Hypar.Revit 
+namespace Hypar.Revit
 {
-    public static class RevitExtensions {
-        public static Vector3 ToVector3(this XYZ xyz) {
+    public static class RevitExtensions
+    {
+        public static Vector3 ToVector3(this XYZ xyz)
```
ikeough: We can decide if we want to add the scaling into this conversion. This is great if it's a point, but not so great if it's considered as a vector, so you could make it optional like `ToVector3(this XYZ xyz, bool scaleToMeters = false)`. This would then simplify a lot of the code below where you're scaling stuff.

src/Elements/Geometry/Profile.cs
wynged: good point.  the only thin missing now is that `transform.OfProfile()` doesn't maintain the name of the profile, but maybe we don't think of profiles as elements that need names?

src/Revit/RevitHyparTools/RevitExtensions.cs
wynged: done

src/Revit/RevitHyparTools/Create.Column.cs
wynged: nope, removed unnecessary usings.

src/Revit/RevitHyparTools/RevitExtensions.cs
wynged: that sounds good to me.

src/Revit/RevitHyparTools/RevitExtensions.cs
wynged: There are lots of these that can be set to internal, is that a worthwhile limitation as well?

src/Revit/RevitHyparTools/Create.Wall.cs
wynged: okay, done.

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -1,11 +1,90 @@
...
+            {
+                if (face.FaceNormal.DotProduct(XYZ.BasisZ) > Math.Cos(Elements.Units.DegreesToRadians(verticalThreshold)) && face.FaceNormal.DotProduct(XYZ.BasisZ) <= 1)
+                {
+                    faces.Add(face);
+                }
+            }
+            return faces.ToArray();
+        }
+
+        private static Polygon ToPolygon(this CurveLoop curveLoop)
```
ikeough: Can we add a scaling parameter here as well? Then push it down to the call to `ToVector3()`? This will save us from the additional scale operation you do above in `GetProfiles(...)`

src/Elements/Geometry/Profile.cs
```diff
@@ -70,6 +70,16 @@ public void Transform(Transform t)
                 this.Voids[i].Transform(t);
             }
         }
```
ikeough: Whitespace.

src/Revit/RevitHyparTools/RevitExtensions.cs
wynged: sounds good.

src/Revit/RevitHyparTools/RevitExtensions.cs
wynged: ?

src/Revit/RevitHyparTools/RevitExtensions.cs
```diff
@@ -1,11 +1,90 @@
...
+    internal static class RevitExtensions
+    {
+        /// <summary>
+        /// Convert the Revit XYZ to an Elements Vector3, with an option to scale the 
+        /// Vector3 from feet to meters.  This scaling should be done with any XYZ that represents
+        /// a geometric property, but should maybe be avoided for other vectors.
+        /// </summary>
+        internal static Vector3 ToVector3(this XYZ xyz, bool scaleToMeters = false)
+        {
+            var scale = 1.0;
```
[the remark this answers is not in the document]
wynged: very clean.  done

src/Elements/Geometry/Profile.cs
```diff
@@ -31,7 +31,7 @@ public Profile(Polygon perimeter) : base(Guid.NewGuid(), null)
         /// <summary>
         /// Get a new profile which is the reverse of this profile.
         /// </summary>
-        public Profile Reversed()
```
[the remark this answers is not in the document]
wynged: it is Reversed() ...

src/Elements/Geometry/Profile.cs
wynged: I believe this is fixed

src/Revit/RevitHyparTools/RevitExtensions.cs
ikeough: Yes. Every public field or method is a field that you have to support. They become part of the public API. Let’s start by hiding them, then expose them when we get feedback that they’re needed.

wynged: gotcha, makes sense

PR #252, opened 2020-02-14
fix determinant calculation

wynged:
BACKGROUND:
transforms were not inverting when they should be able to

DESCRIPTION:
It fixes the determinant calculation (which had a typo) so that the `Inverse()` method doesn't throw erroneous errors

TESTING:
There is a new test called Involutory_Transform_Inverse
  
FUTURE WORK:
maybe a dedicated Determinant test

### DIFF
--- a/src/Elements/Geometry/Matrix.cs
+++ b/src/Elements/Geometry/Matrix.cs
@@ -429,7 +429,7 @@ namespace Elements.Geometry
         /// </summary>
         public double Determinant()
         {
-            return m11 * (m22 * m33 - m22 * m32)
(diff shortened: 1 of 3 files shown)


src/Elements/Geometry/Matrix.cs
```diff
@@ -429,7 +429,7 @@ public Matrix Transpose()
         /// </summary>
         public double Determinant()
```
mattc: Throw in a basic determinant test.

wynged: done!

PR #258, opened 2020-02-20
Make solid operations use INotifyPropertyChanged.

ikeough:
BACKGROUND:
While testing ray intersections with solids, we noticed that solid intersections were calling `GetSolid()`, which is an expensive operation because it creates the extrude, sweep, lamina, etc. We needed a way to compute the solid only when properties were changed that would affect the solid, and to cache that solid for future use.

DESCRIPTION:
- This PR implements `INotifyPropertyChanged` for all solid operation types.
- It adds `PreConstruct` and `PostConstruct` methods to auto-generated classes to allow developers to inject behavior into these classes. 
- It uses the `PostConstruct` method on solid operations to hook the property change notifications and compute a new solid which is cached.
- It adds the `SolidOperation.Solid` property which returns the cached solid.

### DIFF
--- a/src/Elements/Analysis/AnalysisMesh.cs
+++ b/src/Elements/Analysis/AnalysisMesh.cs
@@ -20,7 +20,7 @@ namespace Elements.Analysis
         private Grid2d _grid;
         private Func<Vector3, double> _analyze;
 
-        private List<Tuple<Polygon, double>> _results;
+        private List<(Polygon cell, double result)> _results;
         private double _min;
         private double _max;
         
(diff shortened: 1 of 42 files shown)


src/Elements/Analysis/AnalysisMesh.cs
```diff
@@ -144,5 +131,35 @@ public void Tessellate(ref Mesh mesh)
                 }
             }
         }
+
+        /// <summary>
+        /// Compute a value for each grid cell.
+        /// </summary>
+        public void Analyze()
+        {
+            this._results = new List<Tuple<Polygon, double>>();
```
andrewheumann: May I suggest using C# 7's named tuple types so we don't have to contend with .Item1, .Item2 etc? So `new List<(Polygon cell, double analysisResult)>` or something?

ikeough: That’s a great idea.

ikeough: Done.

PR #260, opened 2020-02-20
Support for grids from non-xy-parallel polygons

andrewheumann:
BACKGROUND:
- @ikeough was using wall polygons and expected that he could create a grid from a polygon not aligned parallel to the XY plane. This is a reasonable assumption 

DESCRIPTION:
- If a user has not supplied a transform to one of the Grid2d constructors asking for a polygon, one will be constructed automatically in the plane of that polygon. 
- if the polygon is parallel to the XY plane, but above or below it, it will only generate a translation transform, so that unexpected orientations do not arise 
- if a transform IS provided, the geometry will be tested to ensure all geometry lies in the XY plane after the transform is applied. 

TESTING:
- create a Grid2d from a planar polygon not located in the XY plane
  
FUTURE WORK:
- Perhaps other mechanisms than a transform to supply the grid alignment would be useful to a user  

COMMENTS:


andrewheumann:
realizing I introduced a bug in Polyline.Plane()... fixing

andrewheumann:
turns out it wasn't my mistake at all! There was a bug in `Transform.OfPlane()`. Fixed now!

src/Elements/Geometry/Vector3.cs
```diff
@@ -503,20 +503,18 @@ public static bool AreCoplanar(this IList<Vector3> points)
         /// <returns></returns>
         public static Transform ToTransform(this IList<Vector3> points)
         {
-            var a = new Vector3(points[1].X - points[0].X,
-                                points[1].Y - points[0].Y,
-                                points[1].Z - points[0].Z).Unitized();
-
+            //var a = new Vector3(points[1].X - points[0].X,
```
ikeough: Delete this commented code.

src/Elements/Geometry/Transform.cs
```diff
@@ -225,7 +225,8 @@ public Line OfLine(Line line)
         /// <returns>A new plane transformed by this transform.</returns>
         public Plane OfPlane(Plane plane)
         {
-            return new Plane(OfPoint(plane.Origin), OfVector(plane.Normal));
+            var tempXform = new Transform(Vector3.Origin, this.XAxis, this.ZAxis);
```
ikeough: My understanding of what this fix does is that, by creating a separate transform with zero translation, and transforming the initial plane's vector by that, we don't accidentally accumulate the translate of the the transform onto the vector. Is that correct? If so, nice!

src/Elements/Geometry/Vector3.cs
ikeough: This is a nice cleanup. This code was probably written before we could subtract vectors.

src/Elements/Spatial/Grid2d.cs
```diff
@@ -134,6 +138,11 @@ public Grid2d(IList<Polygon> boundaries, Transform transform = null)
             toGrid.Invert();
 
             var transformedBoundaries = toGrid.OfPolygons(boundaries);
+            var allVerticesZ = transformedBoundaries.SelectMany(b => b.Vertices).Select(v => v.Z);
```
ikeough: I've talked a bit with @wynged about not wanting Linq if we can avoid it, but especially in geometry code. There are performance penalties to Linq that are subtle: https://stackoverflow.com/questions/14893924/for-vs-linq-performance-vs-future. I am happy to be proved wrong about this. Can we do this as a set of nested for loops with an early out?

test/Elements.Tests/Grid2dTests.cs
```diff
@@ -215,5 +215,41 @@ public void GetSeparators()
                 { "V", vLines }
             };
         }
+
```
ikeough: Remove extra whitepsace.

test/Elements.Tests/PolygonTests.cs
```diff
@@ -647,16 +647,16 @@ public void Reverse()
...
 
             _output.WriteLine("Normals:");
             _output.WriteLine(n1.ToString());
             _output.WriteLine(n2.ToString());
             _output.WriteLine("");
 
             // Check that the start and end have opposing normals.
-            var dot = n1.Dot(n2);
+            var dot = n1.Normal.Dot(n2.Normal);
             _output.WriteLine(dot.ToString());
```
ikeough: We can remove test write lines unless they're really needed forever.

src/Elements/Spatial/Grid2d.cs
```diff
@@ -134,6 +138,11 @@ public Grid2d(IList<Polygon> boundaries, Transform transform = null)
...
+            if(allVerticesZ.Any(z => !z.ApproximatelyEquals(0))) {
+                throw new Exception("After transform, this polygon was not in the XY Plane. Please ensure that all your geometry as well as any provided transform all lie in the same plane.");
```
ikeough: When I throw an exception inside a constructor, I like to say something like:
> The Grid2d could not be constructed. After transform, the polygon was not in the XY plane...

This enables the developer to read the exception method an known exactly where the exception was thrown without having to read the stack trace. It's also nice for our developers who are less familiar with stack traces and would just like the computer to tell them exactly what went wrong in human terms.

src/Elements/Spatial/Grid2d.cs
```diff
@@ -120,12 +120,16 @@ public Grid2d(IList<Polygon> boundaries, Transform transform = null)
...
-            }
-
-            if (!transform.ZAxis.IsParallelTo(Vector3.ZAxis))
-            {
-                throw new ArgumentException("Currently transforms that are not parallel to the XY Plane are not supported.");
+                //if no transform is supplied, calculate one from the normal. 
+                var planeTransform = boundaries.First().Vertices.ToTransform();
+                if (Math.Abs(planeTransform.ZAxis.Dot(Vector3.ZAxis)).ApproximatelyEquals(1)) // we're xy parallel
+                {
+                    transform = new Transform(planeTransform.Origin);
```
ikeough: This is an example of where a note could help. I think what's happening here is that you're saying, if the transform is _almost_ the default, then force it to be the default. If that's correct, it's not super obvious and a note could help.

ikeough: You could also do this as a ternary which may increase the legibility:
```c#
transform = Math.Abs(planeTransform.ZAxis.Dot(Vector3.ZAxis)).ApproximatelyEquals(1) ?
  new Transform(planeTransform.Origin) :
  planeTransform;

ikeough: Finally, I'm not sure if you want to just force this back to default, as `Vertices.Transform(...)` creates a transform with an X axis along the vector between the first and sector point. Forcing this back to the default sets the X axis to the world X axis.

src/Elements/Geometry/Transform.cs
andrewheumann: Exactly! I'll leave a comment to that effect.

src/Elements/Spatial/Grid2d.cs
andrewheumann: done

src/Elements/Spatial/Grid2d.cs
andrewheumann: In this case this is the desired behavior. This only occurs when a user has *not* explicitly provided an orientation, only geometry. If the curves are up in space somewhere, we take our chances (there's no obvious implicit alignment); if they are XY-parallel, then we use the default which is a world-parallel grid.

PR #261, opened 2020-02-21
first stab at meshSolid implementation

andrewheumann:
BACKGROUND:
- In order to support the Isovist workflow using a mesh as representation rather than a Lamina, @ikeough suggested: 
>A more flexible solution might be to create another type of SolidOperation that holds a mesh.

I am fairly new to the solid code so I expect I will have made some mistakes or invalid assumptions in my implementation. If this seems like the wrong interpretation of the above idea I am happy to scrap all of this in favor of a different direction. 

DESCRIPTION:
- Implement a new solid class called MeshSolid that wraps a mesh.
- Implement a new Solid from Mesh generator on the Kernel

TESTING:
  
FUTURE WORK:
- It seems like the moment this "solid" gets tesselated back into a mesh, we're doing a bunch of unnecessary work. Maybe there's a way to have the MeshSolid have special behavior on tesselation?

### DIFF
--- a/src/Elements/Geometry/Kernel.cs
+++ b/src/Elements/Geometry/Kernel.cs
@@ -1,3 +1,4 @@
+using System;
 using Elements.Geometry.Solids;
 
 namespace Elements.Geometry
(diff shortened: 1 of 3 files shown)


ikeough:
Can we start by restating the purpose of this PR? My understanding is that we want to be able to create a custom analysis mesh for things like visibility analysis where the mesh shouldn't be a square grid. An easier way of achieving that goal would be to allow a developer to override the `Tessellate` method in their own extension of an analysis mesh.

But there was a second reason for our discussion which was to support meshes. Meshes aren't necessarily solids, so I hesitate to roll them into the `SolidOperation` code. Perhaps we need to make `Representation` a list of viable representation types including `Solid`, and `Mesh`.

ikeough:
Closing this in favor of https://github.com/hypar-io/Elements/pull/268.

PR #268, opened 2020-03-09
Import Mesh Element

ikeough:
BACKGROUND:
To support developers who want to define elements from existing mesh assets, we need a way of importing those assets into Hypar as Elements. Additionally, developers who wish to construct Element types which implement `ITessellate` and "roll their own" mesh need more access to methods on `Mesh` that are currently private.

DESCRIPTION:
- This PR adds the `ImportMeshElement` type which is constructed by providing the path to an STL file on disk.
- This PR makes `Mesh.AddVertex(...)` and `Mesh.AddTriangle(...)` public.
- This PR switches edge rendering to be _disabled_ by default.

image


ikeough:
See https://github.com/hypar-io/Hypar/pull/149.

ikeough:
@mattc @andrewheumann We've now received a couple of requests to be able to edit meshes directly. I believe that this PR gives a developer the ability to construct a mesh and use that mesh as part of an `ITessellate` implementation. This should solve for Andrew's case of wanting to build a bespoke mesh for isovists.

For cases where we want to use an existing mesh already encoded as an STL (and other formats in the future), like when placing manufacturer's content, I've added a new Element type called `ImportMeshElement`. I'd like some feedback on whether you think this is an acceptable design to allow people to ingest existing meshes.

src/Elements/Geometry/Mesh.cs
```diff
@@ -216,6 +218,74 @@ public Mesh()
...
+        /// <param name="stlPath">The path to the STL file.</param>
+        /// <param name="unit">The length unit used in the file.</param>
+        /// <returns></returns>
+        public static Mesh FromSTL(string stlPath, LengthUnit unit = LengthUnit.Millimeter)
+        {
+            List<Vertex> vertexCache = new List<Vertex>();
+            var mesh = new Mesh();
+
+            var conversion = 1.0;
+            switch(unit)
```
andrewheumann: it feels like this should live in `Units` — e.g. `Units.GetConversionFactor(LengthUnit from, LengthUnit to)` or similar. Or alternately/additionally, a `public Func<double> Units.GetConverter(LengthUnit from, LengthUnit to)` so I could just:
```
var converter = Units.GetConverter(unit, LengthUnit.Meters);
var x =  converter(double.Parse(splits[1]));
etc.
```

src/Elements/ImportMeshElement.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Elements
+{
+    /// <summary>
+    /// An element definition whose representation is provided by an imported mesh like an STL.
+    /// </summary>
+    /// <example>
+    /// [!code-csharp[Main](../../test/Elements.Tests/ImportMeshElementTests.cs?name=example)]
+    /// </example>
+    [UserElement]
+    public class ImportMeshElement : GeometricElement, ITessellate
```
andrewheumann: Is it possible now to define a custom element type that wraps a mesh without making changes to `Elements`? If not, is it possible to define a general MeshElement type (e.g. "Model Curve" but for meshes) that can wrap a mesh that may or may not be imported?

mattc:
Related to what Andrew is saying (but possibly the opposite) I wonder if we should mark this `final`. We shouldn't have a bunch of elements that can only ever contain imported meshes. This is too much of an implementation detail to be part of the interface, and it'd be basically impossible to evolve the geometric representation into something other than a mesh in certain cases.

(Otherwise I'm comfortable with this PR)

PR #273, opened 2020-03-24
winding-based polygon contains, support for holes + xforms for ray intersection

### LINKED ISSUE #257: Ray/Solid intersections do not respect an element's transform or holes.
**Describe the bug**
When intersecting a `Ray` with a `Solid`, the resulting intersections do not respect the transform of the solid's element, and the rays do not traverse through holes in the solid.

**This issue requires the changes in #258 to test.**

**To Reproduce**
1. Run `RayTests.RayShadowTest()`

**Expected behavior**
Rays that intersect solids in elements with transforms will respect the transform. Rays that go through holes in a solid will not cause an intersection.

**Screenshots**
Rays properly intersect a solid in an element without a transform, with the exception of holes where the rays intersect on the face containing the hole.
image
Rays do not properly intersect a solid in an element with a transform.
image

**The Problem with Holes**
The `Ray.Intersects(Face face,...)` API currently uses transform manipulation and clipper operations to test test whether a point is inside a polygon. This involves hundreds of multiplication operations per face. There is an easier and much faster way: http://geomalgorithms.com/a03-_inclusion.html.

andrewheumann:
BACKGROUND:
- Fixes #257 
- Attempts and fails to improve raycasting performance by implementing a Winding Number based polygon contains method. 

DESCRIPTION:
- Adds a `Ray.Intersects(GeometricElement)` method that accounts for the element's transform
- Adds a `Profile.Contains` method that can do region inclusion taking into account holes/voids
- Changes `Ray` to a Struct
- Updates `Ray.Intersects(Face)` to use the new hole-aware containment method
- Adds `Ray.Intersects(Line)`

TESTING:
- Run the updated `RayTests.RayShadowTest`
- Run the `VectorTests.ProfileContains` test (tests 100 points against a complex profile and ensures their inclusion results match those in Rhino / GH).
  
FUTURE WORK:
- The performance of the `Ray.Intersects()` operations could likely still be improved. 
- The code for the winding number stuff introduces a bunch of methods for that purpose only to the Line and Polygon classes. It might be worth consolidating them somehow to avoid cluttering up the classes with single-purpose methods.
- The test for ProfileContains could probably be less obscure.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/273)

### DIFF
--- a/src/Elements/Geometry/Containment.cs
+++ b/src/Elements/Geometry/Containment.cs
@@ -2,7 +2,7 @@
 namespace Elements.Geometry
 {
     /// <summary>
-    /// Represents the state of containment of a point relative to an enclosing polygon, profile, or volume
+    /// Represents the state of containment of a point relative to an enclosing polygon, profile, or solid.
     /// </summary>
     public enum Containment
     {
--- a/src/Elements/Geometry/Containment.cs
+++ b/src/Elements/Geometry/Containment.cs
@@ -1,0 +1,30 @@
+﻿using System;
+namespace Elements.Geometry
+{
+    /// <summary>
+    /// Represents the state of containment of a point relative to an enclosing polygon, profile, or volume
+    /// </summary>
+    public enum Containment
+    {
+        /// <summary>
+        /// The point lies entirely outside.
+        /// </summary>
+        Outside,
+        /// <summary>
+        /// The point lies entirely inside.
+        /// </summary>
+        Inside,
+        /// <summary>
+        /// The point lies exactly or nearly at an edge.
+        /// </summary>
+        CoincidesAtEdge,
+        /// <summary>
+        /// The point lies exactly or nearly at a vertex.
+        /// </summary>
+        CoincidesAtVertex,
+        /// <summary>
+        /// The point lies exactly or nearly at a face.
+        /// </summary>
+        CoincidesAtFace // currently not in use
+    }
+}
--- a/src/Elements/Geometry/Line.cs
+++ b/src/Elements/Geometry/Line.cs
@@ -435,5 +435,45 @@ namespace Elements.Geometry
         {
             return new[] { this.Start, this.End };
         }
+
+        #region WindingNumberCalcs
+        internal Position RelativePositionOf(Vector3 location)
+        {
+            double positionCalculation =
+                (End.Y - Start.Y) * (location.X - Start.X) -
+                (location.Y - Start.Y) * (End.X - Start.X);
+
+            if (positionCalculation > 0) { return Position.Left; }
+
+            if (positionCalculation < 0) { return Position.Right; }
+
+            return Position.Center;
+        }
+
+        internal bool AscendingRelativeTo(Vector3 location)
+        {
+            return Start.X <= location.X;
+        }
+
+        internal bool LocationInRange(Vector3 location, Orientation orientation)
+        {
+            if (orientation == Orientation.Ascending) return End.X > location.X;
+
+            return End.X <= location.X;
+        }
+        internal enum Position
+        {
+            Left,
+            Right,
+            Center
+        }
+
+        internal enum Orientation
+        {
+            Ascending,
+            Descending
+        }
+        #endregion
+
     }
 }
--- a/src/Elements/Geometry/Line.cs
+++ b/src/Elements/Geometry/Line.cs
@@ -135,7 +135,7 @@ namespace Elements.Geometry
         public bool Intersects(Plane p, out Vector3 result)
         {
             var rayIntersects = new Ray(Start, Direction()).Intersects(p, out Vector3 location, out double t);
-            if(rayIntersects && t <= Length())
+            if (rayIntersects && t <= Length())
             {
                 result = location;
                 return true;
@@ -443,9 +443,15 @@ namespace Elements.Geometry
                 (End.Y - Start.Y) * (location.X - Start.X) -
                 (location.Y - Start.Y) * (End.X - Start.X);
 
-            if (positionCalculation > 0) { return Position.Left; }
+            if (positionCalculation > 0)
+            {
+                return Position.Left;
+            }
 
-            if (positionCalculation < 0) { return Position.Right; }
+            if (positionCalculation < 0)
+            {
+                return Position.Right;
+            }
 
             return Position.Center;
         }
@@ -457,10 +463,14 @@ namespace Elements.Geometry
 
         internal bool LocationInRange(Vector3 location, Orientation orientation)
         {
-            if (orientation == Orientation.Ascending) return End.X > location.X;
+            if (orientation == Orientation.Ascending)
+            {
+                return End.X > location.X;
+            }
 
             return End.X <= location.X;
         }
+
         internal enum Position
         {
             Left,
--- a/src/Elements/Geometry/Polygon.cs
+++ b/src/Elements/Geometry/Polygon.cs
@@ -54,7 +54,7 @@ namespace Elements.Geometry
         /// <summary>
         /// Tests if the supplied Vector3 is within this Polygon, using a 2D method. 
         /// </summary>
-        /// <param name="vector">The position to test</param>
+        /// <param name="vector">The position to test.</param>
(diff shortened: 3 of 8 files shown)


src/Elements/Geometry/Ray.cs
```diff
@@ -80,6 +80,35 @@ public bool Intersects(Triangle tri, out Vector3 result)
...
+        public bool Intersects(GeometricElement element, out List<Vector3> result)
+        {
+            List<Vector3> resultsOut = new List<Vector3>();
+            var transformFromElement = new Transform(element.Transform);
+            transformFromElement.Invert();
+            var transformToElement = new Transform(element.Transform);
+            var transformMinusTranslation = new Transform(transformFromElement);
+            transformMinusTranslation.Move(-1 * transformMinusTranslation.Origin); // 
+            var transformedRay = new Ray(transformFromElement.OfPoint(Origin), transformMinusTranslation.OfVector(Direction));
+            //TODO: handle voids
```
ikeough: Is this still `TODO`?

src/Elements/Geometry/Line.cs
```diff
@@ -435,5 +435,45 @@ internal override IList<Vector3> RenderVertices()
...
+            return Start.X <= location.X;
+        }
+
+        internal bool LocationInRange(Vector3 location, Orientation orientation)
+        {
+            if (orientation == Orientation.Ascending) return End.X > location.X;
+
+            return End.X <= location.X;
+        }
+        internal enum Position
```
ikeough: whitespace

src/Elements/Geometry/Line.cs
```diff
@@ -435,5 +435,45 @@ internal override IList<Vector3> RenderVertices()
...
         }
+
+        #region WindingNumberCalcs
+        internal Position RelativePositionOf(Vector3 location)
+        {
+            double positionCalculation =
+                (End.Y - Start.Y) * (location.X - Start.X) -
+                (location.Y - Start.Y) * (End.X - Start.X);
+
+            if (positionCalculation > 0) { return Position.Left; }
```
ikeough: Put all blocks on separate row:
```
if (positionCalculation > 0)
{
...
}
```

src/Elements/Geometry/Line.cs
```diff
@@ -435,5 +435,45 @@ internal override IList<Vector3> RenderVertices()
...
+        }
+
+        internal bool AscendingRelativeTo(Vector3 location)
+        {
+            return Start.X <= location.X;
+        }
+
+        internal bool LocationInRange(Vector3 location, Orientation orientation)
+        {
+            if (orientation == Orientation.Ascending) return End.X > location.X;
```
ikeough: Always use brackets. See message above about blocks.

src/Elements/Geometry/Polygon.cs
```diff
@@ -56,6 +56,55 @@ public bool Contains(Vector3 vector)
...
+        // Adapted from https://stackoverflow.com/questions/46144205/point-in-polygon-using-winding-number/46144206
+        internal static bool Contains2D(IEnumerable<Line> segments, Vector3 location)
+        {
+            int windingNumber = 0;
+
+            foreach (var edge in segments)
+            {
+                // check for coincidence with edge vertices
+                var toStart = location - edge.Start;
+                if (toStart.IsZero()) return true;
```
ikeough: See message above about brackets.

src/Elements/Geometry/Profile.cs
```diff
@@ -185,5 +185,21 @@ public bool Equals(Profile other)
             }
             return this.Perimeter.Equals(other.Perimeter);
         }
+
+        /// <summary>
+        /// Tests if a point is contained within this profile. Returns false for points that are outside of the profile (or within voids). 
+        /// </summary>
+        /// <param name="point">The position to test.</param>
+        /// <returns>True if the point is within the profile.</returns>
+        public bool Contains2D(Vector3 point)
```
ikeough: The name of this method is suspect. What would `Contains3D` do?

src/Elements/Geometry/Ray.cs
andrewheumann: Yes — it is contingent on support for void Solids in elements in the first place!

andrewheumann:
These changes additionally:
- Define a new Containment enum to represent the status of a point within a region or volume (inside, outside, at vertex, at edge, or at face)
- pass out a containment value from all Contains functions
- Define a new constructor for Profile that can take an unordered set of Polygons and figure out which one is "outside".

src/Elements/Geometry/Containment.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+namespace Elements.Geometry
+{
+    /// <summary>
+    /// Represents the state of containment of a point relative to an enclosing polygon, profile, or volume
```
ikeough: `The containment of a point relative to an enclosing polygon, profile, or solid.`

src/Elements/Geometry/Ray.cs
```diff
@@ -128,12 +158,14 @@ internal bool Intersects(Face face, out Vector3 result)
...
             {
                 var boundaryPolygon = face.Outer.ToPolygon();
+                var voids = face.Inner?.Select(v => v.ToPolygon())?.ToList();
                 var transformToPolygon = new Transform(plane.Origin, plane.Normal);
                 var transformFromPolygon = new Transform(transformToPolygon);
                 transformFromPolygon.Invert();
-                var transformedPolygon = transformFromPolygon.OfPolygon(boundaryPolygon);
                 var transformedIntersection = transformFromPolygon.OfVector(intersection);
-                if (transformedPolygon.Contains(transformedIntersection) || transformedPolygon.Touches(transformedIntersection))
+                var profile = new Profile(boundaryPolygon, voids, default, "");
```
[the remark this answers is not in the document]
ikeough: I see that we're doing this to use some of the new methods that were created. I'll allow it. I wonder if the ray crossing method is faster for these situations.

src/Elements/Geometry/Ray.cs
```diff
@@ -80,6 +80,35 @@ public bool Intersects(Triangle tri, out Vector3 result)
...
+        /// <param name="result">The list of intersection results.</param>
+        /// <returns></returns>
+        public bool Intersects(GeometricElement element, out List<Vector3> result)
+        {
+            List<Vector3> resultsOut = new List<Vector3>();
+            var transformFromElement = new Transform(element.Transform);
+            transformFromElement.Invert();
+            var transformToElement = new Transform(element.Transform);
+            var transformMinusTranslation = new Transform(transformFromElement);
+            transformMinusTranslation.Move(-1 * transformMinusTranslation.Origin); //
```
ikeough: Is `Origin.Negate()` better here?

test/Elements.Tests/VectorTests.cs
```diff
@@ -84,34 +85,206 @@ public void DistanceToPlane()
...
+                {
+                    new Vector3(-3.28355342951497,-2.6460367928868,0),
+                    new Vector3(-2.87188183213404,-2.34663926751885,0),
+                    new Vector3(-3.20092267684535,-1.8942081060408,0),
+                    new Vector3(-3.61259427422628,-2.19360563140875,0)
+                })
+            };
+
+            var profile = new Profile(perimeter, voids, default, "");
+            var ptsToTest = new Dictionary<Vector3, bool>
```
ikeough: This seems like a really specific dictionary of points. Do we want to put these in a JSON and read them in? What are they?

andrewheumann: It's just a test of a crazy polygon I drew in rhino along with a hundred random samples and their results. Just making sure our containment results agree with theirs. But I can see how this is overkill.

src/Elements/Geometry/Ray.cs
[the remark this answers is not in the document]
andrewheumann: This is a good point. I'm switching to using the internal method that profile.contains uses, which just tests a collection of segments.

PR #276, opened 2020-04-02
Multi targeting

andrewheumann:
BACKGROUND:
- In order to support the Grasshopper plug-in and other .NET Framework applications, Elements is expanded to target multiple frameworks

DESCRIPTION:
- Target .Net Framework 4.7.2 and .Net Core 2.1 in addition to netstandard2.0.
- Establish conditional compile constants for any changes necessary to support 4.7.2
- Refactor and expand  `Elements.Generate.TypeGenerator` to support the dynamic type compilation needs of the Grasshopper plug-in
- Minor code changes to fix errors that threw in Framework builds (e.g. `Guid id = Guid.Empty;`)

TESTING:
- Have tested with the Grasshopper plug-in and run all Unit Tests
  
FUTURE WORK:
- Additional configuration may be necessary to correctly support a multi-targeted NuGet package (https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks)

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/276)

PR #317, opened 2020-05-10
Multitargeting

andrewheumann:
BACKGROUND:
- Picking back up #276
- In order to support the Grasshopper plug-in and other .NET Framework applications, Elements is expanded to target multiple frameworks

COMMENTS:
- Though the intention is to target only netstandard2.0 and net472, the project targets involve a few additional frameworks / framework changes:
**Elements.csproj** - netstandard2.0;net472;netcoreapp3.1;netcoreapp2.1
- without netcoreapp3.1 targeted in elements.csproj, the test project would not build on mac
- without netcoreapp2.1 targeted in elements.csproj, the test project would not build on windows.
**Elements.Tests.csproj** - netcoreapp3.1
- without this bump, I could not get the test project to build on mac. 

Long story short, I tried every combination under the sun, and this is the only configuration that would build AND test on both mac and windows. 

DESCRIPTION:
- sets up Elements to target multiple frameworks
- explicitly tells Elements.Tests which version of Elements to use
- fixes one minor bug that was throwing a build error 
- updates travis configuration to use dotnet 3.1

TESTING:
- Please do test this with your own build configuration to ensure it's compatible. If you have any errors at all, delete `bin` and `obj` from both the Elements and the Elements.Test project directories, and then rebuild. 
  
FUTURE WORK:
- Some functional changes are necessary to fold in the necessary changes to the .net framework build in order to get it to work with grasshopper
- currently only the netstandard build is tested — getting tests that run for framework will require additional work.  
- travis builds are currently failing

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/317)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,9 @@ language: csharp
 solution: ./hypar.sln
 dist: trusty
 mono: none
-dotnet: 3.1.201
+dotnet: 3.1
+install:
+  - dotnet restore
 script:
   - dotnet build ./Elements.sln
   - dotnet test ./test/Elements.Tests/Elements.Tests.csproj
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: csharp
 solution: ./hypar.sln
 dist: trusty
 mono: none
-dotnet: 2.1.502
+dotnet: 3.1.201
 script:
   - dotnet build ./Elements.sln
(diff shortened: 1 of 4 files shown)


ikeough:
Here's one of the initial announcements about .net standard. https://github.com/dotnet/standard/issues/481. Important to note is the section about test projects where they specify that you should enable automatic binding redirects on test projects to make .net standard work correctly.

PR #335, opened 2020-06-17
use PropertyName not just name

wynged:
BACKGROUND:
- code generation for generation C# class files from schemas was providing bad results when the properties had a space.

DESCRIPTION:
- change the constructor template to use PropertyNames rather than using just the name property.  PropertyName appears to be a cleaned up value, sans spaces

TESTING:
- This change combined with the ThreadSafety flag in a different PR allowed me to update and publish the JSON to Model function which generates code from a large number of schemas.
  
FUTURE WORK:
- We'll need to flex the CLI after these changes, before next release.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/335)


wynged:
this is superceded by #337

PR #387, opened 2020-09-14
add thread-safe and filters to all code gen paths

wynged:
BACKGROUND:
- While working on the export server we found out it was failing to generate the correct code, because it lacked the correct parm filters, and also possibly due to thread issues, although those weren't as obvious, they did result in strange behavior during debug.

DESCRIPTION:
- Add the line for supporting async to each code gen entry point.  
- Add the line for registering the name converting filters to each code gen entry point.

TESTING:
- Code gen when loading multiple templates into memory.  Test code for the ExportServer would help you test this change if you wanted to run it locally. https://github.com/hypar-io/api/tree/migrate-export-server/ExportServer

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/387)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/387)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #388, opened 2020-09-16
Nuget fix

wynged:
BACKGROUND:
- Export server in it's new location in the api repo needed to ref code gen via nuget package, not via local reference.  This didn't work because of missing templates.

DESCRIPTION:
- add a build target that is packaged with the nuget package.  Then when this package is referenced it calls this build target which copies the `Templates` directory into the bin folder of the consuming project.  See the file `Elements.CodeGeneration.targets` and the .csproj file for the meat of the nuget fix.
- Also refactor some classes into separate files.

TESTING:
- Tested with local nuget references in ExportServer, and also in a test proejct.
- Tested after deploying 0.8.0-alpha.7
  
FUTURE WORK:

COMMENTS:
- turns out this resolves a long standing unknown issue where code gen wouldn't work if done with Elements as a nuget package reference unless a custom templates folder was provided and set, as was the case in GrasshopperRunner.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/388)

### DIFF
--- a/Elements.CodeGeneration/src/CompilationResult.cs
+++ b/Elements.CodeGeneration/src/CompilationResult.cs
@@ -1,0 +1,23 @@
+using System.Reflection;
+
+namespace Elements.Generate
+{
+    /// <summary>
+    /// The result of a compilation.
+    /// </summary>
+    public struct CompilationResult
+    {
+        /// <summary>
+        /// True if the compilation succeeded.
+        /// </summary>
+        public bool Success { get; internal set; }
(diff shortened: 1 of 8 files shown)


andrewheumann (approved):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/388)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #390, opened 2020-09-24
check missing texture and null mess

wynged:
BACKGROUND:
- gltf exporting caused failures in the export server.

DESCRIPTION:
- don't add material with missing texture paths.
- ignore meshes where the mesh is null.

TESTING:
- the temporary stack ew-sand has an export server that uses this version of Elements, you can use this workflow to see two functions that normally break being exported.
https://dev.hypar.io/workflows/4442ed8d-b409-409c-96e3-b756b0491258?api=ew-export

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/390)

### DIFF
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -136,7 +136,7 @@ namespace Elements.Serialization.glTF
                     };
                 }
 
-                if (material.Texture != null)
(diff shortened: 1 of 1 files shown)


ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/390)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ikeough)

Issue #401, opened 2020-10-07
ifc -> gltf?


bertt:
Hi, 

I'm converting a ifc 4 model (AC-20-Smiley-West-10-Bldg.ifc from this repository) to glTF using this code:

```
            var newModel = IFCModelExtensions.FromIFC("AC-20-Smiley-West-10-Bldg.ifc");
            newModel.ToGlTF("test.glb");
``` 
I'm using 0.8.0-alpha.9 of Hypar.Elements.Serialization.IFC. 

In the result it looks like there are missing colors and geometries. Is there something possible to improve the output?

Ifc model:

![image](https://user-images.githubusercontent.com/538812/95315819-4bcef080-0893-11eb-935a-89ec55e90b7e.png)

glTF:

![image](https://user-images.githubusercontent.com/538812/95315865-5ab5a300-0893-11eb-8149-842a3f851d9b.png)

ikeough:
There is still much improvement needed on the ifc->gltf pipeline. There are a number of ways which color can be prescribed in IFC. We haven't settled on which one we want to support for colors and material support. You can read a post about this [here](https://forums.buildingsmart.org/t/why-is-it-so-difficult-to-get-colors-to-show-up/2312). Windows and doors are missing because we don't yet support boolean operations for openings. This will come soon as we add CSG support to our solids kernel.

bertt:
ok good to know this is work in progress. About the geometry: I'm wondering why the ifc has 3 building levels, the glTF has 4?

ikeough:
I believe the confusion is that in the gltf you're seeing the porch slab as if it were the porch _roof_. The porch roof and posts are missing, which is interesting because I believe we use this same model in our testing setup.

ikeough:
The work we're doing in #406 get us here:
![image](https://user-images.githubusercontent.com/1139788/95889786-ba9ac680-0d37-11eb-870f-36806984b778.png)

Moult:
Have you considered using the IFC to gltf workflow used by IfcOpenShell's IfcConvert?

bertt:
@Moult  yes works very with IfcConvert, just investigating alternatives for more custom solutions.

PR #406, opened 2020-10-12
Csg

ikeough:
BACKGROUND:
For any case where we want to create an opening in an element, we have thus far had to employ tricks to cut those openings, like clipping opening perimeter polygons out of the perimeter of the opening's target (floors, walls, etc.). This required having this clipping code in an element's `UpdateRepresentation` method, which made the element's instance a container for state about a clipping operation.

DESCRIPTION:
This PR implements csg solids using [Frank Krueger's port of OpenJsCad's csg.js](https://github.com/praeclarum/Csg). A csg is created for every solid operation and used during the creation of meshes for visualization. By moving to csg, we remove any clipping code from the elements themselves and apply csg operations when we are creating the visualization. We follow this work by updating openings to apply their subtraction operation after other solid operations are completed. This enables a more general application of openings to any geometric element. With csgs we can also now support IFC boolean operations. Finally, because operations to create openings are dependent on the geometry and transform of the host, we introduce new APIs to `IHasOpenings` to provide a simpler interface for creating openings in walls and floors.

FUTURE WORK:
- Csgs should probably be hidden behind the `Opening` APIs for users who just want to poke holes in things.
- Creating a solid, then a csg, then a mesh for visualization involves copying a lot of data. It isn't fast. It is possible that we could skip creating solids altogether and work with csgs as the primary representation of a solid.

Csg holes added to beams.
image

Csg operation on a topography mesh preserving the texture coordinates of the mesh.
image

Additive and subtractive operations.
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/406)


ikeough:
@mattc Asking for your review so we can start a conversation about how this may or may not work with the csg work you're doing in Hypar's front end.

ikeough:
Relates to #401.

mattc (commented):
I was doing BRep stuff in the front end, not CSG.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/406#-:-MJUuMdI-76PEwdDtZkW:bibkf3b)* status: 0 of 29 files reviewed, all discussions resolved (waiting on @mattc)

mattc (changes requested):
> Reviewed 19 of 26 files at r1, 6 of 9 files at r2, 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/406)* status: all files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[Elements/src/Interfaces/IHasOpenings.cs, line 4 at r3](https://reviewable.io/reviews/hypar-io/elements/406#-MJlfIji5BMSNR0U3PQN:-MJlfIji5BMSNR0U3PQO:b-1273y3) ([raw file](https://github.com/hypar-io/elements/blob/80d36acde9b02e64eac15bfd40cb45e8152a9227/Elements/src/Interfaces/IHasOpenings.cs#L4)):*
> ```Smalltalk
> 
> using System.Collections.Generic;
> using Elements.Geometry;
> ```

Did this need to be added?
___
*[Elements/test/CsgTests.cs, line 65 at r3](https://reviewable.io/reviews/hypar-io/elements/406#-MJlfbVYE9AwH755Grsz:-MJlfbVZ5yIqY06ZYtzv:b-93qath) ([raw file](https://github.com/hypar-io/elements/blob/80d36acde9b02e64eac15bfd40cb45e8152a9227/Elements/test/CsgTests.cs#L65)):*
> ```Smalltalk
>         //     }
>         //     // </example>
>         // }
> ```

So that this doesn't gradually become broken code, can you add an issue to finish it? (maybe leave the code in the issue)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/406)* status: 27 of 32 files reviewed, 2 unresolved discussions (waiting on @mattc)

___
*[Elements/src/Interfaces/IHasOpenings.cs, line 4 at r3](https://reviewable.io/reviews/hypar-io/elements/406#-MJlfIji5BMSNR0U3PQN:-MK5vWgE8xYM9_qSb-1Z:bj070ae) ([raw file](https://github.com/hypar-io/elements/blob/80d36acde9b02e64eac15bfd40cb45e8152a9227/Elements/src/Interfaces/IHasOpenings.cs#L4)):*

Did this need to be added?
</blockquote></details>

Previously yes, but later no.
___
*[Elements/test/CsgTests.cs, line 65 at r3](https://reviewable.io/reviews/hypar-io/elements/406#-MJlfbVYE9AwH755Grsz:-MK5vAYA-kew-yDWbhWU:b-jeym8v) ([raw file](https://github.com/hypar-io/elements/blob/80d36acde9b02e64eac15bfd40cb45e8152a9227/Elements/test/CsgTests.cs#L65)):*

So that this doesn't gradually become broken code, can you add an issue to finish it? (maybe leave the code in the issue)
</blockquote></details>

This has been moved over to benchmarks.

mattc (approved):
> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/406)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #409, opened 2020-10-16
round domain extension

wynged:
BACKGROUND:
- Got a splitting error while using Grid2d because a value was 1.00000000002

DESCRIPTION:
- round the output of MapFromDomain to the geometric tolerance.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/409)


wynged:
false alarm, this doesn't appear needed any more, and caused some tests to break.

PR #413, opened 2020-10-26
Fix bug with solid glb creation.

ikeough:
The accessor count was inadvertently used where the node count should be, leading to glbs which had incorrect node array sizes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/413)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/413)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #422, opened 2020-10-29
Use tags instead of version.txt.

ikeough:
This PR updates actions to use tags instead of the version.txt file. The version.txt file was an ineffective way to store the version number as updating the version and committing the file is prevented on the protected master branch. Tags can be created on the master branch AND can be the durable store for versions.

The strategy is as follows:

- When a PR is created
  - Build and test.
- When a PR is merged to master
  - The last tag is read and incremented.
  - A new alpha tag is created.
  - A NuGet package with that version number is published.
- When a tag is created (release)
  - The last tag is read.
  - A NuGet package with that version number is published.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/422)


wynged (approved):
> Reviewed 1 of 2 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/422)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #428, opened 2020-11-02
always copy the csg lib

wynged:
BACKGROUND:
- the csg library wasn't found while working on the ExportServer lambda in the api repo
```
There was an error writing an element of type Elements.Beam to IFC: Could not load file or assembly 'Csg, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.

```

DESCRIPTION:
- Add the setting `CopyToOutputDirectory=Always`

TESTING:
- next nuget package will work.  it's also running live on the sam stack `api=ew-export` which you could use from explore while exporting gltf.
  
FUTURE WORK:
- need to confirm this setting is all we need for nuget packages to always have this file.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/428)

### DIFF
--- a/Elements/src/Elements.csproj
+++ b/Elements/src/Elements.csproj
@@ -27,15 +27,16 @@
   <ItemGroup>
     <Content Include="Textures\**\*.*">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>   
+    </Content>
   </ItemGroup>
 
   <ItemGroup>
     <Reference Include="Csg ">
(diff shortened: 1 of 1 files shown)


ikeough (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/428#-:-MLASCDz2Mib-rVyUfie:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @andrewheumann and @ikeough)

PR #445, opened 2020-11-06
Fix issue with inverted vertical curve transforms.

ikeough:
BACKGROUND:
CSGs were failing to boolean properly with strange results. We found that vertical sweeps had their geometry inverted. This was due to an inverted Y axis of the transforms along a line causing transforms to not be "right handed".

DESCRIPTION:
This PR properly orients transforms along vertical curves and adds a test to ensure vertical curves in the up and down direction both work correctly.

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/445)

### DIFF
--- a/Elements/src/Geometry/Transform.cs
+++ b/Elements/src/Geometry/Transform.cs
@@ -98,6 +98,14 @@ namespace Elements.Geometry
                 x = test.Cross(z).Unitized();
                 y = x.Cross(z.Negate()).Unitized();
             }
+            else
+            {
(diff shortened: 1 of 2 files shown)


wynged (approved):
amazing, thanks!

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/445#-:-MLZRmX0BC4UMO9xsPzr:b4c1vik)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #456, opened 2020-11-24
Fix Union All overlap treatment behavior

andrewheumann:
BACKGROUND:
- Polygon.UnionAll did not work as expected when supplied with a list of overlapping polygons. Because we were using Clipper's EvenOdd PolyFillType by default, it treated the overlapping region between two polygons as a "hole" when it should have been added to the union. 
image
 

DESCRIPTION:
- Adds a parameter to BooleanTwoSets to govern which PolyFillType to apply
- Sets UnionAll to use VoidTreatment.IgnoreInternalVoids
- Preserves existing behavior for other non-static cases, which will allow us to join two shapes-with-holes and get a correct result.

TESTING:
- Run the new `PolygonTests.UnionAll` test
  
FUTURE WORK:
- It's conceivable that we might need to expose this option all the way to the user.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/456)


mattc (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/456)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #463, opened 2020-12-09
Fix #461: Support 3d polylines

### LINKED ISSUE #461: Support non-coplanar polylines
**Is your feature request related to a problem? Please describe.**
Elements.Geometry.Polyline have a validator that requires them to be coplanar. This makes it impossible to use polylines to drive arbitrary extrusions, like those needed for expressing MEP systems.

**Describe the solution you'd like**
Remove the `vertices.AreCoplanar()` validation on the current Elements.Geometry.Polyline class and remove any methods that assume coplanar polylines. A separate 2D Polyline Element can be introduced later if needed, e.g. to enforce compatibility with AutoCAD 2d polylines.

**Describe alternatives you've considered**
Keeping the current class coplanar and add a new 3D polyline class.

**Additional context**
Envelope By Centerline not only assumes coplanar polylines, but it assumes that they are parallel to the ground plane. Since the ground plane restriction is only enforced by UI and not by deserialization, it seems reasonable to require that the coplanar restriction will be too.

I propose changing :
- Validators.cs:PolylineValidator: remove AreCoplanar check.
- In the Elements.Geometry.Polyline definition:
  - Update header comments
  - Remove CheckCoincidenceAndThrow. It's protected and not used anywhere.
  - Update FramesInternal to stop assuming coplanar polylines, trying not to cause too much twistyness on extrusions in cases of orthogonal bends.
  - Leave Normal, TransformAt but comment that they're only meaningful for coplanar polylines
  - Leave ToClipperPath and Offset, but comment that they assume coplanar polylines and a Z-up normal. This is already unenforced and it doesn't look like they'll break.

mattc:
BACKGROUND:
See #461.

DESCRIPTION:
This removes the validation that polylines must be coplanar, adjusts docs accordingly, and updates methods to support 3d polylines where appropriate.

TESTING:
- I updated tests and verified that results looked reasonable. Coplanar polyline sweeps are unchanged and non-coplanar sweeps look reasonable:
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/463)

### DIFF
--- a/Elements/src/GeoJSON/Geometry.cs
+++ b/Elements/src/GeoJSON/Geometry.cs
@@ -232,7 +232,7 @@ namespace Elements.GeoJSON
         }
 
         /// <summary>
-        /// Convert the coordinate array to a collection of polylines.
+        /// Convert the coordinate array to a collection of polygons.
         /// The last position of the polygon is dropped.
         /// </summary>
         /// <returns></returns>
--- a/Elements/src/Geometry/Circle.cs
+++ b/Elements/src/Geometry/Circle.cs
@@ -26,7 +26,7 @@ namespace Elements.Geometry
         /// Create a polygon through a set of points along the arc.
         /// </summary>
         /// <param name="divisions">The number of divisions of the arc.</param>
-        /// <returns>A polyline.</returns>
+        /// <returns>A polygon.</returns>
         public Polygon ToPolygon(int divisions = 10)
         {
             var pts = new Vector3[divisions];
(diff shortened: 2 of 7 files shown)


wynged (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/463)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #465, opened 2020-12-10
Utility Conversions for Curve => model curve and Line => Polyline

andrewheumann:
BACKGROUND:

DESCRIPTION:
- Adds implicit conversion from Curve => ModelCurve to make quick visualization of a curve more terse in syntax
- Adds an extension method `Line.ToPolyline()`
- Adds a test ensuring that `Model.AddElement(Curve c)` converts as expected

TESTING:
- New test is added for the modelcurve conversion
  
FUTURE WORK:
- Similar conversions for `Representation` from `SolidOperation` — (deferring until Ian's work on Representations comes through)
- We may also want to make explicit extension methods to perform these conversions — they would not be more terse, but would be more discoverable. 

COMMENTS:
- the `Element(Polygon c)` conversion on the Polygon class may seem redundant, but it is necessary to ensure that the default element conversion for a Polygon is to a ModelCurve instead of a ........ and then he trailed off and decided to put it in a comment instead
- Got the Ian seal of approval on slack

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/465)

### DIFF
--- a/Elements/src/Geometry/Curve.cs
+++ b/Elements/src/Geometry/Curve.cs
@@ -93,5 +93,11 @@ namespace Elements.Geometry
         /// </summary>
         /// <param name="transform">The transform to apply.</param>
         public abstract Curve Transformed(Transform transform);
+
+         /// <summary>
+        /// Implicitly convert a curve to a ModelCurve Element.
+        /// </summary>
+        /// <param name="c">The curve to convert.</param>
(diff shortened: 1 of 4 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/465)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #454, opened 2020-11-23
Make Representation and Element


ikeough:
This issue consolidates one part of the conversation started in #450. The reason that `GeometricElement.IsElementDefinition` and `ElementInstance` exist was solely to allow the sharing of geometric representations between elements with different transforms. It was overkill for the intended purpose however, and did not consider patterns that are already widely used and enforce separation of concerns in C# code like dependency injection. Instead of creating a new type, `ElementInstance` which references a base definition, we could have simply given every instance of a type with a shared representation a reference to that representation. This was impossible of course because we have no way of referencing a shared representation in a `Model` unless `Representation` becomes an element. That is what this enhancement proposes to do.

By making `Representation` inherit from `Element`, we gain the ability to store `Representation` instances in the `Model` in the same way that we store `Material` and `Profile` instances. 

The enhancement proposes the following work.
- Make `Representation` inherit from `Element`.
- Make `Material` a property of `Representation`.
  - This is possibly the most controversial part of this proposal as a "material" in AEC provides both visual characteristics and mechanical properties. As `Material` currently only represents visual characteristics in Elements this associates the material with the representation for that purpose.
- Create a base `Representation` class:
```cs
public abstract class Representation
{
  public Material Material {get;set;}
  public Representation(Material material)
  {
    this.Material = material;
  }
}
```
- Create a new `SolidRepresentation` class.
```cs
public class SolidOperations: Representation, IList<SolidOperation>
{
  
  public SolidRepresentation(List<SolidOperation> operations)
  {
    this.Operations = operations;
  }
}
```
- Replace `MeshElement` with a `MeshRepresentation` that specifies a `Mesh` and deprecate `MeshElement`.
- Replace `ImportElement` with an `ImportRepresentation` that imports a mesh and deprecate `ImportElement`.

```cs

var rep = new SolidRepresentation(List<SolidOperation> solidOperations);

// Create an instance of Foo, which is a GeometricElement
var foo1 = new Foo(instanceProp1, instanceProp2, ... transform1, rep, Guid.NewGuid(), "foo1");
var foo2 = new Foo(instanceProp3, instanceProp4, ..., transform2, rep, Guid.NewGuid(), "foo2");
```

ikeough:
@andrewheumann @wynged I believe this addresses a number of concerns that we discussed.

ikeough:
[...]

PR #468, opened 2020-12-16
call UpdateRepresentations in Ray.Intersects(GeometricElement)

### LINKED ISSUE #467: Ray.Intersects(GeometricElement) should call UpdateRepresentations
A user reasonably expects that calling `Ray.Intersects(GeometricElement e)` on an element they have created should return the intersections with that element. However, if `UpdateRepresentations` has not been called on the GeometricElement, intersection will fail because the element (a StandardWall, for instance) does not yet have a representation on creation. We should call UpdateRepresentations inside the `Intersects` method to ensure this always works.

andrewheumann:
fixes #467

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/468)

### DIFF
--- a/Elements/src/Geometry/Ray.cs
+++ b/Elements/src/Geometry/Ray.cs
@@ -88,6 +88,10 @@ namespace Elements.Geometry
         /// <returns></returns>
         public bool Intersects(GeometricElement element, out List<Vector3> result)
         {
+            if (element.Representation == null || element.Representation.SolidOperations == null || element.Representation.SolidOperations.Count == 0)
+            {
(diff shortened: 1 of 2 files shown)


ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/468)* status: :shipit: complete! all files reviewed, all discussions resolved

ikeough:
This will be superseded by work done for #454. We'll be making all core types create representations in their constructors, so the call to `UpdateRepresentations` will be unnecessary.

PR #470, opened 2020-12-21
round value before ACOS

wynged:
BACKGROUND:
- Got NaN results back from AngleTo because varialbe `r` was -1.0000002, and -1 is not a valid number to use in Acos function.

DESCRIPTION:
- round the value that is going into Acos to the 5th digit.

TESTING:
- Tests pass, works now in my function.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/470)

### DIFF
--- a/Elements/src/Geometry/Vector3.cs
+++ b/Elements/src/Geometry/Vector3.cs
@@ -164,8 +164,8 @@ namespace Elements.Geometry
         }
 
         /// <summary>
-        /// The angle in degrees from this vector to the provided vector. 
-        /// Note that for angles in the plane that can be greater than 180 degrees, 
+        /// The angle in degrees from this vector to the provided vector.
+        /// Note that for angles in the plane that can be greater than 180 degrees,
         /// you should use Vector3.PlaneAngleTo.
         /// </summary>
         /// <param name="v">The vector with which to measure the angle.</param>
@@ -184,7 +184,7 @@ namespace Elements.Geometry
             {
                 return 0.0;
             }
-            var rad = Math.Acos(r);
+            var rad = Math.Acos(Math.Round(r, 5));
             return rad * 180 / Math.PI;
         }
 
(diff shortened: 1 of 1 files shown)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/470)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrewheumann and @wynged)

___
*[Elements/src/Geometry/Vector3.cs, line 187 at r1](https://reviewable.io/reviews/hypar-io/elements/470#-MP5Ztd0AGcX0MXuuvuy:-MP5Ztd0AGcX0MXuuvuz:basmmoy) ([raw file](https://github.com/hypar-io/elements/blob/563e30a50fc059ef260c303e7e961f4fa28af726/Elements/src/Geometry/Vector3.cs#L187)):*
> ```Smalltalk
>                 return 0.0;
>             }
>             var rad = Math.Acos(Math.Round(r, 5));
> ```

We should probably not introduce rounding math independent of system-level tolerance — we should follow the pattern above and use r.ApproximatelyEquals(-1) and return 180.

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/470)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Vector3.cs, line 187 at r1](https://reviewable.io/reviews/hypar-io/elements/470#-MP5Ztd0AGcX0MXuuvuy:-MP5gusk3dDN0L7tnScr:bkwijse) ([raw file](https://github.com/hypar-io/elements/blob/563e30a50fc059ef260c303e7e961f4fa28af726/Elements/src/Geometry/Vector3.cs#L187)):*

We should probably not introduce rounding math independent of system-level tolerance — we should follow the pattern above and use r.ApproximatelyEquals(-1) and return 180.
</blockquote></details>

done, you right.

andrewheumann (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/470)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #471, opened 2020-12-29
Improvements to Type Generation

ikeough:
BACKGROUND:
This PR is the first PR in support of #454. Upon starting that work I found that many of the "core" types that we use did not have schemas, and that generation of C# class using the `TypeGenerator` was a mess. Schemas in this repository were referring to `hypar.io/Schemas` which made local editing and re-generation of types impossible. Several generated types did not have corresponding schemas.

DESCRIPTION:
This PR accomplishes the following:
- It adds the `CoreTypeGenerator` application, which can be called by the `build_types.sh` script. This application uses our type generation APIs to generate C# classes.
- It updates all JSON schemas to have absolute references to the the raw urls from github for a schema. This enables a developer to tweak the core schemas and re-generate without having to push schemas into buckets in S3.
- It cleans up `TypeGenerator`, removing an unused function for local type generation, and improves struct handling through the use of the `x-struct` property (non standard).
- It updates the `TypeGenerator` to make `Element` implement `INotifyPropertyChanged` which is required for further work on #454, but has no run-time impact yet.
- It fixes our liquid template to clean-up tabs in the generated code.
- It removes `JsonInheritanceAttribute` and `UserElementAttribute` in favor of requiring schemas to implement the `discriminator` field for derived types.

FUTURE WORK:
- The core schemas are now 100% accurate within this repo, but most likely do not align with the schema manager. A one-time find and replace can be used to update all references when the schema manager is ready to support the core schemas. Even then, I'm not sure I'd recommend using the schema manager as a developer wants to be able to tweak these schemas by hand and generate.

TESTING:
- If you'd like to test this work, you can check out this branch and run the following command to generate all types from the local schemas:
```
./build_types.sh https://raw.githubusercontent.com/hypar-io/Elements/representation-type-generation/Schemas
```

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)


andrewheumann:
Just a note about this bit: 
```
It updates all JSON schemas to have absolute references to the the raw urls from github for a schema. This enables a developer to tweak the core schemas and re-generate without having to push schemas into buckets in S3.
```
I thought we were moving away from using GitHub for schemas in favor of the schema manager - is there a strong reason why we're still managing schemas on GitHub? I haven't actually looked at the code this refers to so I could be missing something.

ikeough:
@andrewheumann See the note I left in "future work". We can replace all those references in one go to the schema manager. But the schema manager doesn't currently have all of these schemas. Also, even once we do that, a developer will not be able to tweak a schema and do local code generation. This is a problem. The only solution that I can think of is to mirror these schemas in the schema manager.

wynged (changes requested):
I see lots of new mesh schemas, that appear to be new to this repo?  Where did they used to live, do we need to cleanup the old locations?

I get the following error while trying to generate the core types. Not sure exactly what collection is being modified here.
```
Generating type Elements.GeometricElement in /mnt/c/Users/ericr/source/Elements/Elements/src/Generate/GeometricElement.g.cs...
Generating type Elements.ContentElement in /mnt/c/Users/ericr/source/Elements/Elements/src/Generate/ContentElement.g.cs...
Generating type Elements.ContentCatalog in /mnt/c/Users/ericr/source/Elements/Elements/src/Generate/ContentCatalog.g.cs...
Unhandled exception. System.InvalidOperationException: Error while rendering Liquid template CSharp/Class: 
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at DotLiquid.Context.HandleError(Exception ex)
   at DotLiquid.Block.RenderAll(List`1 list, Context context, TextWriter result)
   at DotLiquid.Block.Render(Context context, TextWriter result)
   at DotLiquid.Document.Render(Context context, TextWriter result)
   at DotLiquid.Template.RenderInternal(TextWriter result, RenderParameters parameters)
   at DotLiquid.Template.Render(TextWriter writer, RenderParameters parameters)
   at DotLiquid.Template.Render(RenderParameters parameters)
   at NJsonSchema.CodeGeneration.DefaultTemplateFactory.LiquidTemplate.Render()
 ---> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at DotLiquid.Context.HandleError(Exception ex)
   at DotLiquid.Block.RenderAll(List`1 list, Context context, TextWriter result)
   at DotLiquid.Block.Render(Context context, TextWriter result)
   at DotLiquid.Document.Render(Context context, TextWriter result)
   at DotLiquid.Template.RenderInternal(TextWriter result, RenderParameters parameters)
   at DotLiquid.Template.Render(TextWriter writer, RenderParameters parameters)
   at DotLiquid.Template.Render(RenderParameters parameters)
   at NJsonSchema.CodeGeneration.DefaultTemplateFactory.LiquidTemplate.Render()
   --- End of inner exception stack trace ---
   at NJsonSchema.CodeGeneration.DefaultTemplateFactory.LiquidTemplate.Render()
   at NJsonSchema.CodeGeneration.CodeArtifact..ctor(String typeName, String baseTypeName, CodeArtifactType type, CodeArtifactLanguage language, CodeArtifactCategory category, ITemplate template)
   at NJsonSchema.CodeGeneration.CSharp.CSharpGenerator.GenerateClass(JsonSchema schema, String typeName)
   at NJsonSchema.CodeGeneration.CSharp.CSharpGenerator.GenerateType(JsonSchema schema, String typeNameHint)
   at NJsonSchema.CodeGeneration.GeneratorBase.GenerateTypes()
   at NJsonSchema.CodeGeneration.CSharp.CSharpGenerator.GenerateTypes()
   at Elements.Generate.TypeGenerator.GetCodeForTypesFromSchema(JsonSchema schema, String typeName, String ns, Boolean isUserElement, String[] excludedTypes) in /mnt/c/Users/ericr/source/Elements/Elements.CodeGeneration/src/TypeGenerator.cs:line 428
```

> Reviewed 76 of 77 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471#-:-MPpW11l4ImkgRIbvug4:b-ibd09d)* status: all files reviewed, 13 unresolved discussions (waiting on @ikeough)

___
*[build_types.sh, line 3 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPoYQSS0ArtrPHLpksA:-MPoYQSS0ArtrPHLpksB:b-406t8a) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/build_types.sh#L3)):*
> ```Shell
> #!/usr/bin/env bash
> 
> BASE=$1
> ```

what is BASE for?
___
*[CoreTypeGenerator/Program.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPp9ErL1pSVCC2Pu2sO:-MPp9ErM-nmFfa7X-j3z:bnnybk4) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/CoreTypeGenerator/Program.cs#L1)):*
> ```Smalltalk
> using System;
> ```

rather than add this as a new project can we add this functionality to the generate-types command?  It appears that this isn't actually limited to generating core types, really it accepts a directory for it's input, and runs type generation on all of them, which would be generically useful.
___
*[CoreTypeGenerator/Program.cs, line 41 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpWRT73ANmAcM_9M4T:-MPpWRT8-ReTNckrESwU:b-bt91we) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/CoreTypeGenerator/Program.cs#L41)):*
> ```Smalltalk
>         }
> 
>         private static void ProcessFilesInDir(string dir, string inputDir, string outputDir, ref List<Task<GenerationResult>> tasks)
> ```

we don't need pass by ref here, you only need that if you're going to set the passed object to a new list, and you want to change the reference in the calling code.
For me this would be clearer as a recursive method that returned all the tasks (while generating the necessary directory structure) rather than a method that adds to a global list of tasks, but that might just be me.
___
*[Elements/src/ContentElement.cs, line 8 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpX8-a4W-RyH5LsGlc:-MPpX8-a4W-RyH5LsGld:b-j24f23) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/ContentElement.cs#L8)):*
> ```Smalltalk
> namespace Elements
> {
>     [UserElement]
> ```

shouldn't UserElement be added automatically to the generated file if necessary?  Is ContentElement a UserElement or a Core type?
___
*[Elements/src/Topography.cs, line 9 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpYR8X7bSXtu2-y79Y:-MPpYR8X7bSXtu2-y79Z:b-eyz7o4) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Topography.cs#L9)):*
> ```Smalltalk
> using LibTessDotNet.Double;
> using Newtonsoft.Json;
> using Vertex = Elements.Geometry.Vertex;
> ```

this using alias calls out a problem with our schemas, we have two vertices.  maybe leave a not here about what what we need to fix to not have this problem.
___
*[Elements/src/Geometry/BBox3.cs, line 97 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpaLiC9QwKM2gPxEVx:-MPpaLiC9QwKM2gPxEVy:b-e5ivjy) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/BBox3.cs#L97)):*
> ```Smalltalk
>         }
> 
>         /// <summary>
> ```

Not sure what about the type generator changes required the changes to BBox.
___
*[Elements/src/Geometry/BBox3.cs, line 110 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpaUno6CI5ThVcMBXj:-MPpaUno6CI5ThVcMBXk:b-y8l6gv) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/BBox3.cs#L110)):*
> ```Smalltalk
>             }
>             var a = (BBox3)obj;
>             return this.Min.IsAlmostEqualTo(a.Min) && this.Max.IsAlmostEqualTo(a.Max);
> ```

I think I'm ok with it, but does it seem odd that our Equals is technically an AlmostEquals ?
___
*[Elements/src/Geometry/Color.cs, line 23 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpaeqZ3MmNI2btuUY2:-MPpaeqZ3MmNI2btuUY3:b-1hbft9) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/Color.cs#L23)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="c"></param>
>         public static Color FromArray(double[] c)
> ```

Fine with these changes, but what about type generation required them?
___
*[Elements/src/Geometry/Vertex.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpbOpr4PXZiyrs1NgJ:-MPpbOps2oAQ_mjm-EL9:b-47wk5) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/Vertex.cs#L1)):*
> ```Smalltalk
> using System.Collections.Generic;
> ```

Why the new vertex class that seems to be a subset of the old one?
___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 57 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpbkYt43jEkk4-7tpp:-MPpbkYt43jEkk4-7tpq:b-t7lncc) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements.CodeGeneration/src/TypeGenerator.cs#L57)):*
> ```Smalltalk
>         /// The base url for schemas.
>         /// </summary>
>         public static string SchemaBase = "https://prod-api.hypar.io";
> ```

I don't think that any of the schemas listed below are valid with this SchemaBase.  the api schemas (managed by schema manager) have the format "https://prod-api.hypar.io/schemas/[SchemaName]"  with no .json at the end.
___
*[Schemas/GeometricElement.json, line 13 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpe4VW3KjT_eAcV1OC:-MPpe4VW3KjT_eAcV1OD:bf1dxd8) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Schemas/GeometricElement.json#L13)):*
> ```JSON
>     "allOf": [
>         {
>             "$ref": "https://raw.githubusercontent.com/hypar-io/Elements/representation-type-generation/Schemas/Element.json"
> ```

All of these urls need to be restored to the master branch urls before we can merge right? Otherwise all the schemas will be invalid.  Can we maybe put a variable in here something template-y looking like {{SchemaBasePath}} and then when we do code gen replace that with `https://raw.githubusercontent.com/hypar-io/Elements/master/` by default and allow overrides if someone wants to work with the schemas in a development branch?  
___
*[Schemas/Geometry/Mesh.json, line 2 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpfRQN2E4bksDyhB4p:-MPpfRQN2E4bksDyhB4q:bwg6f7q) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Schemas/Geometry/Mesh.json#L2)):*
> ```JSON
> {
>     "$id": "https://raw.githubusercontent.com/hypar-io/Elements/representation-type-generation/Schemas/Geometry/Mesh.json",
> ```

am I right that this is a brand new file to store the schema for mesh?  
___
*[Schemas/Properties/NumericProperty.json, line 2 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpgQyFEcvdq6WNUrvu:-MPpgQyFEcvdq6WNUrvv:bwl614e) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Schemas/Properties/NumericProperty.json#L2)):*
> ```JSON
> {
>     "$id": "https://raw.githubusercontent.com/hypar-io/Elements/representation-type-generation/Schemas/Properties/NumericProperty.json",
> ```

Is the numeric property used anywhere any more?  can we delete this schema and the generated file?

ikeough:
@wynged The latest update to this PR removes the `JsonInheritanceAttribute` and the `UserElementAttribute` completely. I believe this is a net positive but it has some down-sides that we should consider carefully. 

`JsonInheritanceAttribute` was there to minimize the type-walking that was required by the deserializer to find a derived type. The serializer would find the base type, `Element` say, and would look at its attributes to find all the derived types, match a discriminator and use the type corresponding to the discriminator. In other words, it used the attributes to declare the map of discriminators and derived types. No reflection was required except to get the attributes on the base class. 

Of course this didn't work for user-defined types which derived from `Element` but couldn't add themselves to the set of `JsonInheritanceAttribute` instances on the Element definition, so we added our funky `UserElementAttribute`. We began maintaining a cache in the converter of all `UserElement` attributed types. This strategy introduced a second path in the converter to look for user types. 

This work adds a one-time cache of all public types on assemblies in the directory of the executing assembly, but not including a bunch of dlls that we know to avoid (Microsoft, xunit, etc.) with their type's full name. Ex: `Elements.Geometry.Polygon`. 

The benefits of this strategy are:
- We no longer need to worry about `JsonInheritanceAttribute` or `UserElementAttribute`.
- We always set the `discriminator` to the type's `FullName`.

The drawbacks of this strategy are:
- We alway set the `discriminator` to the type's `FullName`. Maybe we want flexibility in the future?
- We have a potentially costly operation in building the type cache if the application has a huge number of dlls that we don't know that we want to avoid peeking into to find types.
- The `discriminator` is now required for any derived type.

On the whole this seems to vastly simplify the `JsonInheritanceConverter` and simplify the user-story for element type contributors.

wynged (changes requested):
> Reviewed 32 of 37 files at r3, 32 of 34 files at r4, 11 of 12 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: 109 of 110 files reviewed, 6 unresolved discussions (waiting on @ikeough)

___
*[build_types.sh, line 3 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPoYQSS0ArtrPHLpksA:-MQNe_cNDpvc0yZvd2Yl:b-7rl0kd) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/build_types.sh#L3)):*

Base is no longer used as the SchemaBase that it set has been removed from TypeGenerator.
</blockquote></details>

done
___
*[CoreTypeGenerator/Program.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPp9ErL1pSVCC2Pu2sO:-MQY1cuj5bW2lnWcR6tE:b-hhsm61) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/CoreTypeGenerator/Program.cs#L1)):*

It's not limited, and the idea is sound. However, I don't want to add a Hypar dependency to Elements. I'd be happy to add something similar to `hypar generate-types`.
</blockquote></details>

This project already does have an indirect dependency on Elements because it depends on Elements.CodeGeneration.  can't generate code for elements w/out Elements :)
___
*[Elements/src/Topography.cs, line 9 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpYR8X7bSXtu2-y79Y:-MQY2R2_B5DHH01Qc7DS:b-awg3ey) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Topography.cs#L9)):*

It's not a problem We have a solid vertex and a mesh vertex. They are in different namespaces. We need to disambiguate here because we've referenced both namespaces because we have methods further down that cut solids out of meshes.
</blockquote></details>

ahh... any chance we can change the namespace or name somehow to indicate this distinction?  maybe namespace Elements.Geometry.Mesh.Vertex so that it is distinguished the way Solid.Vertex is?
___
*[Elements/src/Geometry/BBox3.cs, line 97 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpaLiC9QwKM2gPxEVx:-MQY66lq3BlRyhPidHpd:b-83tbea) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/BBox3.cs#L97)):*

This is not actually required now, but it will be. In a follow-on PR for the representation work, we'll make the generator implement `INotifyPropertyChanged` on generated classes. INPC does an equality check of properties before raising the `PropertyChanged` event. Several of our types that were used as properties did not override Equals. It's fair to note that this should not have appeared until that later PR, but this is the result of breaking down a 138 file PR into many separate PRs. I'd vote to leave this in if that's OK.
</blockquote></details>

OK :)
___
*[Elements/src/Geometry/BBox3.cs, line 110 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpaUno6CI5ThVcMBXj:-MQY6B5gBR15d_6tseCZ:bb7x46g) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/BBox3.cs#L110)):*

This is following the pattern from `Vector3`. All of our spatial equality are tested within epsilon which is 1e-5 by default. Without this, bounding boxes with points that are 0.00000 and 0.00001 would be seen as different which is, in the vast majority of cases, not desirable. It's possible that in the future we want APIs that make this clearer everywhere like `Vector3.IsAlmostEqualTo`, but I wasn't going to add them here.
</blockquote></details>

ok cool
___
*[Elements/src/Geometry/Vertex.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpbOpr4PXZiyrs1NgJ:-MQY6CkQ8wGrB8DigUlT:bb7x46g) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Geometry/Vertex.cs#L1)):*

There are two different vertex classes and they should be distinct. A solid vertex, which is used by a solid edge, and a mesh vertex which conveys position, normal, uv, and color information for a mesh.
</blockquote></details>

ok cool
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 76 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQXzHzBEzXn3hRiXhSS:-MQXzHzBEzXn3hRiXhST:b-qbv7lc) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L76)):*
> ```Smalltalk
>                 if (fileName.Contains("Hypar.IFC")
>                     || fileName.Contains("Microsoft")
>                     || fileName.Contains("System")
> ```

I think we should be differently specific here, because Systems are a building type I can imagine being in a dll name, ie. `Acme.Building.Systems`  maybe change to `StartsWith(System)`
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 89 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQXzlNU91F-6jatqytC:-MQXzlNU91F-6jatqytD:b-ifby8k) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L89)):*
> ```Smalltalk
>                     continue;
>                 }
>                 var asm = assemblies.FirstOrDefault(a => a.GetName().Name == fileName);
> ```

Could per iterate through all of the assemblies directly rather than iterating through the filenames?  I think the checks for the filenames would work the same for assembly names since we are eventually checking for `Name =fileName`
___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 57 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpbkYt43jEkk4-7tpp:-MQY3FsbDTZxrkjK46HD:bqsz9j5) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements.CodeGeneration/src/TypeGenerator.cs#L57)):*

`SchemaBase` has been removed it wasn't actually required, nor were the full urls of the schemas as we only used that list to get the type names to determine what were "built in" type. All urls have been shortened to their name only.
</blockquote></details>

ok. Makese sense
___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 328 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQY4UoU87ceC2l-2riO:-MQY4UoVCSCLIjut47CZ:b-gtzg76) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements.CodeGeneration/src/TypeGenerator.cs#L328)):*
> ```Smalltalk
>         /// Get a list of the core Hypar types, which should be excluded from code generation.
>         /// </summary>
>         public static string[] GetCoreTypeNames()
> ```

I don't think this is used anywhere.
___
*[Schemas/GeometricElement.json, line 13 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpe4VW3KjT_eAcV1OC:-MQY6KNx7fFGjeC5Luki:b32wtd) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Schemas/GeometricElement.json#L13)):*

That is true, and it will be done right before merging the PR.
</blockquote></details>

ok
___
*[Schemas/Geometry/Solids/Sweep.json, line 26 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQY5Jdi2xZAet42O4Ym:-MQY5Jdi2xZAet42O4Yn:bpkeof2) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Schemas/Geometry/Solids/Sweep.json#L26)):*
> ```JSON
>         "Curve": {
>             "description": "The curve along which the profile will be swept.",
>             "$ref": "https://raw.githubusercontent.com/hypar-io/Elements/representation/Schemas/Geometry/Curve.json"
> ```

This is the only schema file that has been changed to this other branch name, should they all be the same?
Haven't succeeded in generating the files because of an error with these paths I think.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: 106 of 110 files reviewed, 6 unresolved discussions (waiting on @ikeough and @wynged)

___
*[CoreTypeGenerator/Program.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPp9ErL1pSVCC2Pu2sO:-MQejm7r97hcNxTkET-z:b-hog2ck) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/CoreTypeGenerator/Program.cs#L1)):*

This project already does have an indirect dependency on Elements because it depends on Elements.CodeGeneration.  can't generate code for elements w/out Elements :)
</blockquote></details>

What I meant is that I didn't want someone who uses Elements to have to know anything about Hypar. This is a general purpose library. If we then make them install the Hypar CLI to generate types, it introduces a practical dependency, not a binary one.
___
*[Elements/src/Topography.cs, line 9 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpYR8X7bSXtu2-y79Y:-MQekL9kCBd13bgop4XT:bu5oz5o) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Topography.cs#L9)):*

ahh... any chance we can change the namespace or name somehow to indicate this distinction?  maybe namespace Elements.Geometry.Mesh.Vertex so that it is distinguished the way Solid.Vertex is?
</blockquote></details>

I'd like namespaces to be split categorically or thematically.`Solids`is named as such because it contains solids classes as well as solid operation classes. That is, it is all classes related to solid operations. It's already not great as a categorical namespace. `Elements.Mesh` cuts a little too close to the class definition as a namespace. I'd rather keep vertex in `Elements.Geometry`. 
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 76 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQXzHzBEzXn3hRiXhSS:-MQenHMo9S7PhaSebCWY:b952lgg) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L76)):*

I think we should be differently specific here, because Systems are a building type I can imagine being in a dll name, ie. `Acme.Building.Systems`  maybe change to `StartsWith(System)`
</blockquote></details>

This has been removed entirely. It turns out that we can walk assemblies that have the JsonConverterAttribute with the parameter JsonInheritanceConverter.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 89 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQXzlNU91F-6jatqytC:-MQelWVd7-UUG3gRh1ST:bp0tdgz) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L89)):*

Could per iterate through all of the assemblies directly rather than iterating through the filenames?  I think the checks for the filenames would work the same for assembly names since we are eventually checking for `Name =fileName`
</blockquote></details>

Moot. See new strategy above.
___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 328 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQY4UoU87ceC2l-2riO:-MQej167DeBFzi85YyvT:b-gajkh8) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements.CodeGeneration/src/TypeGenerator.cs#L328)):*

I don't think this is used anywhere.
</blockquote></details>

I didn't  remove it because it is public. It's possible that it's used by Hypar in `generate types`.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: 106 of 110 files reviewed, 6 unresolved discussions (waiting on @wynged)

___
*[Schemas/Geometry/Solids/Sweep.json, line 26 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQY5Jdi2xZAet42O4Ym:-MQn7pqNDZNyVWp0ohIc:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Schemas/Geometry/Solids/Sweep.json#L26)):*

This is the only schema file that has been changed to this other branch name, should they all be the same?
Haven't succeeded in generating the files because of an error with these paths I think.


</blockquote></details>

Done.

wynged (changes requested):
> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: 109 of 110 files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[CoreTypeGenerator/Program.cs, line 1 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPp9ErL1pSVCC2Pu2sO:-MQstjpdExfpuyEDWUHM:b-pamx0l) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/CoreTypeGenerator/Program.cs#L1)):*

What I meant is that I didn't want someone who uses Elements to have to know anything about Hypar. This is a general purpose library. If we then make them install the Hypar CLI to generate types, it introduces a practical dependency, not a binary one.
</blockquote></details>

ahhh I see, you don't want a dependency on the hypar cli in this repo. makes sense
___
*[Elements/src/Topography.cs, line 9 at r2](https://reviewable.io/reviews/hypar-io/elements/471#-MPpYR8X7bSXtu2-y79Y:-MQsu6NfBRxM_HAZ4UyD:b32wtd) ([raw file](https://github.com/hypar-io/elements/blob/89254df67bdc8dc446f49f7ad28196dcc94c39a4/Elements/src/Topography.cs#L9)):*

I'd like namespaces to be split categorically or thematically.`Solids`is named as such because it contains solids classes as well as solid operation classes. That is, it is all classes related to solid operations. It's already not great as a categorical namespace. `Elements.Mesh` cuts a little too close to the class definition as a namespace. I'd rather keep vertex in `Elements.Geometry`. 
</blockquote></details>

ok
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 76 at r5](https://reviewable.io/reviews/hypar-io/elements/471#-MQXzHzBEzXn3hRiXhSS:-MQshsMK4620Ccbnhchg:b-b0qrb1) ([raw file](https://github.com/hypar-io/elements/blob/e8179bf8cf57ab47144221d05900b252b1abcdad/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L76)):*

This has been removed entirely. It turns out that we can walk assemblies that have the JsonConverterAttribute with the parameter JsonInheritanceConverter.
</blockquote></details>

Yup, this seems like a great way to get the correct types.  
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 68 at r6](https://reviewable.io/reviews/hypar-io/elements/471#-MQshHuL5hmX4c71C8rE:-MQshHuL5hmX4c71C8rF:b-za2iey) ([raw file](https://github.com/hypar-io/elements/blob/a35c47b283079fc761e2744e449251ad0d7ca443/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L68)):*
> ```Smalltalk
>         /// This includes base types, like elements, and all derived types like Wall.
>         /// We use reflection to find all public types available in the app domain
>         /// that have a JsonConverterAttribute whose first parameter is the 
> ```

maybe add the namespace to the classes in these comments.  (and finish sentence)  Would make it clearer to say
"that have the Newtonsoft.Json.JsonConverterAttribute  whose `ConverterType`property is a Elements.Serialization.JSON.JsonInheritanceConverter. "
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 86 at r6](https://reviewable.io/reviews/hypar-io/elements/471#-MQspOm81uktssz54EGx:-MQspOm81uktssz54EGy:bmckvy) ([raw file](https://github.com/hypar-io/elements/blob/a35c47b283079fc761e2744e449251ad0d7ca443/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L86)):*
> ```Smalltalk
>             var derivedTypes = typesWithConverterAttribute.SelectMany(t=>exportedTypes.Where(et=>t.IsAssignableFrom(et)));
> 
>             foreach(var t in typesWithConverterAttribute)
> ```

I think you can remove one of these loops, just loop over `derivedTypes` to be safe , but these lists are identical.  In the Filter for derived types you will also get the original types because `Element.IsAssignableFrom(Element)` will be true, and all relevant types seem to have the JsonInheritanceAttribute that you filtered for above, so derived types are already included.  
  I tested this by adding 
``` 
var same = new HashSet<Type>(typesWithConverterAttribute).SetEquals(new HashSet<Type>(derivedTypes));
``` 
and the two sets are equal.  also the second loop doesn't increase the number of items in the dictionary.

Confirmed it's true in the UserElementTests where the new type inherits from GeometricElement but doesn't explicitly add the attribute, it appears that the attribute exists on this class because GeometricElement has the attribute.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: 108 of 111 files reviewed, 2 unresolved discussions (waiting on @wynged)

___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 68 at r6](https://reviewable.io/reviews/hypar-io/elements/471#-MQshHuL5hmX4c71C8rE:-MQt5NhaCiWjfOta6Qym:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/a35c47b283079fc761e2744e449251ad0d7ca443/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L68)):*

maybe add the namespace to the classes in these comments.  (and finish sentence)  Would make it clearer to say
"that have the Newtonsoft.Json.JsonConverterAttribute  whose `ConverterType`property is a Elements.Serialization.JSON.JsonInheritanceConverter. "

</blockquote></details>

Done.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 86 at r6](https://reviewable.io/reviews/hypar-io/elements/471#-MQspOm81uktssz54EGx:-MQt5OWMA6sHDLodwFtU:b-73m6vw) ([raw file](https://github.com/hypar-io/elements/blob/a35c47b283079fc761e2744e449251ad0d7ca443/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L86)):*

I think you can remove one of these loops, just loop over `derivedTypes` to be safe , but these lists are identical.  In the Filter for derived types you will also get the original types because `Element.IsAssignableFrom(Element)` will be true, and all relevant types seem to have the JsonInheritanceAttribute that you filtered for above, so derived types are already included.  
  I tested this by adding 
``` 
var same = new HashSet<Type>(typesWithConverterAttribute).SetEquals(new HashSet<Type>(derivedTypes));
``` 
and the two sets are equal.  also the second loop doesn't increase the number of items in the dictionary.

Confirmed it's true in the UserElementTests where the new type inherits from GeometricElement but doesn't explicitly add the attribute, it appears that the attribute exists on this class because GeometricElement has the attribute.  
</blockquote></details>

Nice catch!

wynged (approved):
> Reviewed 1 of 12 files at r5, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/471)* status: all files reviewed, 1 unresolved discussion (waiting on @ikeough)

___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 86 at r6](https://reviewable.io/reviews/hypar-io/elements/471#-MQspOm81uktssz54EGx:-MQtGTlr9Kw_WnnoOpGc:b4iyri5) ([raw file](https://github.com/hypar-io/elements/blob/a35c47b283079fc761e2744e449251ad0d7ca443/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L86)):*

Nice catch!
</blockquote></details>

actually, we should stick with just the `typesWithConverterAttribute`

PR #477, opened 2021-01-13
Add a getter for CoreTypeNames.

ikeough:
BACKGROUND:
During testing of 0.8.1-alpha9 with the Hypar CLI, @wynged found that the `TypeGenerator.GetCoreTypeNames()` API was in fact required.

DESCRIPTION:
This PR introduces the `TypeGenerator.CoreTypeNames` property which wraps `_coreTypeNames`. A property is preferred over the method that we had previously because a method is no longer required to get the core type names.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/477)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Changelog
 
 ## 0.8.1
+
+### Added
+- `TypeGenerator.CoreTypeNames`
+
 ## Removed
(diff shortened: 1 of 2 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/477)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #480, opened 2021-01-15
Fix model serialization

wynged:
BACKGROUND:
- deserialization of some user elements failed.  it turned out it was elements that relied on Profiles created inside the UpdateRepresentation method.  

DESCRIPTION:
- serialize the exportModel that is created.  This model is the one that ensures all necessary child elements are serialized before the parent, so deserialization can happen correctly.
- lint fix in Changelog, looks big, it just removes a bunch of whitespace.

TESTING:
- The test has been modified and will succeed with this change.  Simplest addition to the test is that we assert there are no errors during deserialization.
  
FUTURE WORK:
- I think we might be about to remove UpdateRepresentation, so this might be unnecessary soon.

CHANGELOG:
- [x ] `CHANGELOG.md` is updated as necessary. 

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/480)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.2
+### Fixed
+- `Deserialization of profiles created in UpdateRepresentation`
+
 ## 0.8.1
 
 ### Added
@@ -38,7 +42,7 @@
   - The ContentElement is also added to the model by first checking for an available gltf, and then using a bounding box representation as a fallback.
 - `Transform.Scaled()` - This new method returns a scaled copy of the transform, allowing for a fluent like api.
 - `Transform.Moved(...)` - Return a copy of a transform moved by the specified amount.
-- `Transform.Concatenated(...)` - Return a copy of a transform with the specified transform concatenated with it. 
+- `Transform.Concatenated(...)` - Return a copy of a transform with the specified transform concatenated with it.
 - `IHasOpenings.AddOpening(...)` - `AddOpening` provides an API which hides the internals of creating openings.
 - `Opening.DepthFront` & `Opening.DepthBack` enable the creation of openings which extrude different amounts above and below their XY plane.
 - Solid operations which have `IsVoid=true` now use csg operations.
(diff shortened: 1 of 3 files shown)


Elements/src/Model.cs
```diff
@@ -173,7 +173,7 @@ public string ToJson(bool indent = false)
             }
             exportModel.Transform = this.Transform;
 
-            return Newtonsoft.Json.JsonConvert.SerializeObject(this,
+            return Newtonsoft.Json.JsonConvert.SerializeObject(exportModel,
```
ikeough, line 176: Good catch. This was the result of some work in the representations branch that got reversed, but not fully.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.2
+### Fixed
+- `Deserialization of profiles created in UpdateRepresentation`
```
ikeough, line 5: Please pull this out into an issue, and link the issue here. This makes it possible to track the issue in the future.

ikeough:
A full release should be done when this is merged. Without this, 0.8.1 is broken.

wynged (commented):
Fixes #483 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/480#-:-MRQVpQ05SAFfwlhubob:b74bu4y)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[CHANGELOG.md, line 5 at r1](https://reviewable.io/reviews/hypar-io/elements/480#-MR4JNG3b6KPgBeYlQWU-r1-5:-MRQVt9bA9OheCoLXSWU:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/78e0759c0e587a538f807723334b8e070a500433/CHANGELOG.md#L5)):*

Please pull this out into an issue, and link the issue here. This makes it possible to track the issue in the future.
</blockquote></details>

Done.
___
*[Elements/src/Model.cs, line 176 at r1](https://reviewable.io/reviews/hypar-io/elements/480#-MR4JNG3b6KPgBeYlQWV-r1-176:-MRQVusm7aiLSt27P3xL:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/78e0759c0e587a538f807723334b8e070a500433/Elements/src/Model.cs#L176)):*

Good catch. This was the result of some work in the representations branch that got reversed, but not fully.
</blockquote></details>

Done.

ikeough (approved):

ikeough:
Please do an 0.8.2 release after merge.

PR #482, opened 2021-01-19
Add representations and deprecate representation.

ikeough:
BACKGROUND:
This is the continuation of work for #454. 

Here we make `Representation` inherit from `Element`. The current representation is memorialized in the `SolidRepresentation` class, and numerous other representation classes are added to cover various use cases. We deprecate, but do not remove, the existing functionality including the `GeometricElement.Representation` property, which will now be `GeometricElement.Representations` to support multiple representations. 

To support deserialization of JSON representing previous versions of Elements, the `IMigration` class has been added to express migrations from one version to another. The `Model.ElementsVersion` property has been added to serialize the elements version to the model for use by the migration framework in subsequent deserialization.

DESCRIPTION:
See the [Changelog](https://github.com/hypar-io/Elements/blob/representation-add-representation/CHANGELOG.md).

TESTING:
- All tests currently pass, but testing this branch of the library in an existing function would be beneficial to assess the fitness of the migration framework. After updating to this version of the library, we expect that incoming JSON is properly migrated. 
  
FUTURE WORK:
- Building this branch will result in a large number of deprecation warnings. The next phase of this work will require re-routing all existing code to use the new properties.
- Ensure that the version checking strategy using the `Model.ElementsVersion` property works.

OPEN QUESTIONS:
- Do we need to support backwards migration?

CHANGELOG:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/482)


ikeough:
@andrewheumann I've closed this for the moment because there's another PR that is a shorter path to solving the initial issue.

PR #485, opened 2021-01-20
catch type load errors while building type cache. 

### LINKED ISSUE #484: Failure to deserialize Model if any assembly can't be loaded.
**Describe the bug**
Deserializing models fails with new type finding strategy if some assembly types can't be loaded.  This was observed in the Revit context

revit error logs said this
```
2021-01-19 17:00:15.526 -05:00 [DBG] Failed to load Model from Hypar.
System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
   at System.Reflection.Assembly.GetTypes()
   at Elements.Serialization.JSON.JsonInheritanceConverter.<>c.<BuildAppDomainTypeCache>b__12_0(Assembly asm)
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at Elements.Serialization.JSON.JsonInheritanceConverter.BuildAppDomainTypeCache()
   at Elements.Model.FromJson(String json, List`1 errors)
   at Hypar.Revit.LoadHyparModel.TryLoadModel(Document document, View activeView)
   at Hypar.Revit.LoadModelFromHypar.Execute(ExternalCommandData commandData, String& message, ElementSet elements)
```

**To Reproduce**
1. Try to load a model in the Revit context or in any case when various assemblies fail to return their types.

**Expected behavior**
Successfully load all available types and then deserialize the model.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: windows
 
**Additional context**
During debugging these are the assemblies that failed to load.  Some are from different Elements version issues (where converter uses different version) but some are just built into Revit, so no model loading at all was possible.
image

wynged:
BACKGROUND:
- Model de-serialization was broken in the Revit context due to assembly loading issues.

DESCRIPTION:
- catch type load errors, and simply skip them while building the type cache.
- Allow these errors to be added to the `Model.FromJson` error list
- Refactor to iterate over types only once while building cache.
- fixed #484 

TESTING:
- Use in Revit Context, and/or observe that all tests continue to work.
  
FUTURE WORK:

CHANGELOG:
- [ x] `CHANGELOG.md` is updated as necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/485)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.8.2
 ### Fixed
 - Fixed #483 `Deserialization of profiles created in UpdateRepresentation`
+- Fixed #484 `Failure to deserialize Model if any assembly can't be loaded.`
 
 
 ## 0.8.1
--- a/Elements/src/Model.cs
+++ b/Elements/src/Model.cs
@@ -182,24 +182,35 @@ namespace Elements
         /// </summary>
         /// <param name="json">The JSON representing the model.</param>
         /// <param name="errors">A collection of deserialization errors.</param>
-        public static Model FromJson(string json, List<string> errors = null)
+        /// <param name="forceTypeReload">Option to force reloading the inernal type cache. Use if you add types dynamically in your code.</param>
+        public static Model FromJson(string json, out List<string> errors, bool forceTypeReload = false)
         {
             // When user elements have been loaded into the app domain, they haven't always been
             // loaded into the InheritanceConverter's Cache.  This does have some overhead,
             // but is useful here, at the Model level, to ensure user types are available.
-            errors = errors ?? new List<string>();
-            JsonInheritanceConverter.RefreshAppDomainTypeCache(out var typeLoadErrors);
-            errors.AddRange(typeLoadErrors);
+            var deserializationErrors = new List<string>();
+            if (forceTypeReload)
+            {
+                JsonInheritanceConverter.RefreshAppDomainTypeCache(out var typeLoadErrors);
+                deserializationErrors.AddRange(typeLoadErrors);
+            }
+
             var model = Newtonsoft.Json.JsonConvert.DeserializeObject<Model>(json, new JsonSerializerSettings()
             {
                 Error = (sender, args) =>
                 {
-                    errors.Add(args.ErrorContext.Error.Message);
+                    deserializationErrors.Add(args.ErrorContext.Error.Message);
                     args.ErrorContext.Handled = true;
                 }
             });
+            errors = deserializationErrors;
             JsonInheritanceConverter.Elements.Clear();
             return model;
+        }
+
+        public static Model FromJson(string json, bool forceTypeReload = false)
+        {
+            return FromJson(json, out _, forceTypeReload);
         }
 
         private List<Element> RecursiveGatherSubElements(object obj)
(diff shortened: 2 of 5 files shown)


ikeough:
- Where is the exception being thrown? Is it during `Assembly.GetTypes()`? If so, can we filter for assemblies with types that are loadable wrapped in a try/catch, instead o wrapping each type introspection in one?
- With this change, what happens when you try to load an assembly with 10,000 private types in it? Does it do more or less work than the original?

wynged (commented):
- We will evaluate the same number of assemblies (they are all checked) in both schemes, and the same number of types are checked for `IsPublic && IsClass` (again, all of them).  The new code should return just as fast for a given type as the old one, as the `IsPublic && IsClass` check is the first one evaluated in the IsValidType method.
- The type cache has fewer items, but I believe this is actually a good thing.  Old types that use the UserElement custom attribute fail to be added to the list, but in the old code they are added back in while looking for derived types.  These types aren't actually valid however, they result in errors during model deserialization seen here, so we should just exclude it from the valid types cache when we build the cache.
  



Performance evaluation in a table below shows that this new system seems significantly faster.  I think the big win is avoiding the crossproduct check where all found types are evaluated against all types to get the derived types.  I also found some other interesting details.
- The ReflectionTypeLoadExceptoin occurs during assembly GetTypes() call, while the TypeLoadException occurs later, during the check for CustomAttributes.  I have changed the two catch statement to catch these two exceptions in separate locations in the code.
- The TypeLoadException filtering in the new code excludes elements that use the UserElement attribute.  


| type | Assemblies checked | types checked             | num types found | num errors | time | run in sequence |
|------|--------------------|---------------------------|-----------------|------------|------|-----------------|
|      |                    |                           |                 |            |      |                 |
| new  | 306                | 238531                    | 44              | 7          | 3.25 | 1               |
| new  | 384                | 264572                    | 44              | 7          | 1.53 | 2               |
| new  | 388                | 265129                    | 44              | 7          | 1.11 | 3               |
| new  | 388                | 265129                    | 44              | 7          | 1.12 | 4               |
|      |                    |                           |                 |            |      |                 |
|      |                    | types passing first check |                 |            |      |                 |
| old  |                    | 13127                     | 48              |            | 7.97 | 1               |
| old  |                    | 21743                     | 48              |            | 6.8  | 2               |
| old  |                    | 24823                     | 48              |            | 6.53 | 3               |
| old  |                    | 24973                     | 48              |            | 6.24 | 4               |
| old  |                    | 24973                     | 48              |            | 6.19 | 5               |

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/485#-:-MRVMIDwA3F9IZO4ktoy:bjpr1ge)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @ikeough)

wynged (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/485#-:-MRW1Jeg0eFxhmuYHMv3:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @ikeough)

ikeough (approved):

PR #487, opened 2021-01-22
Add Polygon.TransformSegment()

cdriesler:
BACKGROUND:
- I was developing this method for a specific function and realized it probably belongs in Elements proper.

DESCRIPTION:
- Adds the ability to translate just one 'segment' of a polygon (two consecutive vertices). 
- Throws an exception if the requested transformation is not within the polygon's current plane.
  - Triangles skip this check, since they are always planar.

TESTING:
Tests in the code:

- One-axis transformations of a square on the XY plane
- Two-axis transformations of a square in 3D space
- Confirm no work is done if specified segment does not exist
- Confirm function throws in request motion in non-planar
  
FUTURE WORK:
- There may be valid transformation cases that we support but I didn't realize! These should probably have tests.
- Some shapes may have valid non-planar transformations. (e.g. a flat square's edge can be moved vertically out-of-plane and still be a planar shape). I don't know how to generically check for this, though, and do not need to support this kind of motion. So, for now, I opted for the current more-constrained check.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/487)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - Fixed #483 `Deserialization of profiles created in UpdateRepresentation`
 - Fixed #484 `Failure to deserialize Model if any assembly can't be loaded.`
 
+### Added
+- `Polyline.TransformSegment(...)` - Allows transforms for individual polyline segments. May be optionally flagged as polygon and/or planar motion.
+
 ## 0.8.1
 
 ### Added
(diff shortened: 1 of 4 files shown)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/487)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #488, opened 2021-01-22
Add Polygon.SharedSegments()

cdriesler:
BACKGROUND:
- I need the ability to find shared segments between polygons for another function. I thought it would be useful to include in elements.

DESCRIPTION:
- Adds a static `Polygon.SharedSegments()` method.
- Given two polygons, it returns a set of tuples for any shared segments
  - A 'shared segment' means the start and end points are the same (although possibly reversed)
- The tuple is of the segment indices for their respective polygons. They can be different, hence tuple instead of just `int`.

TESTING:
Tests in code are:

- Concentric polygon circles return no results
- Intersecting polygon circles return no results
- Identical polygon circles return ten accurate results
- Identical polygon circles (one reversed) return ten accurate results
- A pair of squares with one shared edge return one result
- A pair of squares with touching (but not identical) edges returns no results

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/488)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - Fixed #483 `Deserialization of profiles created in UpdateRepresentation`
 - Fixed #484 `Failure to deserialize Model if any assembly can't be loaded.`
 
+### Added
+- `Polyline.SharedSegments()` - Enables search for segments shared between two polylines.
+
 ## 0.8.1
 
 ### Added
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -1003,56 +1003,6 @@ namespace Elements.Geometry
             verts.Add(this.Start);
             return verts;
         }
-
-        /// <summary>
-        /// Identify any shared segments between two polygons.
-        /// </summary>
-        /// <param name="a">The first polygon to compare.</param>
-        /// <param name="b">The second polygon to compare.</param>
-        /// <returns>Returns a list of tuples of indices for the segments that match in each polygon.</returns>
-        public static List<(int, int)> SharedSegments(Polygon a, Polygon b)
-        {
-            var result = new List<(int, int)>();
-
-            // Abbreviate lists to compare
-            var va = a.Vertices;
-            var vb = b.Vertices;
-
-            for (var i = 0; i < va.Count; i++)
-            {
-                var ia = va[i];
-                var ib = va[(i + 1) % va.Count];
-
-                for (var j = 0; j < vb.Count; j++)
-                {
-                    var ja = vb[j];
-
-                    if (ia.IsAlmostEqualTo(ja))
-                    {
-                        // Current vertices match, compare next vertices
-                        var jNext = (j + 1) % vb.Count;
-                        var jPrev = j == 0 ? vb.Count - 1 : j - 1;
-
-                        var jb = vb[jNext];
-                        var jc = vb[jPrev];
-
-                        if (ib.IsAlmostEqualTo(jb))
-                        {
-                            // Match is current segment a and current segment b
-                            result.Add((i, j));
-                        }
-
-                        if (ib.IsAlmostEqualTo(jc))
-                        {
-                            // Match is current segment a and previous segment b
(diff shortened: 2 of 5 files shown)


andrewheumann (changes requested):
Suggestion: 
This code applies nearly perfectly for Polylines too, from which Polygon inherits; is it possible to move it into Polyline?  There might need to be a few subtle conditional checks (whether to loop past the last vertex, etc) but it seems like most of it would hold. Happy to approve without this change, but seems like it would be useful!

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/488#-:-MReBC1e0KhyVrcCzZfY:b-g2b26j)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @cdriesler)

___
*[Elements/src/Geometry/Polygon.cs, line 1013 at r1](https://reviewable.io/reviews/hypar-io/elements/488#-MRe9nXG410VuJcBaX2K:-MRe9nXG410VuJcBaX2L:b-9xwymf) ([raw file](https://github.com/hypar-io/elements/blob/ef2879af65029cf0fd20fbc96124c5844909950b/Elements/src/Geometry/Polygon.cs#L1013)):*
> ```Smalltalk
>         /// <param name="b">The second polygon to compare.</param>
>         /// <returns>Returns a list of tuples of indices for the segments that match in each polygon.</returns>
>         public static List<(int, int)> SharedSegments(Polygon a, Polygon b)
> ```

I like to use named tuples in a case like this — List<(int indexOnA, int indexOnB)> — then whoever receives the result can get result.indexOnA instead of Item1

cdriesler (commented):
Agreed, moved it over and added an optional flag.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/488#-:-MResKLsEoa8uqa81vFj:bngtrdj)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 1013 at r1](https://reviewable.io/reviews/hypar-io/elements/488#-MRe9nXG410VuJcBaX2K:-MResNZh6S9oLDVn5ueX:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ef2879af65029cf0fd20fbc96124c5844909950b/Elements/src/Geometry/Polygon.cs#L1013)):*

I like to use named tuples in a case like this — List<(int indexOnA, int indexOnB)> — then whoever receives the result can get result.indexOnA instead of Item1
</blockquote></details>

Done.

andrewheumann (approved):
> Reviewed 4 of 4 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/488)* status: all files reviewed, 1 unresolved discussion (waiting on @cdriesler)

___
*[Elements/test/PolylineTests.cs, line 56 at r3](https://reviewable.io/reviews/hypar-io/elements/488#-MRtWN1m4WYing8rXvPg:-MRtWN1m4WYing8rXvPh:b-1s5xm3) ([raw file](https://github.com/hypar-io/elements/blob/991b78341fff80a89319f066cc1d9464a8b29fc3/Elements/test/PolylineTests.cs#L56)):*
> ```Smalltalk
> 
>         [Fact]
>         public void SharedSegments_OpenMirroredSquares_OneResult()
> ```

should be "NoResults" not "OneResult," correct?

cdriesler (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/488)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/test/PolylineTests.cs, line 56 at r3](https://reviewable.io/reviews/hypar-io/elements/488#-MRtWN1m4WYing8rXvPg:-MRtzx4n8qTSE40cg-aG:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/991b78341fff80a89319f066cc1d9464a8b29fc3/Elements/test/PolylineTests.cs#L56)):*

should be "NoResults" not "OneResult," correct? 
</blockquote></details>

Done.

PR #489, opened 2021-01-24
Make Representation derive from Element.

### LINKED ISSUE #450: It should be easier to work with Element Instances
**Is your feature request related to a problem? Please describe.**
It currently involves a lot of extra effort to work with `ElementInstance`s. If you want to, say, get all the centerlines of a bunch of beam instances, you have to:
* get all element instances
* filter them by their base element type
* transform the base element's centerline property to the instance's transform

**Describe the solution you'd like**
Two ideas: 
1. a new method on `Model` to simplify retrieving and filtering instances, such as `Model.GetElementInstancesOfBaseType<Beam>()`
2. a new method on `ElementInstance` to get a property from the base element, optionally applying the instance's transform to a geometric property, `ElementInstance.GetBaseProperty(string propertyName, bool shouldTransform = true)`

**Describe alternatives you've considered**
An alternative to 2. above would be something like `ElementInstance.GetTransformedBase` which would automatically give you the full element represented by the instance, transformed appropriately. The difficulty with this approach is that not all geometric properties should be transformed with the element; in the case of a beam, the `Profile` should not be transformed, but the centerline should. We could also consider some sort of type annotation on properties `[TransformsWithElement]` or something, in which case such a method would be possible. 

**Additional context**
This issue is inspired by comments from @Stuart-Gale:
image

ikeough:
BACKGROUND:
This work is a continuation of work to fix #450 as outlined in #454.

DESCRIPTION:
- `Representation` now inherits from `Element`.
  - This PR adds the `IMigration` interface and the `Migrator` class which can migrate model JSON from previous versions to the current version. A migration has been created to migration pre-0.8.3 representations, adding `Name` and `Id` fields and moving the representation to the top level of the `Elements` object in the JSON.
  - Backwards migration is not supported.
- Element graph
  - Two methods are added to `Element` to facilitate the retrieval of dependent `Element` instances. Currently this is used to ensure that dependent elements are added to the `Model` when the element is added.
  - `Element.UpdateChildren()` updates the internal collection of child elements on an element.
  -  `Element.AllChildren()` returns all children that are in the dependency graph of an object. 
      - Example: An element has a `Representation` which has two `SolidOperation` instances, each of which uses a different `Profile`, `AllChildren` will return the `Representation` and the two `Profile` instances.
  - The recursive child search, which previously lived on `Model` is now moved into `Element.GatherSubelements` and tightened to avoid iterating over numeric, string, and `Guid` types which cannot have elements as descendants.

TESTING:
- All tests pass.
 
CHANGELOG:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/489)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## 0.8.3
+### Added
+- `Model.ElementsVersion`
+
 ### Changed
 - `Representation` now inherits from `Element`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.3
+### Changed
+- `Representation` now inherits from `Element`.
+
 ## 0.8.2
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 0.8.3
 ### Added
 - `Model.ElementsVersion`
+- `Element.UpdateChildren()`
+- `Element.AllChildren()`
 
 ### Changed
 - `Representation` now inherits from `Element`.
--- a/Elements.Serialization.IFC/test/IFCTests.cs
+++ b/Elements.Serialization.IFC/test/IFCTests.cs
@@ -105,7 +105,7 @@ namespace Elements.IFC.Tests
                 this.output.WriteLine(e);
             }
 
-            Assert.Equal(7, newModel.Elements.Values.Count);
+            Assert.Equal(10, newModel.Elements.Values.Count);
             newModel.ToGlTF(ConstructGlbPath("IfcFloor2"));
         }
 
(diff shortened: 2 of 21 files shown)


ikeough:
@andrewheumann The more I work on this, the more I think we need to abandon representations in place, and build the new representations structure around them. This work upgrades representations "in place" to elements, but that means that the JSON now has id references instead of objects which will break older versions of elements parsing this. If we can't come up with an idea how to do this, we can remove the representation upgrade from this PR, keeping the elements graph stuff, and add a new schema like `RepresentationBase` from which we can create our new `Representation` types.

ikeough:
We are discussing putting this work on hold until after we've added function versioning to Hypar.

ikeough:
Recent conversations regarding representations have led to the conclusion that representation is perhaps not required at all and that components responsible for representation can be added to classes directly as properties and rendered appropriately using custom renderers (like ECS services).

PR #491, opened 2021-01-26
run alpha release correctly after main release

cdriesler:
BACKGROUND:
- CICD was failing to release alpha versions after a main release. It checks for the existence of 'alpha' or 'beta' on the previous version before incrementing, which a main release doesn't have, and so it failed.

DESCRIPTION:
- Increments version with `-alpha1` if is not already a beta or alpha release

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/491)


cdriesler:
(I misunderstood what was happening.)

PR #492, opened 2021-01-26
allow alternative mesh deserialization

wynged:
BACKGROUND:
- There is an alternative schema format that is used in the front end that we would like to be able to deserialize when it is in the present on an input_schema.

DESCRIPTION:
- Identify the other schema type by the presence of the "vertices" property.
- If this is the "input mesh" then deserialize into the mesh type in a slightly different way.

TESTING:
- The Roof By Sketch function is proof of this functionality working, it has temporarily been published with a local elements reference.
  
FUTURE WORK:
- We should make an effort on both the Hypar web app and in this library to converge on a single mesh schema representation, as outlines in the note above the converter code.

CHANGELOG:
- [x ] `CHANGELOG.md` is updated as necessary. 

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/492)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.8.3
+
+### Changed
+- Make MeshConverter deserialization more flexible to accommodate a schema used in function `input_schema`.
+
 ## 0.8.2
 
 ### Changed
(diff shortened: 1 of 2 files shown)


wynged:
A good sample workflow is this one.  Both Roof By Sketch and JsonToModel are published on dev using this version of Elements righit now. https://dev.hypar.io/workflows/df482ab3-8244-4198-8702-fb4b3e2e2c60

mattc (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/492)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #494, opened 2021-01-26
Add Profile.UnionAll

andrewheumann:
BACKGROUND:
- There did not exist a method to join multiple regions including their internal voids. This is useful for operations like combining overlapping sets of corridor geometry, which may be simple closed curves or contain inner loops.

DESCRIPTION:
- Adds a static `Profile.UnionAll(IEnumerable<Profile> profiles, double tolerance = Vector3.Epsilon)` method which takes a list of profiles, merges them together using clipper, and constructs new profiles from the results.
- Adds a (fairly barebones) test to ensure this works correctly

TESTING:
- Have tested by running the test
- Have tested in use by referring to this method in some private functions
  
FUTURE WORK:
- We should probably expose other kinds of boolean operations on `Profile`
- We should probably write a few more tests for this
- We might want to expose other solve options for union (more of the clipper configuration stuff)

CHANGELOG:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Added
 
+- `Profile.UnionAll(...)` - Create a new set of profiles, merging any overlapping profiles and preserving internal voids.
 - `Polyline.SharedSegments()` - Enables search for segments shared between two polylines.
 - `Polyline.TransformSegment(...)` - Allows transforms for individual polyline segments. May be optionally flagged as polygon and/or planar motion.
 
--- a/Elements/src/Geometry/Profile.cs
+++ b/Elements/src/Geometry/Profile.cs
@@ -314,5 +314,49 @@ namespace Elements.Geometry
             }
             return Polygon.Contains(allLines, point, out containment);
         }
+
+        /// <summary>
+        /// Perform a union operation on a set of multiple profiles.
+        /// </summary>
+        /// <param name="profiles">The profiles with which to create a union.</param>
+        /// <param name="tolerance">An optional tolerance.</param>
+        /// <returns>A new list of profiles comprising the union of all input profiles.</returns>
+        public static List<Profile> UnionAll(IEnumerable<Profile> profiles, double tolerance = Vector3.EPSILON)
+        {
+            Clipper clipper = new Clipper();
+            foreach (var profile in profiles)
+            {
+                var subjectPolygons = new List<Polygon> { profile.Perimeter };
+                if (profile.Voids != null && profile.Voids.Count > 0)
+                {
+                    subjectPolygons.AddRange(profile.Voids);
+                }
+                var clipperPaths = subjectPolygons.Select(s => s.ToClipperPath(tolerance)).ToList();
+                clipper.AddPaths(clipperPaths, PolyType.ptSubject, true);
+            }
+            PolyTree solution = new PolyTree();
+            clipper.Execute(ClipType.ctUnion, solution, PolyFillType.pftPositive);
+            if (solution.ChildCount == 0)
+            {
+                return null;
+            }
+            var joinedProfiles = new List<Profile>();
+            foreach (var result in solution.Childs)
+            {
+                var perimeter = PolygonExtensions.ToPolygon(result.Contour, tolerance);
+                List<Polygon> voidCrvs = new List<Polygon>();
+                if (result.ChildCount > 0)
+                {
+                    foreach (var child in result.Childs)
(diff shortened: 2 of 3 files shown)


cdriesler (approved):
Looks good! Nothing blocking, just some stuff I want to make sure I understand.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/494#-:-MS0MjIT1gEdT63CmG4N:bshwjvr)* status: all files reviewed, 2 unresolved discussions (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Profile.cs, line 341 at r1](https://reviewable.io/reviews/hypar-io/elements/494#-MS0MR3b4W1KxGz1Bqju:-MS0LTeGBFISApnyHb7d:b-9ifx3g) ([raw file](https://github.com/hypar-io/elements/blob/b1e6ea2f98cc2aec9d7a92043dda92288ec7abb3/Elements/src/Geometry/Profile.cs#L341)):*
> ```Smalltalk
>             if (solution.ChildCount == 0)
>             {
> ```

unsure what's most csharp-y - but would it be better to return an empty list here? I know we can't make the return signature `List<Profile> | null` ...
___
*[Elements/test/ProfileTests.cs, line 66 at r1](https://reviewable.io/reviews/hypar-io/elements/494#-MS0M_WK7qzvjwKMevO0:-MS0M_WLB0HzeYctoWmk:bimn2p) ([raw file](https://github.com/hypar-io/elements/blob/b1e6ea2f98cc2aec9d7a92043dda92288ec7abb3/Elements/test/ProfileTests.cs#L66)):*
> ```Smalltalk
>             var p3 = new Profile(outer3, inner3);
>             var union = Elements.Geometry.Profile.UnionAll(new[] { p1, p2, p3 });
>             foreach (var profile in union)
> ```

Is this to assert that a adding the profiles works as expected?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/494)* status: all files reviewed, 2 unresolved discussions (waiting on @cdriesler)

___
*[Elements/src/Geometry/Profile.cs, line 341 at r1](https://reviewable.io/reviews/hypar-io/elements/494#-MS0MR3b4W1KxGz1Bqju:-MS2u9hrBSl9MbInhBXM:b5f3h6x) ([raw file](https://github.com/hypar-io/elements/blob/b1e6ea2f98cc2aec9d7a92043dda92288ec7abb3/Elements/src/Geometry/Profile.cs#L341)):*
<details><summary><i>Previously, cdriesler (Chuck Driesler) wrote…</i></summary><blockquote>

unsure what's most csharp-y - but would it be better to return an empty list here? I know we can't make the return signature `List<Profile> | null` ...
</blockquote></details>

Lists are nullable, so it's valid to return null for a list type. What you suggest is arguable — I think I lean on the side of "null means something went wrong" which is what I expect this case to mean (no children at the top level) rather than empty list.
___
*[Elements/test/ProfileTests.cs, line 66 at r1](https://reviewable.io/reviews/hypar-io/elements/494#-MS0M_WK7qzvjwKMevO0:-MS2zK3574B5OVTk2fmm:b5wfcuf) ([raw file](https://github.com/hypar-io/elements/blob/b1e6ea2f98cc2aec9d7a92043dda92288ec7abb3/Elements/test/ProfileTests.cs#L66)):*
<details><summary><i>Previously, cdriesler (Chuck Driesler) wrote…</i></summary><blockquote>

Is this to assert that a adding the profiles works as expected?
</blockquote></details>

just to test that it correctly unions the two that overlap, and preserves the one that doesn't overlap. It's a pretty weak test, I'll grant.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/494)* status: all files reviewed, 2 unresolved discussions (waiting on @cdriesler)

___
*[Elements/test/ProfileTests.cs, line 66 at r1](https://reviewable.io/reviews/hypar-io/elements/494#-MS0M_WK7qzvjwKMevO0:-MS2z_Ai3J0CcRxhvkon:b-lwo5lw) ([raw file](https://github.com/hypar-io/elements/blob/b1e6ea2f98cc2aec9d7a92043dda92288ec7abb3/Elements/test/ProfileTests.cs#L66)):*

just to test that it correctly unions the two that overlap, and preserves the one that doesn't overlap. It's a pretty weak test, I'll grant. 
</blockquote></details>

I'm also using the "visual test" aspect — it produces a GLB which I inspected to make sure it looks as expected.

PR #504, opened 2021-02-03
Make Grid1d and Grid2d Serializable

andrewheumann:
BACKGROUND:
- @serenayl wanted to pass grids between functions. Grids were previously not serializable — or at least not without a huge amount of redundant information.

DESCRIPTION:
- Modifies Grid1d and Grid2d to be serializable.
- Adds several properties and methods for serialization purposes. I went this route instead of a custom json converter because custom converters can be a pain. 

The tricky bits are these:

Grid1d only serializes its base curve at the top level, and the `curve` property has been modified to track upwards through its chain of parents until it finds the base curve to use (in a lazy way — it only gets it when it's needed.) 

Grid2d doesn't store its full `Cells` array — it only stores its component axes, and any cells that are *not* single cells at any given level. All the "single cells" at a level are an automatic consequence of the U and V 1d grids, so can be reconstituted from them — then we can inject or insert any modified subgrids.  

Code generation needs to be updated so that it won't codegen classes for Grid1d, Grid2d, IndexedCell or Domain1d. 

TESTING:
- New tests have been added, but you can also take any constructed Grid1d or 2d, convert it to json and back (and compare the resulting cells) to make sure nothing was lost in translation.
- Made sure you could reference a schema like https://prod-api.hypar.io/schemas/Grid2dElement and it would codegen correctly.
  
FUTURE WORK:
- Need to test these against the schemas in schema manager

REQUIRED:
- [X] The docs are up to date. To update the docs run `build_docs.sh` locally, and commit the results.
 (this shouldn't change any function signatures

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/504)


serenayl (approved):
> Reviewed 97 of 98 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/504)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #512, opened 2021-02-06
Merge duplicate vertices on polygon creation

andrewheumann:
BACKGROUND:
- A frequent cause of error for hypar devs and contributors alike is the validation on polygon creation. Various operations such as clipper booleans etc may result in duplicate vertices.

DESCRIPTION:
- Modifies Polyline and Polygon Validation logic as follows:
    - remove duplicate vertices in the PostConstruct action
    - move most validation into the PostConstruct action, so as to avoid duplicate calculations 
    - Prevent the polyline validator for running on polygon creation; previously, we would run the validation methods twice for every polygon (once in the Polygon PostConstruct and once in the Polyline PostConstruct.)

TESTING:
- have modified the previous "same vertices throws exception" test to "same vertices removes duplicates"
- Have tested this on real functions (customer work)
- Have validated that there is no difference in performance on polygon creation — in fact, creating 10,000 polygons with 8 vertices apiece is actually faster now that the validation step is not duplicated.  
  

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/512)


Elements/src/Geometry/Vector3.cs
```diff
@@ -650,6 +650,38 @@ public static double CCW(Vector3 a, Vector3 b, Vector3 c)
...
+        {
+            List<Vector3> newList = new List<Vector3> { vertices[0] };
+            for (int i = 1; i < vertices.Count; i++)
+            {
+                var vertex = vertices[i];
+                var prevVertex = newList[newList.Count - 1];
+                if (!vertex.IsAlmostEqualTo(prevVertex))
+                {
+                    // if we wrap, and we're at the last vertex, also check for a zero-length segment between first and last.
+                    if (wrap && i == vertices.Count - 1)
```
ikeough, line 670: I think this can be avoided by simply taking your loop through `vertices.Count` and checking at the last index like so:
```csharp
var vertex = i == vertices.Count ? vertices[0] : vertices[i];
...
if(i == vertices.Count - 1 && !wrap)
{
  continue;
}
```

Elements/src/Validators/Validators.cs
```diff
@@ -348,15 +348,21 @@ public class PolylineValidator : IValidator
 
         public void PostConstruct(object obj)
         {
+            if(obj is Polygon) {
+                // we don't need to validate twice — 
+                // the Polygon PostConstruct validator will handle all of this correctly.
+                return;
+            }
+            var polyline = obj as Polyline;
+            polyline.Vertices = Vector3.RemoveSequentialDuplicates(polyline.Vertices);
```
ikeough, line 357: I always worried that we'd put too much logic into these `PostConstruct`/`PreConstruct` methods and forget that these constructors are _also_ used for JSON deserialization. Is it logical to have a deserialization process that might mutate the incoming polyline? Assuming that Elements is doing most of the creation of these Polylines (perhaps a big assumption), do we really want to do this sequential vertex removal _again_ when we deserialize?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/512)* status: 0 of 28 files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[Elements/src/Geometry/Vector3.cs, line 670 at r2](https://reviewable.io/reviews/hypar-io/elements/512#-MSsu8qNu0aNLEDr8oTS-r2-670:-MTBPYT7A8Ocrt5lJRse:b910adb) ([raw file](https://github.com/hypar-io/elements/blob/3636f86529797437385d63046641c11f76fb3bce/Elements/src/Geometry/Vector3.cs#L670)):*

I think this can be avoided by simply taking your loop through `vertices.Count` and checking at the last index like so:
```csharp
var vertex = i == vertices.Count ? vertices[0] : vertices[i];
...
if(i == vertices.Count - 1 && !wrap)
{
  continue;
}
```
</blockquote></details>

I'm not sure I understand — I don't want to skip the last vertex if we're not wrapping. if we ARE wrapping, I have to check two things on the last vertex — whether it's the same as the previous, and whether it's the same as the first, and only add it if neither are true. Maybe I'm missing something, but I don't see how this achieves that.
___
*[Elements/src/Validators/Validators.cs, line 357 at r2](https://reviewable.io/reviews/hypar-io/elements/512#-MSsu8qNu0aNLEDr8oTT-r2-357:-MTBLTXuCTU3xScZv6wb:b-f9pezk) ([raw file](https://github.com/hypar-io/elements/blob/3636f86529797437385d63046641c11f76fb3bce/Elements/src/Validators/Validators.cs#L357)):*

I always worried that we'd put too much logic into these `PostConstruct`/`PreConstruct` methods and forget that these constructors are _also_ used for JSON deserialization. Is it logical to have a deserialization process that might mutate the incoming polyline? Assuming that Elements is doing most of the creation of these Polylines (perhaps a big assumption), do we really want to do this sequential vertex removal _again_ when we deserialize?
</blockquote></details>

A reasonable concern. 
A possible alternative would be to separate out the runtime constructor (used by a function, or by other methods in Elements) from the serialization constructor — validation / error correction always happens when something is made internally, but not when it is deserialized. But this runs the risk that elements coming from other sources (eg python functions, maybe) don't have the same validations in place, and we run into trouble. 

I think it's safer to err on the side of "always eliminating duplicates," especially because this is not a particularly slow operation (as mentioned above, this code makes polygons faster than prior to this PR, due to the removal of duplicate validation steps — so deduplication isn't free, but it's no more expensive than what we've already been doing). I understand a little bit of "ick" about mutating on deserialization, but I am struggling to imagine a case where a user or creator would be surprised or frustrated by cleanup like this happening at the interfaces. Worst case, a scenario like this might dovetail nicely with previously proposed "Disable validation on construction" flag — if it's truly an issue (for performance, or for preservation of some important information on the creator's behalf), and they REALLY know what they're doing, they can turn it off.

Elements/src/Geometry/Vector3.cs
ikeough: I've updated the code sample above. I might be mis-understanding, but I believe this removes the inner `IsAlmostEqual` check.

Elements/src/Validators/Validators.cs
ikeough: I said all of the above, without suggesting an alternative, which you then described and its sounds worse from a usability perspective to put this into its own constructor which a developer would have to know how to find. Let's leave your change as is.

ikeough (approved):
My code snippet might be misunderstanding the problem and is a very small optimization even if it is correct, so feel free to ignore and proceed.

andrewheumann (commented):
> Dismissed @ikeough from a discussion.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/512)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @ikeough)

___
*[Elements/src/Geometry/Vector3.cs, line 670 at r2](https://reviewable.io/reviews/hypar-io/elements/512#-MSsu8qNu0aNLEDr8oTS-r2-670:-MTHItNu00ItZ-jcdU6F:bu1yo1a) ([raw file](https://github.com/hypar-io/elements/blob/3636f86529797437385d63046641c11f76fb3bce/Elements/src/Geometry/Vector3.cs#L670)):*

I've updated the code sample above. I might be mis-understanding, but I believe this removes the inner `IsAlmostEqual` check.
</blockquote></details>

I think this suffers from a risk of compounding error. since you're always comparing to the previous vertex (instead of the last "approved" vertex) if you had a chain of vertices that were each epsilon apart, you might wind up removing all but 1 of them, resulting in an overall vertex deviation greater than epsilon. Keeping this as-is for this reason and because I find it easier to reason about, but willing to revisit if there's a strong reason. 
___
*[Elements/src/Validators/Validators.cs, line 357 at r2](https://reviewable.io/reviews/hypar-io/elements/512#-MSsu8qNu0aNLEDr8oTT-r2-357:-MTHIoT_DIZpCflXBl1y:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/3636f86529797437385d63046641c11f76fb3bce/Elements/src/Validators/Validators.cs#L357)):*

I said all of the above, without suggesting an alternative, which you then described and its sounds worse from a usability perspective to put this into its own constructor which a developer would have to know how to find. Let's leave your change as is.
</blockquote></details>

Done.

PR #516, opened 2021-02-10
fix for null `Voids` case

andrewheumann:
BACKGROUND:
- I missed that the `Voids` property on a Profile is sometimes null in the `Segments` method, so this was throwing exceptions sometimes.

DESCRIPTION:
- fixes the Profile.Segments() method to tolerate a null `Voids` array

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/516)


cdriesler (approved):
> Reviewed 1 of 1 files at r1, 24 of 24 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/516)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #517, opened 2021-02-10
new Bbox3(polygon) throws exception


andrewheumann:
**Describe the bug**
creating a bbox3 from a polygon or a profile immediately throws a validation exception, because the Bbox3(profile) constructs an illegal vector with infinite components.
![image](https://user-images.githubusercontent.com/31935763/107515897-a8ac2180-6b79-11eb-82e7-726297a46763.png)

![image](https://user-images.githubusercontent.com/31935763/107515935-b497e380-6b79-11eb-9b7b-0a231ae17d52.png)

PR #518, opened 2021-02-10
fix for bbox from polygon

andrewheumann:
BACKGROUND:
- #517

DESCRIPTION:
- Fixes for exception on calculating bounding box from polygon. Two things were wrong:
   - the Bbox(profile) constructor created an illegal Vector3, immediately throwing an exception on validation
   - the Bbox(profile) constructor was called for a polygon thanks to implicit conversion from Polygon => Profile. Since the implicit conversion calls the constructor directly, the validator that normally would force an empty list into voids was not called (possibly also related to why I ran into #516)

TESTING:
- added a new test, which used to throw an exception, and now succeeds
  
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/518)


cdriesler (approved):
> Reviewed 4 of 4 files at r1, 24 of 24 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/518)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #519, opened 2021-02-10
Enable a user to temporarily suspend geometry validation

andrewheumann:
BACKGROUND:
- per discussion w @ikeough, this is a "pro user only" feature that enables a developer to temporarily suspend the action of the validation functions for geometry. This is necessary because sometimes our own methods cause internal exceptions, which a developer cannot otherwise circumvent. For example, a Polygon.Difference can yield an invalid polygon according to our standards (even after automated duplicate vertex removal, per #512). While of course ideally we want to rely on this as little as possible — catch invalid results at the source and clean them up — we cannot guarantee that our own code will never produce an invalid geometry. There should be an escape hatch for a developer if they want to proceed with possibly-imperfect geometry.    

DESCRIPTION:
- adds a static `DisableValidationOnConstruction` boolean flag on `Validator` 
- if true, the Validator retriever will fail to return a validator, thereby skipping all validation. 

TESTING:
- Added a Validation test that verifies that exceptions are thrown normally, but bypassed if this flag is true. 
  
FUTURE WORK:
- We should take note of situations in which this is necessary and aim to remove or reduce them. 

REQUIRED:

COMMENTS:
- I keep delaying adding this since I keep hoping that I can root out whatever underlying problem, but it's whack-a-mole. A lot of the times I need to do this can be attributed to Clipper, so the `ToPolygon` method for clipper geometry would be a good place to start looking at longer-term cleanup solutions for removing, eg, self-intersections.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/519)


ikeough (approved):
Stylistic suggestions only.

> Reviewed 4 of 28 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/519#-:-MTXv2ZU3skkmPvxL0uL:bhnydgc)* status: 4 of 31 files reviewed, 3 unresolved discussions (waiting on @andrewheumann)

___
*[Elements/src/Validators/Validator.cs, line 39 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXtUG60bwe87QFhMHA:-MTXtUG60bwe87QFhMHB:bd1h0xr) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L39)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Use this to disable geometry validation. This can cause lots of unexpected consequences; only use it if you know what you're doing.
> ```

We haven't come up with a standard format for `boolean` comments. I've been doing a fair amount of framing these as a question. Example: "Should geometry validation be disabled during construction?". Another popular format is "A flag indicating whether geometry validation should be disabled during construction." I think we could state the warning associated with this flag more succinctly to the effect of "Disabling validation can have unforeseen consequences. Use with caution."
___
*[Elements/src/Validators/Validator.cs, line 42 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXucew3dJVoqgbD5fA:-MTXucew3dJVoqgbD5fB:b-1l7utv) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L42)):*
> ```Smalltalk
>         /// It's probably a good idea to turn this back to false when you're done circumventing checks for a tricky operation.
>         /// </summary>
>         /// <value></value>
> ```

Remove empty value tag.
___
*[Elements/src/Validators/Validator.cs, line 88 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXuY7Z7zEx-qCoyEOV:-MTXuY7_05aOi2AFQWga:b7ss6bw) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L88)):*
> ```Smalltalk
>             {
>             }
> ```

Whitespace.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/519)* status: 1 of 31 files reviewed, 3 unresolved discussions (waiting on @ikeough)

___
*[Elements/src/Validators/Validator.cs, line 39 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXtUG60bwe87QFhMHA:-MTfbK8XC_GC9LuoVdya:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L39)):*

We haven't come up with a standard format for `boolean` comments. I've been doing a fair amount of framing these as a question. Example: "Should geometry validation be disabled during construction?". Another popular format is "A flag indicating whether geometry validation should be disabled during construction." I think we could state the warning associated with this flag more succinctly to the effect of "Disabling validation can have unforeseen consequences. Use with caution."
</blockquote></details>

Done.
___
*[Elements/src/Validators/Validator.cs, line 42 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXucew3dJVoqgbD5fA:-MTfbbnV4FuQ6wnNTVQN:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L42)):*

Remove empty value tag.
</blockquote></details>

Done.
___
*[Elements/src/Validators/Validator.cs, line 88 at r2](https://reviewable.io/reviews/hypar-io/elements/519#-MTXuY7Z7zEx-qCoyEOV:-MTfbiW135X082h77DEL:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/63fae70f4e3e72bcb5f20c522d5e9e4af2c60c3e/Elements/src/Validators/Validator.cs#L88)):*

Whitespace.
</blockquote></details>

Done.

PR #520, opened 2021-02-11
make ToProfiles recursive, and fix common exception

andrewheumann:
BACKGROUND:
- Certain Profile boolean operations, like difference, were missing internal shapes in their results. This was due to simply not processing the `PolyTree` returned from clipper in a recursive manner.
- There was a consistent failure case coming from Clipper boolean operations that boiled down to a "this segment is too short" being thrown from "CheckSelfIntersectionsAndThrow". Too short segments are automatically removed thanks to #512, but this validation method transforms the segments it receives — and so if a segment was just slightly larger than Epsilon in length, it could fail the "duplicate point" check after transformation because our IsAlmostEqualTo method was not rotation-invariant.

DESCRIPTION:
- Refactors the code for converting a PolyTree Clipper result into profiles into a `ToProfiles` recursive method. For more on this structure, see http://www.angusj.com/delphi/clipper/documentation/Docs/Units/ClipperLib/Classes/PolyTree/_Body.htm
- Changes `Vector3.IsAlmostEqualTo` to use a squared distance metric instead of comparing individual components, in order to fix a common exception case with results from clipper operations. 

TESTING:
- Added a test testing deeply nested difference. Before this change, the result was this:
image
 After the change, it was this: 
image
- Also tested on a real function (subtracting a bunch of corridors from a floor plate shape)
- Added a test verifying that the new "isAlmostEqualTo" logic fixes the problem I was experiencing

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/520)


cdriesler (approved):
> Reviewed 5 of 5 files at r1, 23 of 23 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/520)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #522, opened 2021-02-14
Add planar texture coordinates.

ikeough:
BACKGROUND:
The `Material.Texture` property allows a developer to specify a texture, but getting textures to show up required assigning texture coordinates manually. It's typical in modeling engines that basic texture coordinates are generated automatically.

DESCRIPTION:
- This PR adds planar texture coordinates to meshes generated from `Solid` and `CSG` objects.
- This PR fixes an error where materials with `unlit=true` would not also work with materials with textures.
- This PR handles textures with transparency by testing whether the texture has an alpha channel and properly applying the blend mode.

TESTING:
- The reviewer can create a material with a texture and apply it to any `GeometricElement`
  
FUTURE WORK:
- Texture support is limited to unscaled planar projection.

REQUIRED:

image

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/522)


Elements/src/Geometry/CsgExtensions.cs
```diff
@@ -88,23 +89,37 @@ internal static void Tessellate(this Csg.Solid csg, ref Mesh mesh, Transform tra
...
-                var n = (b - a).Cross(c - a).Unitized();
+                var tmp = (b - a).Unitized();
+                var n = tmp.Cross(c - a).Unitized();
+
+                // Calculate the texture space basis vectors
+                // from the first triangle. This is acceptable
+                // for planar faces.
+                // TODO: Update this when we support non-planar faces.
+                // https://gamedev.stackexchange.com/questions/172352/finding-texture-coordinates-for-plane
+                e1 = n.Cross(n.IsParallelTo(Vector3.XAxis) ? Vector3.YAxis : Vector3.XAxis).Unitized();
```
cdriesler, line 100: I see this usage pattern repeats, and it comes with a special note. Would it be worth extracting it to a `Vector3` method with a disclaimer in the docs?

```
var ( e1, e2 ) = Vector3.ComputeBasis(n)
```

We could test just the planar cases, for now, then add non-planar when necessary.

ikeough: That is a nice suggestion. I'm going to do it.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/522)* status: 0 of 37 files reviewed, 1 unresolved discussion (waiting on @cdriesler)

___
*[Elements/src/Geometry/CsgExtensions.cs, line 100 at r10](https://reviewable.io/reviews/hypar-io/elements/522#-MTToshqHVM4-b7M5irJ-r10-100:-MThiLJIDyLyS6UV-gEt:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/160ed3f613499fe7e74136bc1f678d760c7a5fed/Elements/src/Geometry/CsgExtensions.cs#L100)):*

That is a nice suggestion. I'm going to do it.
</blockquote></details>

Done.

cdriesler (approved):
Approved with one caveat -

I'm not a texture math expert. This all tracks with what I understand, and was confirmed working through use, though.

PR #523, opened 2021-02-16
add a validator for triangles to require 3 non-duplicate vertices

wynged:
BACKGROUND:
- code that works with meshes found invalid triangles.

DESCRIPTION:
- Add a validator for triangles.
- Use the validator in the overloaded constructor.
- Add tests of this validator in action.

TESTING:
- Tests pass
  
FUTURE WORK:
- could do more validation for normals, but currently there are places in the code where normals are corrected anyways, so I don't want to go to far right now.

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/523)


andrewheumann (commented):
Would love to understand more about why this is necessary. In most software meshes are an extremely forgiving, crunchy representation format; introducing this validator increases the likelihood that we will be unable to accept meshes from such sources. What happens if you leave a degenerate triangle alone — does it throw an exception elsewhere? 

I am being extra cautious here because Validation has been the cause of 80% of my Elements headaches. We should carefully weigh the benefit with the risk of adding additional validators. Validators just cause your function to break, full stop, and a function author has a difficult time fixing this problem, especially if the mesh is generated from inside Elements itself (e.g. can't stick a try catch around the relevant part because they don't control the code). My general preference would be to do one of the following:
* if there's an obvious, fast way to correct an invalid case in a mesh, just do that whenever that case is encountered, allowing execution to proceed (eg just remove the bad face)
* try to make the code that fails when it encounters a bad triangle more robust. This may be out of the question if a degenerate triangle has a cascading effect throwing many issues in many downstream places. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/523#-:-MTjpFQs7vYlPXma5Ub3:b-x85wl9)* status: 0 of 4 files reviewed, all discussions resolved

wynged (commented):
changed to doing this test when adding a triangle to mesh using utility methods

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/523#-:-MTlWyN6DJrJcup1afiD:b-laco77)* status: 0 of 3 files reviewed, all discussions resolved

andrewheumann (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 5 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/523)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @wynged)

___
*[Elements/src/Geometry/Mesh.cs, line 241 at r3](https://reviewable.io/reviews/hypar-io/elements/523#-MTlZg0IEXI1N_m00y2X:-MTlZg0IEXI1N_m00y2Y:bq380hl) ([raw file](https://github.com/hypar-io/elements/blob/0278c6b5f7062ec8656a2305dc0b6f00ce5da50f/Elements/src/Geometry/Mesh.cs#L241)):*
> ```Smalltalk
>             if (HasDuplicatedVertices(t, out Vector3 duplicate))
>             {
>                 throw new ArgumentException($"Not a valid Triangle.  The {duplicate} vertex position is duplicated.");
> ```

seems like we should use the same exception in both places. and while I'm being nitpicky... I'd say that "Duplicate vertex at {duplicate}" is a little clearer and leaves the big old "toString" at the end of the message.

andrewheumann (approved):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/523)* status: all files reviewed, 1 unresolved discussion (waiting on @wynged)

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/523)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Mesh.cs, line 241 at r3](https://reviewable.io/reviews/hypar-io/elements/523#-MTlZg0IEXI1N_m00y2X:-MTlcfSEDHrbl-Gg3jBk:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/0278c6b5f7062ec8656a2305dc0b6f00ce5da50f/Elements/src/Geometry/Mesh.cs#L241)):*

seems like we should use the same exception in both places. and while I'm being nitpicky... I'd say that "Duplicate vertex at {duplicate}" is a little clearer and leaves the big old "toString" at the end of the message. 
</blockquote></details>

Done.

Issue #509, opened 2021-02-05
docfx cicd blocked by master branch protections


cdriesler:
On merge to `master`, the `build-docs` step runs fine until it has to commit the new docs:

![0205_error](https://user-images.githubusercontent.com/20760128/107055810-ac1d6280-679f-11eb-9fd0-04d905c55d91.PNG)

We should either find a way to allow a 'bot user' through, or come up with another solution.

The usual alpha build and release step is still working, even if the repo shows an angry red X for now.

mattc:
What about doing this on push-to-branch or PR? I'll want to review and/or correct the docs sometimes before merging.

cdriesler:
Fair call, will have a PR with this adjustment in shortly.

cdriesler:
Resolved in #510

PR #525, opened 2021-02-17
Improve docfx ci (again)

cdriesler:
BACKGROUND:
- Already annoying myself with the current docfx ci integration that pollutes rolling commit history.

DESCRIPTION:
- Replaces on-branch commits with an action that will build and push just the docs to their own branch.

TESTING:
- Used a copy of the workflow that ran on branch pushes. You can see the result [here](https://github.com/hypar-io/Elements/tree/docs).
  
FUTURE WORK:
- Once merged, all we have to do is point the repo github pages settings to the new branch.
- We can also delete `/docs` from the tracked repo, since the files are all managed by the action.

NOTES:
- @mattc made a comment ( #509 ) that he'd like to review docs before merging sometimes. Since these are auto-generated, is it enough to inspect annotation changes?

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/525)


mattc (commented):
Stray thought: What if we stopped committing the generated pages to code at all? Or if we do, make it a totally different repo?

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/525#-:-MTkxBAZ3Bm3p3STYdIt:biwevyh)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @ikeough)

ikeough:
Just reading @mattc's comment above, and I think I agree. Running docfx means that every time I go to push a local change, I now have to do a pull first because docfx ran previously as the result of some other commit and my local is now behind master. This is a small but annoying thing.

cdriesler:
Agreed. How do we feel about this branch-strategy PR as a step, and I look into how we can pull the generated docs out of commit altogether?

cdriesler:
Closing in favor of impending work to do a build-push-docs style action instead of commit-based.

PR #503, opened 2021-02-03
Create a Grid2d from a polygon, origin, and u and v directions

serenayl:
BACKGROUND:
- Custom grid functions required more fine control of Grid2d work, especially in being able to create a bounded grid from an origin and UV directions, and automatically expanding that domain to encompass the boundary.

DESCRIPTION:
- Adds additional functionality to `Line.ExtendTo(IEnumerable<Line>)` to specify whether we want to take the maximum extension or whether we want to stop at the first intersection found.
- Adds `Line.ExtendTo(Polygon)` to disambiguate in that situation.
- Adds Grid2d constructors for origin and direction, and from existing UV Grid1d that has been pre-calculated.

TESTING:
- Try creating Grid2d with origin and UV directions. It should expand to fill the boundaries given, including if the UV vectors are skewed.
  
FUTURE WORK:
- N/A

REQUIRED:
- [x] All changes are up to date in `CHANGELOG.md`.
- [ ] ~~The docs are up to date. To update the docs run `build_docs.sh` locally, and commit the results.~~

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/503)

PR #526, opened 2021-02-17
fix for subcell splitting at point

andrewheumann:
BACKGROUND:
- I noticed while reviewing another PR (#503) that sub cells did not split properly with `SplitAtPoint`

DESCRIPTION:
- Adds a private version of SplitAtOffset that takes into account the curve's full domain, since this is what SplitAtPoint was assuming was happening.
- This ensures that splits happen at the correct points

TESTING:
- Added a new test which validates that splits occur in the correct place, even with skewed grid axes:
image

REQUIRED:

COMMENTS:
- Let's just verify that this fixes the problem I identified in your PR — if so we can merge them in any order.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/526)


serenayl (approved):
> Reviewed 1 of 3 files at r1, 28 of 28 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/526)* status: 29 of 31 files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Spatial/Grid1d.cs, line 320 at r2](https://reviewable.io/reviews/hypar-io/elements/526#-MTpjntT6r7MqcBwNVgE:-MTpjntUCRC0Zdzp0JT1:b-6xvcqr) ([raw file](https://github.com/hypar-io/elements/blob/27bcec835e8d5ab728dc8e728cee7e83fbbe04cf/Elements/src/Spatial/Grid1d.cs#L320)):*
> ```
> private void InternalSplitAtOffset
> ```
Can we add a comment for why we added this and how it's different than the plain `SplitAtOffset`?

PR #527, opened 2021-02-18
Profile.Split()

andrewheumann:
BACKGROUND:
- For space planning applications, needed a way to take a polygon or profile and split it with an overlapping polyline. 
- Clipper does not support this OOTB

DESCRIPTION:
- Adds a `Profile.Split()` method
- Adds some safety checks and improvements to reduce the likelihood of a catastrophic clipper failure
- Corrects a bug where some `ToProfiles` were called without tolerance, meaning if a tolerance other than the default was supplied, they would come back scaled wrong
- update Vector3.RemoveSequentialDuplicates to take a tolerance value

TESTING:
- Polygon split test verifies that a split is executed correctly and returned with appropriate areas:
image

  
FUTURE WORK:
- We should probably write our own, proper polygon split method rather than relying on this clipper kludge, but thinking about how to do this gave me agita.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/527)


cdriesler (approved):
First reaction: "We have to be able to do this a better way! Turn the line into a clip region!"

Then:



Future work note heard loud and clear. Not jealous of the person that has to solve this problem.

This test is also super brave, cool to see it works.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/527#-:-MTryuNVB3tqfeoUzY-k:bmauvam)* status: 0 of 37 files reviewed, all discussions resolved

Issue #528, opened 2021-02-19
Drawing a very small circle results in an ArithmeticOverflow exception


andrewheumann:
**Describe the bug**
`Model.AddElement(new Circle(Vector3.Origin, 0.03));` results in an exception:
```
System.OverflowException : Arithmetic operation resulted in an overflow.
    Stack Trace:
       at Elements.Geometry.Arc.GetSampleParameters(Double startSetback, Double endSetback) in /Users/andrewheumann/Dev/elements/Elements/src/Geometry/Arc.cs:line 145
   at Elements.Geometry.Arc.RenderVertices() in /Users/andrewheumann/Dev/elements/Elements/src/Geometry/Arc.cs:line 159
   at Elements.Serialization.glTF.GltfExtensions.GetRenderDataForElement(Element e, Gltf gltf, Dictionary`2 materialIndexMap, List`1 buffers, List`1 allBuffers, List`1 schemaBuffers, List`1 bufferViews, List`1 accessors, List`1 materials, List`1 textures, List`1 images, List`1 samplers, List`1 meshes, List`1 nodes, Dictionary`2 meshElementMap, Dictionary`2 meshTransformMap, List`1 lines, Boolean drawEdges) in /Users/andrewheumann/Dev/elements/Elements/src/Serialization/glTF/GltfExtensions.cs:line 952
   at Elements.Serialization.glTF.GltfExtensions.InitializeGlTF(Model model, List`1& allBuffers, Boolean drawEdges) in /Users/andrewheumann/Dev/elements/Elements/src/Serialization/glTF/GltfExtensions.cs:line 731
   at Elements.Serialization.glTF.GltfExtensions.SaveGlb(Model model, String path, Boolean drawEdges) in /Users/andrewheumann/Dev/elements/Elements/src/Serialization/glTF/GltfExtensions.cs:line 615
   at Elements.Serialization.glTF.GltfExtensions.ToGlTF(Model model, String path, Boolean useBinarySerialization, Boolean drawEdges) in /Users/andrewheumann/Dev/elements/Elements/src/Serialization/glTF/GltfExtensions.cs:line 59
```

Issue #380, opened 2020-08-27
Visual Center of Polygons & Profiles


wynged:
**Is your feature request related to a problem? Please describe.**
I want to find a point that is near center  of a shape, but is required to actually be inside the polygon.  Centroid does not satisfy this.

**Describe the solution you'd like**
Something like this VisualCenter https://blog.mapbox.com/a-new-algorithm-for-finding-a-visual-center-of-a-polygon-7c77e6492fbc 
This is a code base where it is implemented. https://github.com/mapbox/polylabel 

**Describe alternatives you've considered**
I'm going to use a workaround where I cast a line from the centroid and trim that line with the surrounding polygon.  The midpoint of this trimmed line is guarunteed to be inside the polygon.

**Additional context**

ikeough:
Substantially solve in #529. We can reopen if we think we want an exact solution.

PR #529, opened 2021-02-19
find a point interior to a polygon

andrewheumann:
BACKGROUND:
- often you want a point internal to a polygon (eg for other inclusion tests) and centroid won't cut it, since it could conceivably fall outside. 
- Partially addresses #380 — not quite a visual center, but guaranteed to be internal with much less work. 
- in the course of writing a test, found another bug #528, fixing that one too. 

DESCRIPTION:
- Adds a Polygon.PointInternal method that returns a point guaranteed to be internal to the polygon.
- fix overflow exception from trying to draw an arc segment smaller than the minimum chord length.

TESTING:
- added a test with a very concave polygon
image

  
FUTURE WORK:

REQUIRED:

COMMENTS:
- If this doesn't get merged before 0.8.3 release I will just bump it up to 0.8.4, not urgent.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/529)


wynged (changes requested):
> Reviewed 3 of 4 files at r1, 28 of 28 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/529)* status: all files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/test/PolygonTests.cs, line 727 at r3](https://reviewable.io/reviews/hypar-io/elements/529#-MU9_pQ63gNDCx4VcQ9r:-MU9_pQ63gNDCx4VcQ9s:b-kdebl8) ([raw file](https://github.com/hypar-io/elements/blob/e43e1882a86485e7f67b8d0f1923f915f676962c/Elements/test/PolygonTests.cs#L727)):*
> ```Smalltalk
>         {
>             Name = "PointInternal";
>             var extremelyConvexPolygon = JsonConvert.DeserializeObject<Polygon>("{\n  \"Vertices\": [\n    {\n      \"X\": 5.894565217391305,\n      \"Y\": 0.0,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 5.894565217391305,\n      \"Y\": 0.69347826086956488,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 0.19082958701549974,\n      \"Y\": 0.13222235119778919,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 0.0,\n      \"Y\": 2.3964022904166224,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 6.2310064053894925,\n      \"Y\": 3.9731847432442322,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 5.9682093299182242,\n      \"Y\": 4.5513383092810225,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": -0.085624933213594254,\n      \"Y\": 2.4045397339325851,\n      \"Z\": 0.0\n    },\n    {\n      \"X\": 0.0,\n      \"Y\": 0.0,\n      \"Z\": 0.0\n    }\n  ]\n}");
> ```

I don't think that many of our tests use this de serialization pattern to get a test shape, and it shouldn't be too hard to make a very concave polygon that would be interesting.. .can we make these vertices explicit?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/529)* status: 31 of 32 files reviewed, 1 unresolved discussion (waiting on @wynged)

___
*[Elements/test/PolygonTests.cs, line 727 at r3](https://reviewable.io/reviews/hypar-io/elements/529#-MU9_pQ63gNDCx4VcQ9r:-MUJf8J54IyT8FwwT-BT:b95upg0) ([raw file](https://github.com/hypar-io/elements/blob/e43e1882a86485e7f67b8d0f1923f915f676962c/Elements/test/PolygonTests.cs#L727)):*

I don't think that many of our tests use this de serialization pattern to get a test shape, and it shouldn't be too hard to make a very concave polygon that would be interesting.. .can we make these vertices explicit?
</blockquote></details>

My feeling was looking at the vertices doesn't do you much good anyway, and makes the test super long — but done.

wynged (approved):
> Reviewed 1 of 1 files at r4, 24 of 24 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/529)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Elements/test/PolygonTests.cs, line 727 at r3](https://reviewable.io/reviews/hypar-io/elements/529#-MU9_pQ63gNDCx4VcQ9r:-MUKsoqFAPg01MGDKusA:b-xc6rlp) ([raw file](https://github.com/hypar-io/elements/blob/e43e1882a86485e7f67b8d0f1923f915f676962c/Elements/test/PolygonTests.cs#L727)):*

My feeling was looking at the vertices doesn't do you much good anyway, and makes the test super long — but done. 
</blockquote></details>

yea, I might not have used the exact numbers because they're kinda wacky, but this is still better I think even if just from a formatting perspective.

PR #531, opened 2021-02-22
Everything ifc

ikeough:
BACKGROUND:

DESCRIPTION:

TESTING:
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.


ikeough:
One small piece I still need to do. I added a method `FromIFC2` to get around the existing `FromIFC`. If we agree with the direction of this work, the original method will be removed and replaced with the new one.

ikeough:
After conversation with @andrewheumann, the following changes are proposed. 
- Add a new base type `BuildingElement`, from which Things like `Wall` and `Beam` can inherit. Add the `Openings` property to this type. 
- Move building elements to a separate library `Elements.BuildingElements`, which contributes to keeping the core small. 
- Add the `BuildingElement.IfcEntityType` property to allow a user to define which type of IFC Entity they would like to map to during export. This will allow any building element type to map to any IFC entity type, but will also run the risk of creating illegal IFC entities.

ikeough:
This is waiting on work to support `IndexedPolycurve`.

PR #532, opened 2021-02-23
Color scale interpolation

serenayl:
BACKGROUND:
- The original implementation of color range tended to skew colors to the high range of the scale when we constructed it with `colorCount` other than 10.
- I had some desire to add new functionality to ColorScale that allows intentionally uneven interpolation

DESCRIPTION:
- Deprecates discrete banded color scale constructor in favor of a default smooth even interpolation
- Removes limitation for color scale to be between 0 and 1
- Makes obsolete the previous default constructor
- Adds logic for custom color values corresponding to each anchor color

TESTING:
- Try using AnalysisMesh or make panels interpolated along a scale and check for visual match.
- Pass in a corresponding list of color values and check for intentional skew along your gradient
  
FUTURE WORK:
- N/A

REQUIRED:

COMMENTS:
- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/532)


ikeough (changes requested):
> Reviewed 1 of 3 files at r1, 5 of 28 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/532)* status: 6 of 31 files reviewed, 6 unresolved discussions (waiting on @serenayl)

___
*[Elements/src/Analysis/ColorScale.cs, line 27 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULDnnD4x5SejOfm5qK:-MULDnnD4x5SejOfm5qL:b-ecp1ar) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L27)):*
> ```Smalltalk
>         /// The domain of the scale
>         /// </summary>
>         public List<Domain1d> Domains { get; } = null;
> ```

Will users need access to the Domains after construction? I think we could get away with this being 
___
*[Elements/src/Analysis/ColorScale.cs, line 35 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MUL5uh0DOwx63yfcPQp:-MUL5uh0DOwx63yfcPQq:b-j7kczv) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L35)):*
> ```Smalltalk
>         /// <param name="domains">The domains which the colors map to</param>
>         [JsonConstructor]
>         public ColorScale(List<Color> colors, List<Domain1d> domains)
> ```

The Dynamo color range tool used this same methodology and almost everyone is confused by it. That's not to say that we can't get it right, it's just hard. For example, I provide `[Red, Blue, Green] and `[0,1],[0,1],[0,1]` (typical for Dynamo users). What happens? I think I just have duplicate domains, and if I do `GetColorForValue(0.5)` I get purple. Is that correct?

Maybe this constructor is really meant for people who know exactly what they're doing? If so, can we help with a check that the domains are exclusive and a warning if they are not?
___
*[Elements/src/Analysis/ColorScale.cs, line 47 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULE_u393A2LeuXrD2A:-MULE_u393A2LeuXrD2B:b-xtabw8) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L47)):*
> ```Smalltalk
>         /// <param name="colorCount">The number of colors in the final color scale
>         /// These values will be interpolated between the provided colors.</param>
>         [Obsolete("colorCount is no longer required, and this constructor will simply create an evenly interpolated scale between 0 and 1.")]
> ```

Before we obsolete this, I want to make sure that we can replicate this capability. This was originally designed so that the user could say, I want a gradient between red and blue with exactly 5 colors. Now I either have to make the 5 colors and domains, or I have to know to look for the color at 0.2, 0.4, 0.6, etc.

The thinking was that, in many analytical visualization cases, the user wants to control what "buckets" values are visualized in. 
___
*[Elements/src/Analysis/ColorScale.cs, line 78 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MUL6slM8r-ogVGDsYbQ:-MUL6slNA3tVW9deYYz5:b5w7yz4) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L78)):*
> ```Smalltalk
>             else
>             {
>                 throw new ArgumentException("If you provide a list of custom values, it must match your list of colors in its count of items");
> ```

I typically put these checks right up front. It's a small win, but you'd skip even doing the `values==null` check if you don't have the right number of colors/values.

I also like to say what was provided and what is required. Ex: "3 values were provided with 4 colors. The number of values and colors must be the same."
___
*[Elements/src/Analysis/ColorScale.cs, line 88 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULDWRl9HWI4qdXorrI:-MULDWRm81TokS4dwo2F:b-8ilv7r) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L88)):*
> ```Smalltalk
>         /// <param name="t">A number between 0.0 and 1.0</param>
>         /// <returns>A color.</returns>
>         public Color GetColorForValue(double t)
> ```

Looking at my original wording "ForValue" doesn't add anything here. We could just make this `GetColor(double t)` or more specifically,`GetColorAtParameter(double t)` since we're now using `Domain1d`.
___
*[Elements/src/Analysis/ColorScale.cs, line 116 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULCNKK5l6cJjIBixy-:-MULCNKK5l6cJjIBixy0:bghmxac) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L116)):*
> ```Smalltalk
>         private int? GetDomainIndex(double t)
>         {
>             if (t < this.Domains.First().Min || t > this.Domains.Last().Max)
> ```

Do we also have to check that `t` is between 0 and 1?

serenayl (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/532)* status: 6 of 32 files reviewed, 6 unresolved discussions (waiting on @ikeough)

___
*[Elements/src/Analysis/ColorScale.cs, line 27 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULDnnD4x5SejOfm5qK:-MUiCcTI9BYk-wCO7LD1:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L27)):*

Will users need access to the Domains after construction? I think we could get away with this being 
</blockquote></details>

Done.
___
*[Elements/src/Analysis/ColorScale.cs, line 35 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MUL5uh0DOwx63yfcPQp:-MUjIpqdD2KOsRqywjg5:btbygqt) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L35)):*

The Dynamo color range tool used this same methodology and almost everyone is confused by it. That's not to say that we can't get it right, it's just hard. For example, I provide `[Red, Blue, Green] and `[0,1],[0,1],[0,1]` (typical for Dynamo users). What happens? I think I just have duplicate domains, and if I do `GetColorForValue(0.5)` I get purple. Is that correct?

Maybe this constructor is really meant for people who know exactly what they're doing? If so, can we help with a check that the domains are exclusive and a warning if they are not?
</blockquote></details>

Oh, I wrote this for just for serialization and deserialization! But you are right, this would be visible to users too. I edited the description to tell people not to use this constructor.
___
*[Elements/src/Analysis/ColorScale.cs, line 47 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULE_u393A2LeuXrD2A:-MUj4DPG64mtFubaPJd3:b-sw7z6w) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L47)):*

Before we obsolete this, I want to make sure that we can replicate this capability. This was originally designed so that the user could say, I want a gradient between red and blue with exactly 5 colors. Now I either have to make the 5 colors and domains, or I have to know to look for the color at 0.2, 0.4, 0.6, etc.

The thinking was that, in many analytical visualization cases, the user wants to control what "buckets" values are visualized in. 
</blockquote></details>

I understand! I unobsoleted this constructor but am using a different interpolation strategy that I checked would give me an even color dispersal.
___
*[Elements/src/Analysis/ColorScale.cs, line 78 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MUL6slM8r-ogVGDsYbQ:-MUj3mH--y4oteavoX3d:b-52tjoc) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L78)):*

I typically put these checks right up front. It's a small win, but you'd skip even doing the `values==null` check if you don't have the right number of colors/values.

I also like to say what was provided and what is required. Ex: "3 values were provided with 4 colors. The number of values and colors must be the same."
</blockquote></details>

I ended up moving some stuff around so this part is cleaner, and the exception throws at the top.
___
*[Elements/src/Analysis/ColorScale.cs, line 88 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULDWRl9HWI4qdXorrI:-MUjdaqI9OjuzNQJlngQ:bp3n7ak) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L88)):*

Looking at my original wording "ForValue" doesn't add anything here. We could just make this `GetColor(double t)` or more specifically,`GetColorAtParameter(double t)` since we're now using `Domain1d`.
</blockquote></details>

This name as it is makes sense to me than `GetColorAtParameter` because the Domain1d is not necessarily only 0-1, as we are able to provide a list of values like [0, 5, 7, 10], as color anchor points and then `GetColorForValue(7.5)`?

But I do think `GetColor` is nicer. I am adding an obsolete flag to `GetColorForValue` and just forwarding it to `GetColor`.
___
*[Elements/src/Analysis/ColorScale.cs, line 116 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MULCNKK5l6cJjIBixy-:-MUiCLrb3YAmftTGeLUj:bs1ppy8) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L116)):*

Do we also have to check that `t` is between 0 and 1?
</blockquote></details>

This should no longer be a limitation: you should be able to pass in custom values for your colors, which means the user does not need to normalize it themselves. I added checking for between 0 and 1 if this was constructed as a discrete color scale.

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/532)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Analysis/ColorScale.cs, line 35 at r2](https://reviewable.io/reviews/hypar-io/elements/532#-MUL5uh0DOwx63yfcPQp:-MXYs6xT7kIoVCiLAkio:b-btj0v0) ([raw file](https://github.com/hypar-io/elements/blob/606e16dfb1f06d8fd22cda46fa5d54708b30478a/Elements/src/Analysis/ColorScale.cs#L35)):*

Oh, I wrote this for just for serialization and deserialization! But you are right, this would be visible to users too. I edited the description to tell people not to use this constructor.
</blockquote></details>

I believe you can make this method `Internal` and JSON.net will still use it for serialization.

PR #538, opened 2021-03-01
add reference content to content elements, other tweaks

andrewheumann:
BACKGROUND:
- In the interest of building, for example, dynamic furniture layouts, expand the Content Catalog schema to include a list of element instances, representing how elements were arranged at export. Other minor tweaks too.

DESCRIPTION:
- Don't throw an exception for a bad bounding box in UpdateRepresentations(), if one comes up.
- Add a `ReferenceConfiguration` property to the ContentCatalog schema 
- (random) remove some unused exceptions from a catch

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/538)


wynged (changes requested):
> Reviewed 7 of 7 files at r1, 31 of 31 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/538)* status: all files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 1166 at r2](https://reviewable.io/reviews/hypar-io/elements/538#-MUpTUh9Ahuj9u_IyZqk:-MUpTUh9Ahuj9u_IyZql:buai67l) ([raw file](https://github.com/hypar-io/elements/blob/256406da0de63bf7f91097e4416d9c0e82a361a6/Elements/src/Geometry/Polygon.cs#L1166)):*
> ```Smalltalk
>                 return new Polygon(converted);
>             }
>             catch 
> ```

was ArgumentException the wronge exception to catch?  is there a different type we should catch instead?  I feel like generally it's better to have a type unless we're really trying to protect against all possible exceptions.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/538)* status: :shipit: complete! all files reviewed, all discussions resolved

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/538)* status: 4 of 38 files reviewed, all discussions resolved (waiting on @wynged)

___
*[Elements/src/Geometry/Polygon.cs, line 1166 at r2](https://reviewable.io/reviews/hypar-io/elements/538#-MUpTUh9Ahuj9u_IyZqk:-MUtPQDe5GtFp0wkfgUq:bp66jti) ([raw file](https://github.com/hypar-io/elements/blob/256406da0de63bf7f91097e4416d9c0e82a361a6/Elements/src/Geometry/Polygon.cs#L1166)):*

was ArgumentException the wronge exception to catch?  is there a different type we should catch instead?  I feel like generally it's better to have a type unless we're really trying to protect against all possible exceptions.
</blockquote></details>

It was just that we weren't doing anything with the exception at all, so it was resulting in warnings on build. I guess I can put it back and log it?

PR #540, opened 2021-03-02
regex in js alpha bumping strategy

wynged:
BACKGROUND:
- current nuget alphas will appear broken to user after alpha9 because we are not following semver 2 like nuget does.

DESCRIPTION:
- abandon the pre-existing package for alpha incrementing, in favor of using regex and scripts
- it handles either alpha format, with or w/out a `.` but always add the `.` back in.
image

TESTING:
- I discovered a cool tool called Act which lets you test github actions locally. https://github.com/nektos/act 
- other than that, we'll just have to see!
  
FUTURE WORK:
- This does not solve the problem were we have to manually create an alpha release after a full "minor" release

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)


ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, 2 unresolved discussions (waiting on @wynged)

___
*[.github/workflows/build-and-publish-alpha.yml, line 34 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu58Qe92eDZz9k4qwe:-MUu58Qe92eDZz9k4qwf:b-zgxjos) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L34)):*
> ```YAML
>                 const incoming = "${{steps.version.outputs.tag}}"
>                 console.log("Incoming Tag: " + incoming)
>                 const oldNum = incoming.match(/alpha[.]*(\d+)/)[1]
> ```

What is `oldnum`when the incoming version doesn't include `alpha`? This is a failure mode that will be introduced by us forgetting to start a new alpha stream. I think the following should happen:
- We catch that the incoming tag couldn't be matched for alpha.
- We fail the action with a message like: "The incoming version, 0.8.4, does not have an alpha prerelease tag. This might have happened because a new alpha stream was not created after a previous release. Try creating a new annotated release with an semver 2.0 alpha tag, like 0.8.4-alpha.1 and running this action again."
___
*[.github/workflows/build-and-publish-alpha.yml, line 43 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu61rjAqo830Kolrdr:-MUu61rjAqo830Kolrds:be84fvm) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L43)):*
> ```YAML
>         run: echo "NUGET_VERSION=${{steps.bump-alpha.outputs.result}}" >> $GITHUB_ENV
>       - name: Bump if beta.
>         if: ${{contains(steps.version.outputs.tag, 'beta')}}
> ```

We can remove this now correct?

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[.github/workflows/build-and-publish-alpha.yml, line 34 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu58Qe92eDZz9k4qwe:-MUu6TzGEGPwVVlNgkby:b-66rmth) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L34)):*

What is `oldnum`when the incoming version doesn't include `alpha`? This is a failure mode that will be introduced by us forgetting to start a new alpha stream. I think the following should happen:
- We catch that the incoming tag couldn't be matched for alpha.
- We fail the action with a message like: "The incoming version, 0.8.4, does not have an alpha prerelease tag. This might have happened because a new alpha stream was not created after a previous release. Try creating a new annotated release with an semver 2.0 alpha tag, like 0.8.4-alpha.1 and running this action again."
</blockquote></details>

I believe this action won't run at all if the previous tag doesn't contain action.  the `${{contains(steps.version.outputs.tag, 'alpha')}}` handles that
___
*[.github/workflows/build-and-publish-alpha.yml, line 43 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu61rjAqo830Kolrdr:-MUu83ldBMwkQdaSaE_y:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L43)):*

We can remove this now correct?
</blockquote></details>

Done.

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, 2 unresolved discussions (waiting on @ikeough)

___
*[.github/workflows/build-and-publish-alpha.yml, line 34 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu58Qe92eDZz9k4qwe:-MUuC8_U2W0wgZvWCE3c:bd7wcqe) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L34)):*

I believe this action won't run at all if the previous tag doesn't contain action.  the `${{contains(steps.version.outputs.tag, 'alpha')}}` handles that
</blockquote></details>

edits:  this action won't run if the previous tag doesn't contain ALPHA and the line to look for starts with `if: `

ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, 1 unresolved discussion (waiting on @wynged)

___
*[.github/workflows/build-and-publish-alpha.yml, line 34 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu58Qe92eDZz9k4qwe:-MUuC84Q3zc6Npf3sUp1:b-m9qdqb) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L34)):*

edits:  this action won't run if the previous tag doesn't contain ALPHA and the line to look for starts with `if: `
</blockquote></details>

Ah right. But we actually want it to fail so that we can message why it failed, right? If the action just doesn't run, we don't know why. We could remove the `if` condition altogether, and assume that `outputs.tag` has alpha in it, then fail in instructive way.

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, 1 unresolved discussion (waiting on @ikeough)

___
*[.github/workflows/build-and-publish-alpha.yml, line 34 at r2](https://reviewable.io/reviews/hypar-io/elements/540#-MUu58Qe92eDZz9k4qwe:-MUuHmtaEHxP02LJE-qF:b-p8axr7) ([raw file](https://github.com/hypar-io/elements/blob/ea2016d8e8b8385e94b73b86658edf4e93180fdb/.github/workflows/build-and-publish-alpha.yml#L34)):*

Ah right. But we actually want it to fail so that we can message why it failed, right? If the action just doesn't run, we don't know why. We could remove the `if` condition altogether, and assume that `outputs.tag` has alpha in it, then fail in instructive way.
</blockquote></details>

As discussed we can go even further and always add the alpha if it doesn't exist.

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/540)* status: 0 of 188 files reviewed, all discussions resolved

PR #543, opened 2021-03-04
Profiles

ikeough:
BACKGROUND:
Previously the `ProfileServer` was a singleton which maintained a dictionary of all profiles for a catalog. In many cases, we only use a small number of profiles from a catalog. It doesn't make sense to create ALL the available profiles, and there is a non-negligible startup cost for doing so, incurred during the first call to get the instance of the singleton. By refactoring this code using a factory pattern, we can store the table of profile data when we construct a factory instance, then only instantiate profiles when requested using the `GetProfileByName` or `GetProfileByTypes` methods.

DESCRIPTION:
- Refactors `ProfileServer` as a base factory class. 
- Moves the `GetProfileByName` and `GetProfileByType` methods into the inherited factories.
- Implements all former profile servers as profile factories.
- Implements RHS and SHS profile types.

REQUIRED:

SHS profiles:
image

RHS profiles:
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/543)


serenayl (approved):
> Reviewed 18 of 18 files at r1, 47 of 47 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/543)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #545, opened 2021-03-05
Polygon and profile split

andrewheumann:
BACKGROUND:
- The first iteration of `Profile.Split` #527 used a pretty nasty hack — offsetting split geometry into a very thin polygon, and using Profile.Subtract to calculate the split. This proved ineffective in some situations — certain polygons, especially at an angle, would fail to split.

DESCRIPTION:
- Adds `Polygon.Split(Polyline)`
- Adds `Profile.Split(IEnumerable<Profile>, Polyline)`
- Updates `Profile.Split(IEnumerable<Profile>, IEnumerable<Polyline>)` to utilize the new methods. 

TESTING:
- New tests are added for `Polygon.Split`, and the previous `Profile.Split` tests continue to pass. I have validated the edge cases I could think of:
   - split curve totally contained
   - split curve doesn't pass all the way through
   - split curve coincident with polygon edge
   - split curve doesn't intersect polygon at all
   - split curve passes directly through polygon vertex
 
image

  
FUTURE WORK:
- There may be less verbose / more efficient ways to do the same calculations.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/545)


mattc (approved):
> Reviewed 4 of 4 files at r1, 35 of 35 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/545)* status: all files reviewed, 1 unresolved discussion (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 429 at r2](https://reviewable.io/reviews/hypar-io/elements/545#-MVF0iOd4_XZmeCBXKlc:-MVF0iOd4_XZmeCBXKld:b9k4x1d) ([raw file](https://github.com/hypar-io/elements/blob/f13ff4f9ca96e82a8bf009d923844871ee1e4e04/Elements/src/Geometry/Polygon.cs#L429)):*
> ```Smalltalk
>             // construct polygons from half edge graph.
>             // remove edges from edgesPerVertex as they get "consumed" by a polygon,
>             // and stop when you run out of edges.
> ```

Can you add a comment demonstrating that this and the next `while` loop will halt, and describe the worst-case behavior.

PR #547, opened 2021-03-09
failing test and potential solution

wynged:
BACKGROUND:
- I have a situation where a function receives an element that deserializes with some AdditionalProperties.  That element is then added to another model, and this fails, causing a stack overflow exception.

DESCRIPTION:
- Add a test that tests for this case, using the `Identity.AddOverrideIdentity()` method as a way of explicitely adding some additional properties
- Add two strategies that get around the stack overflow error.

TESTING:
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/547)


wynged:
This has not been an issue lately and so I am closing this PR

PR #549, opened 2021-03-11
Fixes + Improvements to profile splitting.

andrewheumann:
# BACKGROUND:
- The last round of Profile / Polygon split work (#545) made an invalid assumption: that a split line would cross all the way through the perimeter of a profile. This doesn't work for "donut" shapes: 
image
- Split also used a single polygon and a single split line at a time. To split with multiple splitters, I was looping over the polygon and splitting one at a time. However, since I was then using clipper to subtract out the voids, internal edges were getting lost:
image
- Lastly, I was cleaning up the "overhanging" segments by simply removing all duplicate segments. This also fails in the above scenario - there is a valid duplicate edge that must be preserved where the split line is inserted. 

Note: it is probably worthwhile to read the PR for #545 as well to understand the overall logic of this operation. Here is a summation (sorry for the length, lol): 

> 1. between the polygon and the polyline, find all intersections between any individual pair of segments. We don't bother checking for polygon or polyline self-intersections, only between the polyline and the polygon.
> 2. we construct unintersected segments — for the original segments that had no intersections, we just add them. For any pair of intersecting segments (one from the polygon, one from the polyline) we sort intersections along the line and construct intermediate lines between each subsequent pair of vertices. These are stored as half-edges, and their ends are added to a vertex array (using a bit of tolerance, so we make sure we have a shared index for any two segment ends that nearly touch). The polygon only adds half-edges going one way, so we don't wind up with the extraneous "outer loop" that bounds the polygon.
> 3. So at this point, we have a clean "noded" network — a set of vertices and a set of half edges. edgesPerVertex stores all of the halfEdges radiating from a given vertex index.
> 4. Each half edge gets used exactly once. We stop the outer loop when there are no more edges left to consume. I don't believe there's any scenario where we'd wind up with a floating half edge that didn't make at least a degenerate loop ((A to B), (B to A)).
> 5. We start with the edges radiating from the first vertex. This is an arbitrary starting point, we could pick any set of edges, since we'll eventually hit them all.
> 6. We remove the first edge from the current set of edges. We use this as the first segment of the new polygon we're creating.
> 7. Then we take the "to" index of the current edge, and find all the edges that radiate from that vertex. We order those edges by their counter-clockwise angle to the current edge, and pick the one forming the smallest angle in a consistent direction — the "leftmost" relative to the current edge — there can be no intermediate edge bisecting this angle, so we know we have found a corner of the polygon.
> 8. We make that selected edge the current edge, and repeat until we have reached the starting point of this polygon, and we have completed a closed polygonal loop. By then, all the edges of this polygon have been removed from edgesPerVertex.
> 9. We do a little bit of cleanup finally on the closed loop we created — if a line didn't intersect all the way through, or we had some degenerate shape, these get detected and discarded.
> 10. Then we proceed to pick another vertex, and another edge, and repeat

# DESCRIPTION:
- Refactors the Polygon.Split method into two main functions: ConstructHalfEdgeGraph and Polygonize.
- Updates ConstructHalfEdgeGraph to consume a list of polygons and a list of splitters, instead of a single polygon and a single splitter
    - this also requires adding a check for polyline-polyline intersections, which before we could safely ignore.
- Updates the logic for removing "degenerate" segments. Before, all duplicates were removed. Now, a procedure finds any vertex with two edges adjacent to it, and removes both adjacent edges. It also checks the edges on either side of that pair of bad edges, to see if THEY are duplicates, and repeats until it is not looking at a bad pair:
image
- Adds overloads and methods for Polygon.Split and Profile.Split (list and single inputs)

# TESTING:
- additional profile split tests have been added, and existing polygon.split tests continue to pass. 
  
# FUTURE WORK:
- Polygonizer / construct edge graph might want to be public methods at some point, they're common and useful utilities for which we have historically relied on NetTopologySuite.

# REQUIRED:

# COMMENTS:
- Because of the refactoring, it looks like this is a bigger change than it is: the procedures for building the half edge graph and the polygonization are largely the same, except where noted above.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/549)


cdriesler (approved):
🏆 model PR imo

Spent more time diving into half-edges than reviewing the code. But it was a real pleasure reading it alongside the graphics.

> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/549#-:-MVdra2mDX8j2XQ_TgZn:baxw79)* status: 1 of 34 files reviewed, 3 unresolved discussions (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 462 at r4](https://reviewable.io/reviews/hypar-io/elements/549#-MVdjfME439Urk6lbP-K:-MVdjfME439Urk6lbP-L:b-9foy2l) ([raw file](https://github.com/hypar-io/elements/blob/8181e431949fabeea726f2fda765dff4a34e4f2c/Elements/src/Geometry/Polygon.cs#L462)):*
> ```Smalltalk
>             ConstructHalfEdgeGraph(new[] { pg }, new[] { pl }, out vertices, out edgesPerVertex);
>         }
>         internal static void ConstructHalfEdgeGraph(IEnumerable<Polygon> pg, IEnumerable<Polyline> pl, out List<Vector3> vertexList, out List<List<(int from, int to)>> edgesPerVertex)
> ```

non-blocking question: why two `out` values instead of returning something? A strategy for legibly returning multiple kinds of things we care about, maybe?
___
*[Elements/src/Geometry/Profile.cs, line 417 at r4](https://reviewable.io/reviews/hypar-io/elements/549#-MVdmxhwCVNeJeznZ7Pv:-MVdmxhwCVNeJeznZ7Pw:b-b3n176) ([raw file](https://github.com/hypar-io/elements/blob/8181e431949fabeea726f2fda765dff4a34e4f2c/Elements/src/Geometry/Profile.cs#L417)):*
> ```Smalltalk
>                     polygons.AddRange(inputProfile.Voids);
>                 }
>                 // construct a half-edge graph from all polygons and spliter polylines.
> ```

🚨🚨 spliter => splitter 🚨🚨
___
*[Elements/src/Geometry/Profile.cs, line 424 at r4](https://reviewable.io/reviews/hypar-io/elements/549#-MVdrpqC6myaT-y-68af:-MVdrpqC6myaT-y-68ag:b-hpi2ap) ([raw file](https://github.com/hypar-io/elements/blob/8181e431949fabeea726f2fda765dff4a34e4f2c/Elements/src/Geometry/Profile.cs#L424)):*
> ```Smalltalk
>                 // so we check if it includes any of the others, and subtract the original profile's voids
>                 // as well. 
>                 for (int i = 0; i < perimSplits.Count; i++)
> ```

This is the only part that felt weird to me. This is because the polygonizer works with the edges, which means we lose void information? But if so, when would a result polygon have its own void that isn't also a void in un-split polygon?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/549)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @cdriesler)

___
*[Elements/src/Geometry/Polygon.cs, line 462 at r4](https://reviewable.io/reviews/hypar-io/elements/549#-MVdjfME439Urk6lbP-K:-MVldcMV7BmpsfHm_kx2:bvgj7ew) ([raw file](https://github.com/hypar-io/elements/blob/8181e431949fabeea726f2fda765dff4a34e4f2c/Elements/src/Geometry/Polygon.cs#L462)):*
<details><summary><i>Previously, cdriesler (Chuck Driesler) wrote…</i></summary><blockquote>

non-blocking question: why two `out` values instead of returning something? A strategy for legibly returning multiple kinds of things we care about, maybe?
</blockquote></details>

real answer: laziness — this is what vs code's autorefactor did when I extracted the method. 
Post-rationalized answer: the vertices / edges are equally important — it seemed weird to have one be the "return" value and one to be the "out" value. But no strong feeling either way. 
The truly correct thing to do here would have been to create a "halfedge graph" class containing both the vertices and the edges, and return that. So, thanks for the comment — inspired me to do exactly that and refactor into a separate class. 
___
*[Elements/src/Geometry/Profile.cs, line 424 at r4](https://reviewable.io/reviews/hypar-io/elements/549#-MVdrpqC6myaT-y-68af:-MVlnQIA2t7IEooGMFoH:bjpakdg) ([raw file](https://github.com/hypar-io/elements/blob/8181e431949fabeea726f2fda765dff4a34e4f2c/Elements/src/Geometry/Profile.cs#L424)):*
<details><summary><i>Previously, cdriesler (Chuck Driesler) wrote…</i></summary><blockquote>

This is the only part that felt weird to me. This is because the polygonizer works with the edges, which means we lose void information? But if so, when would a result polygon have its own void that isn't also a void in un-split polygon?
</blockquote></details>

The trouble is the polygonizer doesn't differentiate (really) between voids and not-voids — you could theoretically wind up filling in a void if you didn't make the effort to remove voids after the fact. I *think* we're mostly protected against this because we only draw a single half-edge for polygons, but I wasn't sure enough not to do this double-check.

PR #551, opened 2021-03-12
Add constraints to recursion.

ikeough:
BACKGROUND:
When a user calls `Model.AddElement(...)` currently, we inspect the element's properties recursively to find other elements that should be added to the model. This enable us to identify, for example, a material used by an element that should be added to the model. This is convenient for the user as they can use elements that reference other elements, and contribute all of those to the model with only one call to `Model.AddElement`. 

The downside to this strategy is that we need to look at _every_ property which is a reference type, recursively inspecting its properties. This can be an extremely expensive operation, and it's not clear to the developer currently when they might inadvertently cause this situation to arise. As an example of this, we recently made meshes serializable on `MeshElement`. Doing so caused `Model.AddElement` to inspect the `Mesh` property on the incoming `MeshElement`, then to inspect each of its sub-properties (`Vertices`, `Triangles`, etc.) even though nothing in this reference chain would ultimately need to be added to the model.

This PR does not remove recursion. It only makes the recursion more restricted and tested.

DESCRIPTION:
- This PR defines 3 cases in which a property should be considered for inspection and limits the properties returned to only those 3 cases.

TESTING:
- Three tests have been added which identify that test the cases that we would like to support. As additional cases are identified that we want to support, test cases should be added, and documentation updated to make it clear what is supported.
- As a test, this branch was applied to the Location function where we serialize the topographic meshes. With no changes to the function logic the result was a 75% reduction in time to add elements to the model and serialize:
```json
// Adding all elements to the model and serializing with deep recursion.
10803.1922ms

// Adding all elements to the model and serializing with constrained recursion.
2747.0406ms
```
  
FUTURE WORK:
- Concerns have been raised that this work may inadvertently expose cases where we assumed sub-elements would be added, and they no longer will be. 

REQUIRED:
- [ ] All changes are up to date in `CHANGELOG.md`. **n/a** (no changes to public APIs)

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)


ikeough:
Inviting both @wynged and @andrewheumann to review and comment as they both voiced opinions and alternatives for this.

andrewheumann (changes requested):
IFCTests.Floor is also failing.

> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551#-:-MVm0mvB48SWU4RRSJaA:b2g7efj)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @ikeough and @wynged)

___
*[Elements/src/Model.cs, line 252 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVlq54992c8_RjhnyFh:-MVlq54992c8_RjhnyFi:b-dffpxc) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/src/Model.cs#L252)):*
> ```Smalltalk
>             var props = t.GetProperties(BindingFlags.Public | BindingFlags.Instance)
>                             .Where(p => typeof(Element).IsAssignableFrom(p.PropertyType)
>                             || typeof(IList<Element>).IsAssignableFrom(p.PropertyType)
> ```

These list / dictionary checks will not work in all cases. 
If you try this check on a property of type `List<Floor>` or `Dictionary<string,Floor>`, it will return false and skip over those properties.
For a list, I'd recommend bumping up to IEnumerable, from which IList inherits. `typeof(List<Floor>)` is assignable from `IEnumerable<Element>` but not from `IList<Element>`. 
You could also be more explicit and say
`(typeof(IList).IsAssignableFrom(p.PropertyType) && typeof(Element).IsAssignableFrom(type.GetGenericArguments().First()))`
For a dictionary, you need to be more explicit, with something like this:
`(typeof(IDictionary).IsAssignableFrom(p.PropertyType) && typeof(Element).IsAssignableFrom(type.GetGenericArguments().Last()))`
___
*[Elements/test/ModelTests.cs, line 295 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVm0SOnEIluE5RdfJ1F:-MVm0SOnEIluE5RdfJ1G:bmzoa7u) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L295)):*
> ```Smalltalk
> 
>         [Fact]
>         public void SubDictionaryOfElementIsAddedToModel()
> ```

This test fails, no?

andrewheumann (commented):
> Reviewed 1 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: all files reviewed, 2 unresolved discussions (waiting on @ikeough and @wynged)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[Elements/src/Model.cs, line 252 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVlq54992c8_RjhnyFh:-MVrZwDd-AxUKyq0ySif:by1dduf) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/src/Model.cs#L252)):*

These list / dictionary checks will not work in all cases. 
If you try this check on a property of type `List<Floor>` or `Dictionary<string,Floor>`, it will return false and skip over those properties.
For a list, I'd recommend bumping up to IEnumerable, from which IList inherits. `typeof(List<Floor>)` is assignable from `IEnumerable<Element>` but not from `IList<Element>`. 
You could also be more explicit and say
`(typeof(IList).IsAssignableFrom(p.PropertyType) && typeof(Element).IsAssignableFrom(type.GetGenericArguments().First()))`
For a dictionary, you need to be more explicit, with something like this:
`(typeof(IDictionary).IsAssignableFrom(p.PropertyType) && typeof(Element).IsAssignableFrom(type.GetGenericArguments().Last()))`
</blockquote></details>

This is a good catch. It was even trickier than the proposed solution! See the Model.IsValidForRecursiveAddition method which now handles all of our cases.
___
*[Elements/test/ModelTests.cs, line 295 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVm0SOnEIluE5RdfJ1F:-MVsLAFFB5mOh8B5y7Pg:bt2vr1b) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L295)):*

This test fails, no? 
</blockquote></details>

Fixed.

andrewheumann (changes requested):
> Reviewed 29 of 29 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann, @ikeough, and @wynged)

___
*[Elements/src/Model.cs, line 300 at r4](https://reviewable.io/reviews/hypar-io/elements/551#-MVugZt2ACaBsLLPgl7n:-MVugZt2ACaBsLLPgl7o:b-5wkmxh) ([raw file](https://github.com/hypar-io/elements/blob/7bfec56363944f386eaa2cd7f49ae8d958a9c8ba/Elements/src/Model.cs#L300)):*
> ```Smalltalk
>             if (t.IsGenericType)
>             {
>                 if (typeof(IList<>).MakeGenericType(t.GetGenericArguments()[0]).IsAssignableFrom(t))
> ```

unless I'm missing something, this no longer does what you set out to do — which is verify that `t` is a collection of Elements or a dictionary containing Elements. Is this an intentional change in strategy (let's just recursively traverse all lists + dictionaries, regardless of their contents)?   It seems like in the Array case you are still verifying that `Element` is part of the equation.

andrewheumann (changes requested):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann, @ikeough, and @wynged)

___
*[Elements/src/Model.cs, line 289 at r4](https://reviewable.io/reviews/hypar-io/elements/551#-MVuiSon7SmXgkx8Ei3B:-MVuiSon7SmXgkx8Ei3C:b-w2xrsh) ([raw file](https://github.com/hypar-io/elements/blob/7bfec56363944f386eaa2cd7f49ae8d958a9c8ba/Elements/src/Model.cs#L289)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// All types that are valid for introspection including
> ```

unfinished comment

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[Elements/src/Model.cs, line 252 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVlq54992c8_RjhnyFh:-MVvX-ub4rGGzdX2VP4I:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/src/Model.cs#L252)):*

This is a good catch. It was even trickier than the proposed solution! See the Model.IsValidForRecursiveAddition method which now handles all of our cases.
</blockquote></details>

Done.
___
*[Elements/src/Model.cs, line 289 at r4](https://reviewable.io/reviews/hypar-io/elements/551#-MVuiSon7SmXgkx8Ei3B:-MVvVfiX9RjQg0caZmE_:bt2vr1b) ([raw file](https://github.com/hypar-io/elements/blob/7bfec56363944f386eaa2cd7f49ae8d958a9c8ba/Elements/src/Model.cs#L289)):*

unfinished comment
</blockquote></details>

Fixed.
___
*[Elements/src/Model.cs, line 300 at r4](https://reviewable.io/reviews/hypar-io/elements/551#-MVugZt2ACaBsLLPgl7n:-MVvVhUODVfCzMh0V236:book6sh) ([raw file](https://github.com/hypar-io/elements/blob/7bfec56363944f386eaa2cd7f49ae8d958a9c8ba/Elements/src/Model.cs#L300)):*

unless I'm missing something, this no longer does what you set out to do — which is verify that `t` is a collection of Elements or a dictionary containing Elements. Is this an intentional change in strategy (let's just recursively traverse all lists + dictionaries, regardless of their contents)?   It seems like in the Array case you are still verifying that `Element` is part of the equation.
</blockquote></details>

You weren't missing anything! I failed to add the contrary condition to the test. Doing so unveiled the failure to test for Elements (and SolidOperations). This is resolved with another utility method.

andrewheumann (approved):
> Reviewed 2 of 2 files at r5, 26 of 26 files at r6.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Model.cs, line 309 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwCykp8xI0-R2IJKP7:-MVwCykp8xI0-R2IJKP8:b-sc36h5) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L309)):*
> ```Smalltalk
> 
>                     return true;
>                 }
> ```

Another generic type to add here is the Nullable type.  I needed to add this.  the count check seems necessary because the dictionary types will have the wrong number of generic types to succeed the the `MakeGenericType` call.
```
                else if (genT.Count() == 1 && typeof(System.Nullable<>).MakeGenericType(genT[0]).IsAssignableFrom(t))
                {
                    if (typeof(Element).IsAssignableFrom(genT[0]))
                    {
                        return true;
                    }
                    else
                    {
                        return false;
                    }
                }
```
___
*[Elements/test/ModelTests.cs, line 312 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVqAut-2twMsca7sP8t:-MVqAut-2twMsca7sP8u:b-pqatce) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L312)):*
> ```Smalltalk
>             var line = new Line(Vector3.Origin, new Vector3(5, 5, 5));
>             var ue = new TestUserElement(line, new Profile(Polygon.L(1, 2, 0.5)));
>             ue.UpdateRepresentations();
> ```

Can we remove this UpdateRepresentations call?  AddElement should handle that and we want the test to run code similar to what users will be writing.

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Model.cs, line 302 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwFRaM7JhRULkehy3X:-MVwFRaN2jstRIPAeSV1:b-hcwnsu) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L302)):*
> ```Smalltalk
>                 var genT = t.GetGenericArguments();
>                 if (typeof(IList<>).MakeGenericType(genT[0]).IsAssignableFrom(t))
>                 {
> ```

I think we should also check for IEnumerable here.  Serialization failed for one of my functions because I had added a utility getter to an element.  I don't even need this prop serialized, but it caused adding to a model to fail.  I think we need this.
` || typeof(IEnumerable<>).MakeGenericType(genT[0]).IsAssignableFrom(t)`

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Model.cs, line 296 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwTLSD1Mn_SSh3xhtb:-MVwTLSD1Mn_SSh3xhtc:b-b5bavy) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L296)):*
> ```Smalltalk
>         /// <param name="t">The type to check.</param>
>         /// <returns>Return true if a type is valid for introspection, otherwise false.</returns>
>         internal static bool IsValidForRecursiveAddition(Type t)
> ```

The two places where I've made suggestions below were errors with the code that checked if a type is valid, which resulted in a failure to add an element to a model.  Could we wrap the code inside that `IsValidForRecursiveAddition` in a try/catch to throw a more clear error than this?  I checked running with a try/catch, and the time impact in the success case seems minimal.
```
An exception of type 'System.ArgumentException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'GenericArguments[0], 'Elements.Piping.IPipeNetworkComponent', on 'System.Nullable`1[T]' violates the constraint of type 'T'.'
```

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[Elements/src/Model.cs, line 296 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwTLSD1Mn_SSh3xhtb:-MVxJqRB6rL3eSgJvVhT:b170ezs) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L296)):*

The two places where I've made suggestions below were errors with the code that checked if a type is valid, which resulted in a failure to add an element to a model.  Could we wrap the code inside that `IsValidForRecursiveAddition` in a try/catch to throw a more clear error than this?  I checked running with a try/catch, and the time impact in the success case seems minimal.
```
An exception of type 'System.ArgumentException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'GenericArguments[0], 'Elements.Piping.IPipeNetworkComponent', on 'System.Nullable`1[T]' violates the constraint of type 'T'.'
```
</blockquote></details>

What's the expected behavior when you catch an exception? Silently handle it and continue? If so it'll be challenging to know why a property wasn't searched.
___
*[Elements/src/Model.cs, line 302 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwFRaM7JhRULkehy3X:-MVwniheE2YKVpoco1oC:b-xxlew0) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L302)):*

I think we should also check for IEnumerable here.  Serialization failed for one of my functions because I had added a utility getter to an element.  I don't even need this prop serialized, but it caused adding to a model to fail.  I think we need this.
` || typeof(IEnumerable<>).MakeGenericType(genT[0]).IsAssignableFrom(t)`
</blockquote></details>

This seems to be handled now without adding the condition above. Test it in your code and see.
___
*[Elements/src/Model.cs, line 309 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwCykp8xI0-R2IJKP7:-MVxKu0dEMsmHDKL2wvs:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L309)):*

Another generic type to add here is the Nullable type.  I needed to add this.  the count check seems necessary because the dictionary types will have the wrong number of generic types to succeed the the `MakeGenericType` call.
```
                else if (genT.Count() == 1 && typeof(System.Nullable<>).MakeGenericType(genT[0]).IsAssignableFrom(t))
                {
                    if (typeof(Element).IsAssignableFrom(genT[0]))
                    {
                        return true;
                    }
                    else
                    {
                        return false;
                    }
                }
```
</blockquote></details>

Done.
___
*[Elements/test/ModelTests.cs, line 312 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVqAut-2twMsca7sP8t:-MVx5QuO3g7hWA276_iE:bierre4) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L312)):*

Can we remove this UpdateRepresentations call?  AddElement should handle that and we want the test to run code similar to what users will be writing.  
</blockquote></details>

Done. Not only did I remove it here, I moved it out of `ToJson` and into `AddElement`.

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann, @ikeough, and @wynged)

___
*[Elements/src/Model.cs, line 296 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwTLSD1Mn_SSh3xhtb:-MVxo4WJ6fHjqIRF4HJs:bo103my) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L296)):*

What's the expected behavior when you catch an exception? Silently handle it and continue? If so it'll be challenging to know why a property wasn't searched.
</blockquote></details>

I would say still throw an error, but one that is more clear.  The error a user will get if they have an invalid property will be totally opaque, the error above.  If we could throw an error that was more like "The property {name} caused the element to fail adding to the model because the type {type} is not supported"
___
*[Elements/test/ModelTests.cs, line 312 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVqAut-2twMsca7sP8t:-MVxodH2DyLe80NiXSmJ:bl16pfm) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L312)):*

Done. Not only did I remove it here, I moved it out of `ToJson` and into `AddElement`.
</blockquote></details>

I think it should also be in ToJson, because elements are mutable they can change even after being added to a model, and at the moment of serialization we still want to be sure that we have the most up to date geometry right?

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Model.cs, line 302 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwFRaM7JhRULkehy3X:-MW-d2Ar2XThuAb9lh-z:bvzt5nu) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L302)):*

This seems to be handled now without adding the condition above. Test it in your code and see.
</blockquote></details>

done. seems resolved, and more generically, we no longer end up in the state where an invalid type breaks this validation step.
___
*[Elements/src/Model.cs, line 309 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwCykp8xI0-R2IJKP7:-MW-cygX9cfBB-G_oKb4:bjjz7ej) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L309)):*

Done.
</blockquote></details>

yup, looks like that change to logic has fixed it.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[Elements/src/Model.cs, line 296 at r6](https://reviewable.io/reviews/hypar-io/elements/551#-MVwTLSD1Mn_SSh3xhtb:-MW1PpD68LLPrkGq9hmL:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/f5e71d4e68b87960888fe2f1b75db594fe5f03bd/Elements/src/Model.cs#L296)):*

I would say still throw an error, but one that is more clear.  The error a user will get if they have an invalid property will be totally opaque, the error above.  If we could throw an error that was more like "The property {name} caused the element to fail adding to the model because the type {type} is not supported"
</blockquote></details>

Done.
___
*[Elements/test/ModelTests.cs, line 312 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVqAut-2twMsca7sP8t:-MW1OGDV8OIevhH3PpOC:bdv37bv) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L312)):*

I think it should also be in ToJson, because elements are mutable they can change even after being added to a model, and at the moment of serialization we still want to be sure that we have the most up to date geometry right?
</blockquote></details>

It _is_ in `ToJson` by virtue of being in `AddElement` (which is called in `ToJson`).

ikeough (commented):
> Reviewed 1 of 4 files at r1, 1 of 29 files at r4, 5 of 6 files at r7, 25 of 26 files at r8.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/551)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Model.cs, line 312 at r8](https://reviewable.io/reviews/hypar-io/elements/551#-MW1hM8G3LfYdpcF8_5D:-MW1hM8G3LfYdpcF8_5E:b572pv6) ([raw file](https://github.com/hypar-io/elements/blob/c5582a02f4bdc50979e406cd51104240d5ffcbd2/Elements/src/Model.cs#L312)):*
> ```Smalltalk
>                 {
>                     if (typeof(IList<>).MakeGenericType(genT[0]).IsAssignableFrom(t))
>                     //|| typeof(IEnumerable<>).MakeGenericType(genT[0]).IsAssignableFrom(t))
> ```

you'll wanna remove this commented code no that we don't need it.
___
*[Elements/test/ModelTests.cs, line 312 at r2](https://reviewable.io/reviews/hypar-io/elements/551#-MVqAut-2twMsca7sP8t:-MW1gySC6y_6Ew0cC9eX:bjdw27k) ([raw file](https://github.com/hypar-io/elements/blob/110cebcec18e8b9b3ac90e0757f8125acc9de82b/Elements/test/ModelTests.cs#L312)):*

It _is_ in `ToJson` by virtue of being in `AddElement` (which is called in `ToJson`).
</blockquote></details>

ahhhh... because we add all elements to a new model, forgot about that. it's a little bit sneaky that.

PR #560, opened 2021-03-18
Add an example for traversal.

ikeough:
BACKGROUND:
- I wanted to test @serenayl's `CellComplex`, so I wrote a traversal example. 

DESCRIPTION:
- This tests three dimensional traversal of vertices and edges in a cell complex.

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/560)


serenayl (changes requested):
> Reviewed 14 of 44 files at r2, 26 of 26 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/560)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/test/CellComplexTests.cs, line 285 at r4](https://reviewable.io/reviews/hypar-io/elements/560#-MW9jzxi1DTycJfUD1Fz:-MW9jzxjAAEbk_sRlddv:b-3kbqyv) ([raw file](https://github.com/hypar-io/elements/blob/199fbe298342673d491df4fe17c1b811a2d15ea1/Elements/test/CellComplexTests.cs#L285)):*
> ```Smalltalk
>                 Elements.Spatial.CellComplex.Vertex newStart = null;
>                 Elements.Spatial.CellComplex.Edge chosenEdge = null;
>                 foreach (var e in neighbors)
> ```

I'm curious about whether a version of this would work after you've determined your first segment, to just keep using `edge.GetClosestNeighbor` (which will return closest edge to the point that shares a vertex). Those sibling-neighbor functions for `Edge`, `Face`, and `Cell` automatically discard neighbors that aren't closer than the current child and return null if none are found

PR #563, opened 2021-03-24
Make Import SolidOperation public

andrewheumann:
BACKGROUND:
- For the purposes of building Rhino importers and other functionality that might want to construct a solid that isn't an extrude, sweep, or lamina, make `Import` public

DESCRIPTION:
- Makes the `Elements.Geometry.Solids.Import` class public

TESTING:
- Added a test, involving constructing an import from scratch.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/563)


ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/563)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Solids/Import.cs, line 6 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa4sO214NLnwA1fbkJ:-MWa4sO214NLnwA1fbkK:b-kq2nlf) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/src/Geometry/Solids/Import.cs#L6)):*
> ```Smalltalk
>     /// Create a custom SolidOperation from imported geometry.
>     /// </summary>
>     public class Import : SolidOperation
> ```

If we're going to make this public, we have an opportunity to make the name something more general. This was originally implemented for support of `IfcManifoldSolidBrep` imports from IFC. We could name it something more general like `ConstructedSolid`, because it's not always going to be an import.
___
*[Elements/test/SolidTests.cs, line 216 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa6RvrDYMMBmvlw8uM:-MWa6RvrDYMMBmvlw8uN:b-xxobro) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/test/SolidTests.cs#L216)):*
> ```Smalltalk
>             solid.AddFace(new Polygon(new[] { C, D, H, G }));
>             solid.AddFace(new Polygon(new[] { D, A, E, H }));
>             var import = new Import(solid, false);
> ```

What happens when we add a solid with no faces? I don't know, but I think we'll get empty geometry which will cause glTF to barf. Can we test that here and protect?
___
*[Elements/test/SolidTests.cs, line 221 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa6CG1C_otF6Lk1o1f:-MWa6CG1C_otF6Lk1o1g:br7lzf9) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/test/SolidTests.cs#L221)):*
> ```Smalltalk
>             var model = new Model();
>             model.AddElement(userElement);
>             var json = model.ToJson();
> ```

I don't believe these calls `ToJson` and `ToGlTF` are necessary if this is a `ModelTest`. Json and glTF will be created in the tear down method and the test will fail if that step fails.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/563)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/test/SolidTests.cs, line 216 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa6RvrDYMMBmvlw8uM:-MWaGqVTEr4K8k4Kaif4:boi8ner) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/test/SolidTests.cs#L216)):*

What happens when we add a solid with no faces? I don't know, but I think we'll get empty geometry which will cause glTF to barf. Can we test that here and protect?
</blockquote></details>

It works fine, actually. Updated test to include this case. Should the behavior be something other than swallow silently and proceeed?
___
*[Elements/test/SolidTests.cs, line 221 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa6CG1C_otF6Lk1o1f:-MWaGxIv9B5wfaKuHf_V:bpxurgs) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/test/SolidTests.cs#L221)):*

I don't believe these calls `ToJson` and `ToGlTF` are necessary if this is a `ModelTest`. Json and glTF will be created in the tear down method and the test will fail if that step fails.
</blockquote></details>

SolidTests was not a ModelTest, which is why I did this manually. Have made it a ModelTest so I can do as suggested. 
___
*[Elements/src/Geometry/Solids/Import.cs, line 6 at r2](https://reviewable.io/reviews/hypar-io/elements/563#-MWa4sO214NLnwA1fbkJ:-MWaFhOn1k8EbwJwI1EX:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/bc48a6633b352d6f5078a45814903abc6ec10008/Elements/src/Geometry/Solids/Import.cs#L6)):*

If we're going to make this public, we have an opportunity to make the name something more general. This was originally implemented for support of `IfcManifoldSolidBrep` imports from IFC. We could name it something more general like `ConstructedSolid`, because it's not always going to be an import.
</blockquote></details>

Done.

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/563)* status: :shipit: complete! 1 of 1 approvals obtained

PR #565, opened 2021-03-26
Polygon.Contains expanded 3D support

serenayl:
BACKGROUND:
- I found some cases where polygons reported a false containment with a vertex that was above the polygon

DESCRIPTION:
- Adds an internal middleware function that checks for vertices on the XY plane, and runs the calculation on a transformed polygon and point if it is not an XY plane polygon

TESTING:
- See the new test for the previously failing case
  
FUTURE WORK:
- Make more cohesive our strategy for 2D and 3D geometries.

REQUIRED:

COMMENTS:
- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/565)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - `Polyline.OffsetOnSide`
 ### Changed
 - `Solids.Import` is now public.
+- `Polygon.Contains` was modified to better handle polygons that are not on the XY plane.
 
 ### Fixed
 ## 0.8.4
@@ -30,7 +31,7 @@
 - `Profile.Split(IEnumerable<Profile>, IEnumerable<Polyline> p)`
 - `Elements.Spatial.HalfEdgeGraph2d`
   - `.Construct()`
-  - `.Polygonize()` 
+  - `.Polygonize()`
 - Release helper github action
 
 ### Changed
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -75,7 +75,31 @@ namespace Elements.Geometry
         /// <returns>Returns true if the supplied Vector3 is within this polygon.</returns>
         public bool Contains(Vector3 vector, out Containment containment)
         {
-            return Contains(Segments(), vector, out containment);
+            return Contains3D(Segments(), vector, out containment);
+        }
+
+        // Projects non-flat containment request into XY plane and returns the answer for this projection
+        internal static bool Contains3D(IEnumerable<Line> segments, Vector3 location, out Containment containment)
+        {
+            var is3D = false;
+            var vertices = segments.Select(segment =>
+            {
+                if (segment.Start.Z != 0)
+                {
+                    is3D = true;
+                }
+                return segment.Start;
+            }).ToList();
+            if (!is3D)
+            {
+                return Contains(segments, location, out containment);
+            }
+            var transformTo3D = vertices.ToTransform();
+            var transformToGround = new Transform(transformTo3D);
+            transformToGround.Invert();
+            var groundSegments = segments.Select(segment =>  (Line)segment.Transformed(transformToGround)).ToList();
+            var groundLocation = transformToGround.OfPoint(location);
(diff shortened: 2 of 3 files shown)


andrewheumann (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/565)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Geometry/Polygon.cs, line 89 at r1](https://reviewable.io/reviews/hypar-io/elements/565#-MX22PYf2xPxpHcQtmcK:-MX22PYf2xPxpHcQtmcL:bmowhvg) ([raw file](https://github.com/hypar-io/elements/blob/446ad7d696910e4ede7464eddd1bdc303d3b1995/Elements/src/Geometry/Polygon.cs#L89)):*
> ```Smalltalk
>                 if (segment.Start.Z != 0)
>                 {
>                     is3D = true;
> ```

It's a little strange to have side effects in a `.Select()`, because of deferred execution of linq statements. You avoid that here by forcing execution with `ToList()` but I think for clarity / predictability I would write this as: 
```
var vertices = segments.Select(segment => segment.Start);
var is3D = vertices.Any(vertex => vertex.Z != 0);
```
even though it's possibly a tiny bit more looping. 
___
*[Elements/src/Geometry/Polygon.cs, line 100 at r1](https://reviewable.io/reviews/hypar-io/elements/565#-MX23Qe42QY52Udieacn:-MX23Qe556bylp71seI-:blvczbz) ([raw file](https://github.com/hypar-io/elements/blob/446ad7d696910e4ede7464eddd1bdc303d3b1995/Elements/src/Geometry/Polygon.cs#L100)):*
> ```Smalltalk
>             var transformToGround = new Transform(transformTo3D);
>             transformToGround.Invert();
>             var groundSegments = segments.Select(segment =>  (Line)segment.Transformed(transformToGround)).ToList();
> ```

You can avoid the cast by calling `segment.TransformedLine()` instead of `(Line)segment.Transformed()`, and I don't think this last .ToList() is necessary since `Contains` takes `IEnumerable`.

serenayl (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/565)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 89 at r1](https://reviewable.io/reviews/hypar-io/elements/565#-MX22PYf2xPxpHcQtmcK:-MX2ey4b7IMFVgt0NOB9:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/446ad7d696910e4ede7464eddd1bdc303d3b1995/Elements/src/Geometry/Polygon.cs#L89)):*

It's a little strange to have side effects in a `.Select()`, because of deferred execution of linq statements. You avoid that here by forcing execution with `ToList()` but I think for clarity / predictability I would write this as: 
```
var vertices = segments.Select(segment => segment.Start);
var is3D = vertices.Any(vertex => vertex.Z != 0);
```
even though it's possibly a tiny bit more looping. 
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Polygon.cs, line 100 at r1](https://reviewable.io/reviews/hypar-io/elements/565#-MX23Qe42QY52Udieacn:-MX2exr-5QVa4lWljmH8:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/446ad7d696910e4ede7464eddd1bdc303d3b1995/Elements/src/Geometry/Polygon.cs#L100)):*

You can avoid the cast by calling `segment.TransformedLine()` instead of `(Line)segment.Transformed()`, and I don't think this last .ToList() is necessary since `Contains` takes `IEnumerable`. 
</blockquote></details>

Done.

andrewheumann (approved):
> Reviewed 1 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/565)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #567, opened 2021-04-03
Topography Cut/Fill Features


Gytaco:
**Is your feature request related to a problem? Please describe.**
I would like a way to procedurally setup Cut/Fill features for my sites and load in more detailed sites from other sources. Currently, this is not possible. We are building site selection tools that focus on the potential costs of site works and optimising for this issue. There are no tools that allow you to quickly iterate through topography cut/fill options.

As an output we would require the following.
- An ability to get the Final edited Topography
- Get the Cut/Fill for each object as well as the total for the edited mesh

**Describe the solution you'd like**
An ability to create Custom Topography from dxf or solids that work like the current topography from Mapbox.
An ability to use solids or surfaces to Cut (Boolean) into Topography and get the Cut Volume.
An ability to use solids or surfaces to Fill creates a new solid between the difference of the current topography and the solid/surfaces. An additional ability to set a gradient of the Fill as well.

This is basically a creation of Revit Site tools with a grading feature to specify the gradient of the site from the surface.

**Describe alternatives you've considered**
Revit currently has most functions but it is not exposed to the API, we have to manually update the pad footings then extract the data then perform another transaction so it is very slow.
We have built similar functions with Grasshopper but we have issues with booleans so we had to use mesh conversions to get the work done.

**Additional context**
Cut/Fill example using a Surface. (Expected Functionality)
![image](https://user-images.githubusercontent.com/5660565/113492273-adf27080-9519-11eb-80ca-6d73ce17f4a3.png)

Ability to set a Gradient for the Topography
![image](https://user-images.githubusercontent.com/5660565/113492307-dc704b80-9519-11eb-91b1-1c5140a0d7a3.png)

ikeough:
Thanks @Gytaco. This is great. I'd propose the following new APIs:

```csharp
// Cut a Topography with a polygon to a specified depth and return the volume of cut in meters^3
double Topography.Cut(Polygon perimeter, double depth)

// Fill a Topography with a polygon extruded to a specified height, with a specified batter angle, and return the volume of fill in meters^3
double Topography.Fill(Polygon perimeter, double topElevation, double batterAngle)
```

Does this sound about right?

Gytaco:
I think the only issue you've highlighted is what if a Polygon is creating both a cut and fill? I wonder if we just have single command and let the API take of all of it. I know it will be a bit more work but users have to guess which polygons cut and which sections fill could be problematic. Refer to the image above to see what I mean.

[...]

PR #571, opened 2021-04-13
Cut and fill.

ikeough:
BACKGROUND:
This PR addresses #567.

DESCRIPTION:
- This PR adds one new API: `Topography.CutAndFill(...)` which can be used to cut and fill and topography to a specified elevation within a specified polygon.

TESTING:
- Experiment with various cut and fill elevations and shapes. See the `TopographyTests.Cut` and `TopographyTests.Fill` tests for examples.
- [This workflow](https://dev.hypar.io/workflows/9ddfa07b-e7a1-4fb4-87e8-3069fdb98d0c) uses the cut and fill APIs.
  
FUTURE WORK:
- CSG operations still involve a lot of conversion from Solid -> Mesh -> CSG -> Mesh -> Buffers. This is especially noticeable when working with topography due to the size of the mesh.

REQUIRED:

An example of fill with a 45d batter angle.
image

An example of a cut
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)


wynged (changes requested):
There's a failing test for `AddInvalidTriangle`

> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571#-:-MYKxZW0D4bNMExxHRnL:b-sey7k9)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Topography.cs, line 501 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwGuG1FPlYSo2NxdD:-MYKwGuG1FPlYSo2NxdE:be3opke) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Topography.cs#L501)):*
> ```Smalltalk
>         /// <param name="cutVolume">A mesh representing the cut volume in the topography.</param>
>         /// <returns>The cut and fill volumes.</returns>
>         public (double Cut, double Fill) CutAndFill(Polygon perimeter, double elevation, out Mesh cutVolume, out Mesh fillVolume, double batterAngle = 45.0)
> ```

the out var and return names seem backwards to me?  the return is called Cut, but it is the volume, and the out var is called cutVolume, but is a mesh.
___
*[Elements/src/Geometry/BBox3.cs, line 28 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwYpOC9nyLV4lprH8:-MYKwYpOC9nyLV4lprH9:b-y5kyv6) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/BBox3.cs#L28)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="bounds">The existing bounding box.</param>
>         public BBox3(BBox3 bounds)
> ```

Is this used anywhere?  would a Clone() method be more discoverable for this purpose?
___
*[Elements/src/Geometry/CsgExtensions.cs, line 270 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwpBj8L1W8ejJjOLi:-MYKwpBj8L1W8ejJjOLj:b-ymr7t8) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/CsgExtensions.cs#L270)):*
> ```Smalltalk
>                 polygons.Add(polygon);
>             }
>             var solid = new Csg.Solid();
> ```

does this line do anything?
___
*[Elements/src/Geometry/Mesh.cs, line 250 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKpDQx72fOqYf3MSMs:-MYKpDQx72fOqYf3MSMt:b-6i0iud) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L250)):*
> ```Smalltalk
>             {
>                 // throw new ArgumentException($"Not a valid Triangle.  Duplicate vertex at {duplicate}.");
> ```

remove this if we want to allow triangles with duplicate vertices, but this also causes a test to fail.
___
*[Elements/src/Geometry/Mesh.cs, line 272 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKpulb8JLWJB4w9qEQ:-MYKpulb8JLWJB4w9qER:bncrvq8) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L272)):*
> ```Smalltalk
>                                 Vector3 normal = default(Vector3),
>                                 Color color = default(Color),
>                                 bool merge = false,
> ```

any risks to making having merge default to `true` ?  it's the behavior we would rather have in general right?
___
*[Elements/src/Geometry/Mesh.cs, line 332 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKtZjhEUyb4u1oSBsN:-MYKtZjhEUyb4u1oSBsO:b-lsj7fx) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L332)):*
> ```Smalltalk
>         }
> 
>         private void ResetIndices()
> ```

I don't think this method is used anywhere, and it seems like it might break triangles if it was?

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @wynged)

___
*[Elements/src/Topography.cs, line 501 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwGuG1FPlYSo2NxdD:-MYMcvC59GfJ8t18O7xo:b-pf77jx) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Topography.cs#L501)):*

the out var and return names seem backwards to me?  the return is called Cut, but it is the volume, and the out var is called cutVolume, but is a mesh.
</blockquote></details>

Good point. I've reversed them.
___
*[Elements/src/Geometry/BBox3.cs, line 28 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwYpOC9nyLV4lprH8:-MYMcyj31TGUB1S2BYzF:b-upu603) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/BBox3.cs#L28)):*

Is this used anywhere?  would a Clone() method be more discoverable for this purpose?
</blockquote></details>

This was vestigial from another change that I reverted. It's removed.
___
*[Elements/src/Geometry/CsgExtensions.cs, line 270 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKwpBj8L1W8ejJjOLi:-MYMdlgK1gF_mqXxFnfZ:bss4a2b) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/CsgExtensions.cs#L270)):*

does this line do anything?
</blockquote></details>

Removed.
___
*[Elements/src/Geometry/Mesh.cs, line 250 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKpDQx72fOqYf3MSMs:-MYMeG6E1NdcvnM48E9T:b2u6gm) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L250)):*

remove this if we want to allow triangles with duplicate vertices, but this also causes a test to fail.
</blockquote></details>

This is no longer required because I fixed the condition that caused it.
___
*[Elements/src/Geometry/Mesh.cs, line 332 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKtZjhEUyb4u1oSBsN:-MYMeLtsEWPd8KNSUI3k:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L332)):*

I don't think this method is used anywhere, and it seems like it might break triangles if it was?  
</blockquote></details>

Done.

wynged (changes requested):
> Reviewed 3 of 6 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Topography.cs, line 507 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS7XLV-QPiGEWu3huN:-MYS7XLV-QPiGEWu3huO:bkd8x94) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L507)):*
> ```Smalltalk
>         {
>             var topoCsg = this.Mesh.ToCsg();
>             var trim = new Extrude(perimeter, 200000, Vector3.ZAxis, false);
> ```

comment here about why up 200000 and then transform -100000?  took me a second to figure out that we were moving down in the necessary way to be sure we punch all the way through the topo.
___
*[Elements/src/Topography.cs, line 516 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS71HGB4ckA2KLhlzS:-MYS71HGB4ckA2KLhlzT:b-svm9o5) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L516)):*
> ```Smalltalk
> 
>             // We've trimmed the mesh. We now need to
>             // reset the min and max elevation.    
> ```

some extra spaces at end of this line.
___
*[Elements/src/Topography.cs, line 864 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS5JUA61eokxmdMwS2:-MYS5JUA61eokxmdMwS3:bxt524p) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L864)):*
> ```Smalltalk
>         {
>             // Ignore any vertices that have the depth of the sides.
>             var minHeight = this.AbsoluteMinimumElevation.HasValue ? this.AbsoluteMinimumElevation.Value : this.MinElevation - this.DepthBelowMinimumElevation;
> ```

doesn't look like `minHeight` does anything.
___
*[Elements/src/Geometry/CsgExtensions.cs, line 298 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS8DW-0er4VckBf-I7:-MYS8DW-0er4VckBf-I8:b-hlvlfj) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Geometry/CsgExtensions.cs#L298)):*
> ```Smalltalk
>                     mesh.AddTriangle(av, bv, cv);
>                 }
>                 catch (ArgumentException)
> ```

Why are we swallowing these ArgumentExceptions?  Can we detect which polygons will fail instead?  That might be faster in the case where we are looping below. 

If not, and we definitely don't care about these vertices, maybe leave a note about why it's ok to skip these invalid vertices.

Also, any chance that we end up adding vertices, but then we fail and we have orphaned vertices that aren't used in any triangles?
___
*[Elements/test/TopographyTests.cs, line 284 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS9c2U8MffgPmf4A_M:-MYS9c2U8MffgPmf4A_N:b-er9ojj) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/test/TopographyTests.cs#L284)):*
> ```Smalltalk
>             topo.Tunnel(tunnelPath, 15.0);
>             this.Model.AddElement(topo);
>             this.Model.AddElement(new ModelCurve(tunnelPath));
> ```

Are there any useful `Assert`s we can add in this test?  Even if they're contrived might help us in the future.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Topography.cs, line 507 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS7XLV-QPiGEWu3huN:-MYW0PSX-v7RgezaHkOx:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L507)):*

comment here about why up 200000 and then transform -100000?  took me a second to figure out that we were moving down in the necessary way to be sure we punch all the way through the topo.
</blockquote></details>

Done.
___
*[Elements/src/Topography.cs, line 516 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS71HGB4ckA2KLhlzS:-MYW0TnKCeqsKALLHCfi:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L516)):*

some extra spaces at end of this line.
</blockquote></details>

Done.
___
*[Elements/src/Topography.cs, line 864 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS5JUA61eokxmdMwS2:-MYW0aXZ1yJsDUcDkZuM:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Topography.cs#L864)):*

doesn't look like `minHeight` does anything.
</blockquote></details>

Done.
___
*[Elements/src/Geometry/CsgExtensions.cs, line 298 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS8DW-0er4VckBf-I7:-MYW1KXPCk0bz6Q30Zyp:b-wzanwb) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Geometry/CsgExtensions.cs#L298)):*

Why are we swallowing these ArgumentExceptions?  Can we detect which polygons will fail instead?  That might be faster in the case where we are looping below. 

If not, and we definitely don't care about these vertices, maybe leave a note about why it's ok to skip these invalid vertices.

Also, any chance that we end up adding vertices, but then we fail and we have orphaned vertices that aren't used in any triangles?
</blockquote></details>

I've refactored this to test the triangle outside of `AddTriangle` (which coincidentally is the way that we do it in STL creation), and to avoid adding the triangle if the duplicated vertex check fails. It's not great that we _may_ have extra vertices leftover as a result of this but we can't just discard the vertices because they may be a vertex that already exists in the mesh. These "dead" vertices can be cleaned up later with a sweep to see if there are vertices with no associated triangles or they can be left in place and will result in a small bloat in the output geometry buffers.
___
*[Elements/src/Geometry/Mesh.cs, line 272 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKpulb8JLWJB4w9qEQ:-MYW0kvYDSI7K7DEsPgO:b3y0pmy) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L272)):*

any risks to making having merge default to `true` ?  it's the behavior we would rather have in general right?
</blockquote></details>

I'd be ok doing this in the future, but it's still somewhat experimental with this PR.
___
*[Elements/test/TopographyTests.cs, line 284 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS9c2U8MffgPmf4A_M:-MYWDMfu0sAIikmVl-b5:b9konyu) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/test/TopographyTests.cs#L284)):*

Are there any useful `Assert`s we can add in this test?  Even if they're contrived might help us in the future.
</blockquote></details>

Those tests are to ensure that the whole shebang works and they don't contain assertions because any failure will throw an exception which will fail the test, which is what we want to measure. There are however specific things that we want to measure like whether cut and fill volumes are correct and I've added a couple of specific tests for exactly that.

wynged (approved):
> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[Elements/src/Geometry/CsgExtensions.cs, line 298 at r3](https://reviewable.io/reviews/hypar-io/elements/571#-MYS8DW-0er4VckBf-I7:-MYjmWZnC93MPCTKSE_i:b-7m74ok) ([raw file](https://github.com/hypar-io/elements/blob/e9625612f1e48520e23da174921279830e6618be/Elements/src/Geometry/CsgExtensions.cs#L298)):*

I've refactored this to test the triangle outside of `AddTriangle` (which coincidentally is the way that we do it in STL creation), and to avoid adding the triangle if the duplicated vertex check fails. It's not great that we _may_ have extra vertices leftover as a result of this but we can't just discard the vertices because they may be a vertex that already exists in the mesh. These "dead" vertices can be cleaned up later with a sweep to see if there are vertices with no associated triangles or they can be left in place and will result in a small bloat in the output geometry buffers.
</blockquote></details>

nice.
___
*[Elements/src/Geometry/Mesh.cs, line 272 at r1](https://reviewable.io/reviews/hypar-io/elements/571#-MYKpulb8JLWJB4w9qEQ:-MYjm_D77oc3ruwSb1ZY:b-srggdd) ([raw file](https://github.com/hypar-io/elements/blob/da5490c6a71a9bc18e3d28dec4f28947982f8769/Elements/src/Geometry/Mesh.cs#L272)):*

I'd be ok doing this in the future, but it's still somewhat experimental with this PR.
</blockquote></details>

ok gotcha.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/571)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #572, opened 2021-04-19
`ConstructedSolid` doesn't serialize


andrewheumann:
We recently exposed a `ConstructedSolid` SolidOperation. This handily supports imports with planar faces. However, because it relies on its internal `Solid` property, which does not serialize, Elements with ConstructedSolids in their representations aren't passed to subsequent functions. Exposing `Solid` is not straightforward because the Solid type itself is full of bi-directional / circular references.

PR #574, opened 2021-04-26
Serialize the solid associated w/ a constructedSolid

andrewheumann:
BACKGROUND:
- #572 
- We need to be able to pass the Solid held by a Constructed Solid from function to function

DESCRIPTION:
- Ensures the `Solid` of a `ConstructedSolid` serializes.

TESTING:
- Added a test which serializes + deserializes a model containing ConstructedSolid
  
FUTURE WORK:
- `ConstructedSolids` don't convert to IFC correctly, this should be corrected.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/574)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Changelog
 
+## 0.9.0
+
+### Changed
+
+- `ConstructedSolid` serializes and deserializes correctly.
+
 ## 0.8.5
 
 ### Added
+
 - `Elements.Spatial.CellComplex`
 - `Grid2d.ToModelCurves()`
 - Alpha release of `Hypar.Elements.Components`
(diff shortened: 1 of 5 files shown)


andrewheumann (commented):
So, a little bit of scope creep here — after chatting with @ikeough I realized that the serialization was basically working as intended, the problem was with how I was constructing the solid. Each new Face I added was adding new edges + vertices — so a simple cube (as in the test) was creating 24 vertices + 24 edges, instead of 8 / 12. I've added an optional `mergeVerticesAndEdges` argument to `Solid.AddFace`, which will allow easier creation of solids "from scratch" to prevent this kind of topology. 

ADDITIONAL FUTURE WORK:
It is still possible to create bonkers solid topologies when constructing your own solid. We should implement some checks to prevent eg. nonmanifold conditions.
*If* a nonmanifold or open solid is created, there will still be a difference between the solid pre- and post-serialization — in that a given half edge has to have a face associated with it to get serialized. We should probably serialize the "naked" `Right` half-edge in open cases, for consistency's sake — but currently I think we don't officially support "open" solids anyway, so I think it's reasonable not to solve this now. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/574#-:-MZIkBmE7HgQYFd4orh_:biu7y83)* status: 0 of 1 approvals obtained (waiting on @serenayl)

serenayl (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/574)* status: :shipit: complete! 1 of 1 approvals obtained

PR #575, opened 2021-04-27
Add AnalysisImage class

serenayl:
BACKGROUND:
- AnalysisMesh only supports a limited number of vertices when converted to glb. We could have extended this serialization, but given that analyses may end up becoming quite large, opted for another class that would apply analysis colors via a texture image rather than mesh vertices.

DESCRIPTION:
- Adds AnalysisImage class and test
- Adds `InterpolateTexture` member to `Material`, which if false will serialize the material so that pixels render as hard edges rather than blurred interpolation. This is true by default so that the image behavior will not change on existing code.
- Small thing: add vscode settings to gitignore so that local user settings do not pollute the repository.

TESTING:
- Run the AnalysisImage test
- Try using AnalysisImage in your function
- See https://hypar.io/workflows/66091bef-6cf7-4253-96a5-69ae8e3c99d4 for an example of AnalysisImage in use
  
FUTURE WORK:
- N/A

REQUIRED:

COMMENTS:
- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/575)


ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/575)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Material.cs, line 16 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPSqZg6BdD89S6HExD:-MZPSqZg6BdD89S6HExE:b-h3oja0) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Material.cs#L16)):*
> ```Smalltalk
>         /// If false, renders hard pixels in the texture rather than fading between adjacent pixels.
>         /// </summary>
>         /// <value></value>
> ```

`<value>` is the place where we can put your comment above. "If false, renders hard pixels..."
___
*[Elements/src/Material.cs, line 17 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPXc_444aQwvPM6IN7:-MZPXc_444aQwvPM6IN8:b-ydw599) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Material.cs#L17)):*
> ```Smalltalk
>         /// </summary>
>         /// <value></value>
>         public bool InterpolateTexture { get; set; } = true;
> ```

This is a nice addition, but we'll need to add it to the `Material` schema and auto-generate a new `Mesh.g.cs` because the `InterpolateTexture` property will need to be set during JSON deserialization.
___
*[Elements/src/Analysis/AnalysisImage.cs, line 40 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPT6vt2Gf9aJn0oha3:-MZPT6vt2Gf9aJn0oha4:b-b8tayd) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisImage.cs#L40)):*
> ```Smalltalk
>         /// False by default
>         /// </summary>
>         /// <value></value>
> ```

"If false, renders hard pixels..." in the `value`.
___
*[Elements/src/Analysis/AnalysisImage.cs, line 54 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPTyBl0H27KWWdj_GA:-MZPTyBl0H27KWWdj_GB:b-s64l9s) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisImage.cs#L54)):*
> ```Smalltalk
>         /// <param name="id"></param>
>         /// <param name="name"></param>
> ```

If the change below to override `Tessellate` is made, then this constructor is not required (I don't think). It's a duplicate of the existing constructor. You can move the comment about why this class exists onto the class itself if needed.
___
*[Elements/src/Analysis/AnalysisImage.cs, line 66 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPVFra4dZJsPwxvxbS:-MZPVFra4dZJsPwxvxbT:bfc2qee) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisImage.cs#L66)):*
> ```Smalltalk
>         /// Gives an element with a mapped texture.
>         /// </summary>
>         public MeshElement GetMeshElement()
> ```

I recommend you make this `public override void Tessellate(...)` and mark the base version of this method as `virtual`. No need for a new API unless we absolutely need one, and we don't need to create a new element.
___
*[Elements/src/Analysis/AnalysisImage.cs, line 129 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPVg8AE68e9Kc09XD4:-MZPVg8AE68e9Kc09XD5:b-j4yc2e) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisImage.cs#L129)):*
> ```Smalltalk
>             image.Save(imagePath);
> 
>             var material = new Material($"Analysis_{Guid.NewGuid().ToString()}", Colors.White, 0, 0, null, true, true, Guid.NewGuid());
> ```

I would do`this.Material` = new Material(....)` It's a little strange to create the material inside the tesselate method, but it seems a necessary evil.
___
*[Elements/src/Analysis/AnalysisMesh.cs, line 17 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPXsGr2R-RNRHwHNnH:-MZPXsGr2R-RNRHwHNnI:baknt6u) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisMesh.cs#L17)):*
> ```Smalltalk
>     public class AnalysisMesh : GeometricElement, ITessellate
>     {
>         internal List<(BBox3 cell, double value)> _results = new List<(BBox3 cell, double value)>();
> ```

These can be `protected` if `AnalysisMesh` is inheriting. Unless there's a better reason for making them `internal` that I'm not seeing.
___
*[Elements/test/AnalysisImageTests.cs, line 30 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPWHdKEqP3Hxhp7uSp:-MZPWHdL5XXEGxf7P8rL:b-d3kvhg) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/test/AnalysisImageTests.cs#L30)):*
> ```Smalltalk
>             var analysisImage = new AnalysisImage(shape, 10, 10, colorScale, analyze);
>             analysisImage.Analyze();
>             this.Model.AddElement(analysisImage.GetMeshElement());
> ```

When you've made the change to override `Tessellate`, this will just become `this.Model.AddElement(analysisImage)`.

ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/575)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Analysis/AnalysisImage.cs, line 19 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPYE952wm3Rj17uurY:-MZPYE952wm3Rj17uurZ:ba3ub58) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Analysis/AnalysisImage.cs#L19)):*
> ```Smalltalk
>     /// [!code-csharp[Main](../../Elements/test/AnalysisImageTests.cs?name=example)]
>     /// </example>
>     public class AnalysisImage : AnalysisMesh
> ```

I don't think we need to deprecate `AnalysisMesh` just yet, but we should add a note to that class to the effect of "A maximum of 65,535 analytical values is allowed. For an unlimited number of analytical values, use AnalysisImage."

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/575)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Material.cs, line 17 at r1](https://reviewable.io/reviews/hypar-io/elements/575#-MZPXc_444aQwvPM6IN7:-Ma0Wuh6BGqpsx8V1iU-:b-9755fn) ([raw file](https://github.com/hypar-io/elements/blob/141610373b53653a56aa4ea8178c07358200f583/Elements/src/Material.cs#L17)):*

This was done, and while I was at it, also moved `RepeatTexture` and `NormalTexture`
</blockquote></details>

Thank you!

PR #576, opened 2021-04-28
allow isTrimmed to return true for cells outside the boundary

andrewheumann:
BACKGROUND:
- We had a request from some customers to simplify excluding incomplete cells from a grid2d. Prior to this PR, you had to do this to get "untrimmed cells within the boundary":
```
Grid2d.GetCells().Where(c => c.IsTrimmed() && c.GetTrimmedCellGeometry.Count() > 0);
```
which is cumbersome.

DESCRIPTION:
- Adds a `treatFullyOutsideAsTrimmed` flag to `IsTrimmed()` (true by default) which counts cells entirely outside of the boundary as "trimmed." This is left as a flag, in case a user wants the old behavior (for instance, tallying facade panels that are odd shapes and need a different manufacturing process — in this case we would not want to consider non-existent panels.)

TESTING:
- Updated the test 
image

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/576)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,19 @@
 # Changelog
 
+## 0.9.0
+
+### Added
+
+- `Grid2d.IsOutside()`
+
+### Changed
+
+- `Grid2d.IsTrimmed()` now takes an optional boolean parameter `treatFullyOutsideAsTrimmed`
+
 ## 0.8.5
 
 ### Added
+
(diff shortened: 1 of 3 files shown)


cdriesler (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/576)* status: :shipit: complete! 1 of 1 approvals obtained

PR #581, opened 2021-05-06
Set correct last cell typename in Grid1d.DivideByPattern when there i…

nadiia-volyk:
…s no remainder

BACKGROUND:
- If the grid1d length is a multiple of the total pattern length, after DivideByPattern call last cell of grid does not receive a type name according to the pattern.

DESCRIPTION:
- Include last cell segment to patternSegments in Cycle function.
  - After this fix if grid length is equal to first pattern item length, NullReferenceException is thrown in DivideWithPatternAndOffset method. Add handling for this case.

TESTING:
- Added a new test, which used to fail, and now succeeds
  
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/581)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/581)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/581)* status: :shipit: complete! 1 of 1 approvals obtained

PR #590, opened 2021-05-27
better handle closestPosition on polyline axes

andrewheumann:
BACKGROUND:
- Previously, Grid.SplitAtPoint() and other methods would wind up splitting at the wrong point, due to their reliance on a dumb algorithm, which ignored the length of segments and calculated a position that assumed a straight-line axis.

DESCRIPTION:
- Adds a more sophisticated ClosestPosition algorithm that actually calculates the closest point on a polyline axis and computes the cumulative length up to that point.

TESTING:
- Added a new test that failed in previous versions of Elements, and passes now. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/590)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.9.1
 
 ### Added
+
 - `GetSolid()` method on GeometricElement that returns the Csg solid.
 - `Polygon.ToTransform()`
 - `Elements.Anaysis.AnalysisImage`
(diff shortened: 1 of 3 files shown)


cdriesler (approved):
ty for fixing this! Grid had enough other tools, so this was never a blocker, but I did want to use this for some other functions

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/590#-:-MajVQiPEdQo0ubuBrDO:b1kiqco)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #585, opened 2021-05-18
Solid silently fails to generate glb (Regression)


andrewheumann:
**Describe the bug**
A `ConstructedSolid` that used to generate a GLB easily prior to #573 now fails silently.

**To Reproduce**
check out branch `glb-generation-fails` and run `SolidTests.ConstructedSolidProducesValidGlb`

**Expected behavior**
It should produce a glb that looks like this, as it used to: 
![image](https://user-images.githubusercontent.com/31935763/118577114-cd56fd00-b757-11eb-94bf-321d60fedb72.png)

**Additional context**
I tracked this as far as the `Csg.Solid ToCsg(this Solid solid)` method, but I was unable to isolate a specific problem. In the previous version, the sold was triangulated, resulting in a csg with 74 polygons. This solid proceeds through `ProcessSolidsAsCSG` with no problem. The new version, post vertex merging, creates a solid with 10 polygons, but when it goes through `ProcessSolidsAsCsg` it outputs an invalid CSG with 0 faces from `csg.Union(solids);` That's all I know!

ikeough:
This seems to be a clockwise wound polygon face created from the solid.
![image](https://user-images.githubusercontent.com/1139788/118904232-51cb8c00-b8ce-11eb-9b25-a1c549d540df.png)
And the original data in the JSON is CW as well.
![image](https://user-images.githubusercontent.com/1139788/118904409-b0910580-b8ce-11eb-8581-fb55163af218.png)
I believe these CW wound polygons would cause a CSG that is inside out, and this is perceived by the CSG algorithm to be a void, so the Solid voids itself during the Union call.

ikeough:
The cause of this is that Solids with planar faces (all of our solids) are no longer having those faces tessellated when converted to CSG unless the face has a hole. My guess would be that the tessellator took the CW wound data and produced enough facets that the CSG operation new what to do, or that the facets were generated with the correct orientation. The data has a lot of colinear points. It's also possible that we can't send a polygon with colinear vertices directly through to CSG.

andrewheumann:
The polygon in the JSON is not incorrectly wound. It is CCW with respect to the -Z — you're looking at the bottom face of the Brep. I double-checked the input data: all polygons are wound correctly, CCW with respect to outward-facing normals

![winding](https://user-images.githubusercontent.com/31935763/119487684-b2702400-bd27-11eb-89b0-6c55e87186c2.gif)
.

wynged:
[...]

PR #591, opened 2021-05-28
Don't perform unnecessary csg booleans

andrewheumann:
BACKGROUND:
- GLTF creation was failing sometimes — either creating a totally empty model, or rendering only a few odd triangulated pieces. https://github.com/hypar-io/Elements/issues/585. It appears this is due to the CSG library's own Union operation, possibly related to solids with non-triangular and/or non-convex faces, or laminae.
- This is not a complete fix — elements with multiple solids may still fail in this way. But considering that most elements have a single solid operation as representation, this should fix a great many cases.
- UPDATE: once this first fix was in, it revealed another issue by breaking one of @wynged's tests: we were generating unnecessary vertices in some cases, because our check for whether to use the tesselate method was just checking if `Inner != null` and not checking for an empty array.

DESCRIPTION:
- Only call `CSG.Union` and `CSG.Substract` [sic] if multiple solids / voids are present.
- Only use tesselate method if Inner is not null and not empty.

TESTING:
- All tests are passing
- Two new tests are added which demonstrate the erroneous behavior noted in #585. Both tests fail prior to this change and pass now. 
  
FUTURE WORK:
- We should still get to the bottom of why csg.Union() is failing for some shapes. This fix only covers the case where there is only one object with no voids — others may still run into this issue.
 
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/591)


wynged (changes requested):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/591)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/GeometricElement.cs, line 94 at r1](https://reviewable.io/reviews/hypar-io/elements/591#-ManW0VF-zXuycjYKuIR:-ManW0VG4zdP6JjA5fxr:b-hh352r) ([raw file](https://github.com/hypar-io/elements/blob/2dde9bae320883c109421ab1e9d68e8236cbeb43/Elements/src/GeometricElement.cs#L94)):*
> ```Smalltalk
>                                                       .Select(op => op._csg.Transform(o.Transform.ToMatrix4x4())))).ToArray();
>             }
>             // Only try CSG booleans if we have multiple solids.
> ```

Nit, but this comment feels backwards since the line directly below is about a single solid.  `Don't try CSG booleans if we only have one solid`
___
*[Elements/src/GeometricElement.cs, line 103 at r1](https://reviewable.io/reviews/hypar-io/elements/591#-ManWE-u3moQo42oK4J2:-ManWE-u3moQo42oK4J3:boo7stq) ([raw file](https://github.com/hypar-io/elements/blob/2dde9bae320883c109421ab1e9d68e8236cbeb43/Elements/src/GeometricElement.cs#L103)):*
> ```Smalltalk
>                 csg = csg.Union(solids);
>             }
>             if (voids.Count() > 0)
> ```

is voids guaranteed to be non-null?
___
*[Elements/test/SolidTests.cs, line 256 at r1](https://reviewable.io/reviews/hypar-io/elements/591#-Man_bz857GzDae1kb7h:-Man_bz93mK0qhbJgO5i:bopb0hh) ([raw file](https://github.com/hypar-io/elements/blob/2dde9bae320883c109421ab1e9d68e8236cbeb43/Elements/test/SolidTests.cs#L256)):*
> ```Smalltalk
>         {
>             Name = nameof(ConstructedSolidProducesValidGlb);
>             var allPolygons = JsonConvert.DeserializeObject<List<(Polygon outerLoop, List<Polygon> innerLoops)>>(File.ReadAllText("../../../models/Geometry/TroubleSolidPolygons.json"));
> ```

could we change the name to indicate what kind of trouble?  "JaggedPerimeter" or "StaggeredEdges"

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/591)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @wynged)

___
*[Elements/src/GeometricElement.cs, line 103 at r1](https://reviewable.io/reviews/hypar-io/elements/591#-ManWE-u3moQo42oK4J2:-Mao8yKQDauDFu7yNSWF:bczziyd) ([raw file](https://github.com/hypar-io/elements/blob/2dde9bae320883c109421ab1e9d68e8236cbeb43/Elements/src/GeometricElement.cs#L103)):*

is voids guaranteed to be non-null?
</blockquote></details>

Yes — as long as SolidRepresentations is not null, which I think is assumed in this code. 
___
*[Elements/test/SolidTests.cs, line 256 at r1](https://reviewable.io/reviews/hypar-io/elements/591#-Man_bz857GzDae1kb7h:-Mao99PCAYZi3123rAjX:bouvgci) ([raw file](https://github.com/hypar-io/elements/blob/2dde9bae320883c109421ab1e9d68e8236cbeb43/Elements/test/SolidTests.cs#L256)):*

could we change the name to indicate what kind of trouble?  "JaggedPerimeter" or "StaggeredEdges"
</blockquote></details>

They're not really jagged or staggered... Have renamed to reflect their purpose.

wynged (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/591)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

Issue #493, opened 2021-01-26
Provide a surface with holes Representation


anthonyhauck:
**Is your feature request related to a problem? Please describe.**
Space planners wish to represent 2D spatial zones that may have gaps created by other immovable space allocations, such as building service cores, horizontal circulation, or previously allocated space. Elements lacks a managed representation of a displayable surface with holes in it.

**Describe the solution you'd like**
An Element representation similar to Lamina that would accept a list of openings similar to Floor creation.

**Describe alternatives you've considered**
Overlapping Polygons, but much management of their relationship is necessary by the developer.

**Additional context**
This came as a customer request from the Jacobs Krakow development team on 2021.01.25.

ikeough:
Solved in #592

PR #592, opened 2021-05-28
Lamina with holes

andrewheumann:
BACKGROUND:
- Addresses #493 
- Allows Laminae to have Voids, so they can be constructed from a profile.

DESCRIPTION:
- Adds `voids` to the lamina schema, and updates the associated methods and constructors.

TESTING:
- I tested to ensure I could create a lamina from a profile, and it would show as expected
- I tested to ensure I could create a lamina using the old constructor, and nothing would break. 
  
FUTURE WORK:
- It feels a little like a hack to be constructing two-sided laminae instead of single-faced objects, (which obviously wouldn't work well in a "Solid" context. At some point we may want single-sided shapes. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/592)


wynged (approved):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/592)* status: :shipit: complete! 1 of 1 approvals obtained

PR #593, opened 2021-05-31
Add default arguments to codegenned constructors

andrewheumann:
BACKGROUND:
- When working with generated element classes, it's kind of a pain to manually input all the default arguments.

DESCRIPTION:
- Codegens default arguments in the constructor for element types that inherit directly from Element + GeometricElement
- Updates a few validators to make sure those properties get valid values
- I also fixed a malformed xml end tag, and updated the changelog so that it was pointing to the correct version (we are still building 0.9.1 alphas, so our version changes should go under 0.9.1, not 0.9.2 yet.)

TESTING:
- I used a local build of the CLI with this elements version to codegen a range of elements from the schema manager, and ensured that they compiled without error, and could be used without supplying inherited arguments.
   - Before: `var sb = new SpaceBoundary(rect, null, new Transform(), BuiltInMaterials.Default, rep, false, System.Guid.NewGuid(), null);`
   - After: `var sb = new SpaceBoundary(rect, null, representation: rep);`
  
FUTURE WORK:
- Users new to the system may not understand how representations work — we may need to find some way to let novice users know that they are required to provide a representation for geometric elements if they want them to show up. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/593)


wynged (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/593)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements.CodeGeneration/src/HyparFilters.cs, line 101 at r1](https://reviewable.io/reviews/hypar-io/elements/593#-Mb773cEEg1GACdnUYYn:-Mb773cEEg1GACdnUYYo:b-9ob2wk) ([raw file](https://github.com/hypar-io/elements/blob/7a09f578e5b41bd6749f0247dfd3b4709a3d7264/Elements.CodeGeneration/src/HyparFilters.cs#L101)):*
> ```Smalltalk
>                 {"transform", "null"},
>             };
>             if ((baseClass == "Element" || baseClass == "GeometricElement") && defaults.ContainsKey(input))
> ```

Why restrict ourselves to Element and GeometricElement?  This seems like it would work with any base constructor that had some of these properties.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/593)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @wynged)

___
*[Elements.CodeGeneration/src/HyparFilters.cs, line 101 at r1](https://reviewable.io/reviews/hypar-io/elements/593#-Mb773cEEg1GACdnUYYn:-Mb77Qu58O2ViDfxRRB_:bigcho7) ([raw file](https://github.com/hypar-io/elements/blob/7a09f578e5b41bd6749f0247dfd3b4709a3d7264/Elements.CodeGeneration/src/HyparFilters.cs#L101)):*

Why restrict ourselves to Element and GeometricElement?  This seems like it would work with any base constructor that had some of these properties.
</blockquote></details>

I'm being conservative and sticking to just Element + GeometricElement, since otherwise you might accidentally create broken code. If a class had one of those properties, eg Name or Transform, and it was in the middle of the list of arguments, you'd stick a default argument in the middle of your constructor, which wouldn't compile. Default arguments have to come at the end

PR #594, opened 2021-06-01
Allow tolerance in collinearity point count check

andrewheumann:
BACKGROUND:
- An exception could be thrown when doing a Line/Line intersect between two lines nearly meeting at a point, due to tolerance issues.

DESCRIPTION:
- Adds an internal Vector3 comparer, which can be used in a `.Distinct` call to permit a little tolerance.
- Counts distinct vertices only when doing a collinearity check, to avoid the exception.

TESTING:
- Added a new test drawn from the real-world case that was triggering this issue — it failed prior to this fix and now passes.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/594)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/594)* status: :shipit: complete! 1 of 1 approvals obtained

PR #595, opened 2021-06-01
Discriminator warnings

wynged:
BACKGROUND:
- When working with local elements it is frustrating to deserialize a model where many elements are unknown, because you get a caught exception for every single element that can't be serialized.

DESCRIPTION:
- Rather than throwing a new exception when an object fails to be de-serialized, we gather the messages up in a private static list and when we access the list after de-serialization we clear it.

TESTING:
- Functionality should be unchanged.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/595)


wynged:
This is old and hasnt' been an issue lately, closing

PR #597, opened 2021-06-07
fix transform orientation

andrewheumann:
BACKGROUND:
- Transforms were coming out after our glb instancing changes. It turns out, not only were our instances-of-instances coming out oriented wrong, even just a plain old single-level-deep of content elements was also coming out wrong (for example: make a bunch of instances of a duck, duck would now be wrong way).

DESCRIPTION:
- Fixes the transform orientation for our WIP instance-merged-glbs branch

Our GLBs are composed of pairs of layers — one "layer" operates in the +Z up space, and is where we do things like instance positioning transforms, and then at the next layer up we switch everything back to being +Y up. This is my best stab at understanding what's happening:
image
- a little cleanup elsewhere

TESTING:
- Modified test to include the creation of a "1 level deep" glb as well, and using that as the basis for the "2 levels deep" glb.


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/597)* status: :shipit: complete! 1 of 1 approvals obtained

PR #599, opened 2021-06-08
Add more explanation for IsClockwise.

ikeough:
BACKGROUND:
- Daniel Fink noted that the `Polygon.IsClockwise` method doesn't work for 3d polygons.

DESCRIPTION:
- This adds a note (yes, we should add a function), that clarifies that the method is only appropriate for 2D polygons.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/599)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/599)* status: :shipit: complete! 1 of 1 approvals obtained

PR #605, opened 2021-06-23
3D Polygon Trimming

ikeough:
BACKGROUND:
- A general 3D polygon trimming method would be useful for scenarios like building the allowable envelope for a building from the intersection of sky planes.

DESCRIPTION:
- This PR implements polygon trimming in 3D against a set of polygons. 
- This PR makes `Polyline.Plane()` virtual and overrides `Polygon.Plane()` to use `Plane.Normal`.

TESTING:
- See the `PolygonTests.PolygonIsTrimmedByPolygons()` test method.
  
FUTURE WORK:

REQUIRED:

In the image below you can see a star oriented in 3D within a set of vertical planes formed by the faces of an extruded hexagon. The trim operation on the star results in a turquoise truncated star, while reversing the clipping planes and running the same trim, results in 5 separate star point polygons. Note that two of the star tips in this scenario are intersected by two of the clipping planes forming concave trims.

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)


andrewheumann (changes requested):
> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 244 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-McuZ-b17yI95IIDpKD6:-McuZ-b17yI95IIDpKD7:b-qlkkxf) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L244)):*
> ```Smalltalk
>         /// <param name="result">The points resulting from the intersection
>         /// of the two polygons.</param>
>         /// <returns>A collection of point sorted along the polygon's plane.</returns>
> ```

point => points
___
*[Elements/src/Geometry/Polygon.cs, line 250 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcu_XPr7UaZSpdP0dAu:-Mcu_XPr7UaZSpdP0dAv:bhxpynp) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L250)):*
> ```Smalltalk
>             result = new List<Vector3>();
>             var targetP = polygon.Plane();
>             var d = this.Plane().Normal.Cross(targetP.Normal).Unitized();
> ```

I realize this is a private method, but do we need to protect against trying to do a 3d intersect with a coplanar polygon? This would give a zero-length cross product.
___
*[Elements/src/Geometry/Polygon.cs, line 708 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcud-jaEK-R_lt54vSS:-Mcud-jaEK-R_lt54vST:b-z2zd31) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L708)):*
> ```Smalltalk
>         /// Trim the polygon with a collection of polygons.
>         /// </summary>
>         /// <param name="polygons">A collection of trimming polygons.</param>
> ```

I think it would be good to be more descriptive about the 3d nature of this method — I think it would be easy to look at this and think this was a 2d trim or boolean operation. 
___
*[Elements/src/Geometry/Polygon.cs, line 725 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-McudIra0GQ_pj9UslK8:-McudIra0GQ_pj9UslK9:bpyu5ug) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L725)):*
> ```Smalltalk
>             foreach (var p in polygons)
>             {
>                 var plane = p.Plane();
> ```

Here it would probably be good to check / warn / throw if this polygon and its trimming polygon are coplanar.

ikeough:
Fixed an issue identified during testing where single polygon trims did not work because resulting segments "behind" the plane which did not intersect the plane were not discarded.

image

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 244 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-McuZ-b17yI95IIDpKD6:-McvSEmiE8C6O6G_BANL:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L244)):*

point => points
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Polygon.cs, line 250 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcu_XPr7UaZSpdP0dAu:-McvSONhDRVvdj3QH6kj:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L250)):*

I realize this is a private method, but do we need to protect against trying to do a 3d intersect with a coplanar polygon? This would give a zero-length cross product.
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Polygon.cs, line 708 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcud-jaEK-R_lt54vSS:-McvSIqTCguvD4ESXK6Q:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L708)):*

I think it would be good to be more descriptive about the 3d nature of this method — I think it would be easy to look at this and think this was a 2d trim or boolean operation. 
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Polygon.cs, line 725 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-McudIra0GQ_pj9UslK8:-McvSJe4EIbjisprHpT8:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L725)):*

Here it would probably be good to check / warn / throw if this polygon and its trimming polygon are coplanar.
</blockquote></details>

Done.

andrewheumann (changes requested):
This is working much better, but it's still pretty easy to break it: 

One plane totally occluded by another:


Plane that passes through completely, but would intersect another piece of a convex polygon if extended:


Polygons that intersect each other:


JSON dumps of all of the above cases here: https://www.dropbox.com/sh/jo9pk8zexzmegrl/AACM9mDOk8pq3md9CMFGNPIqa?dl=0

Modeled on 

```
    struct SavedInputs
    {
        public Polygon polygon;
        public string error;
        public List<Polygon> trimShapes;
    }
```

> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605#-:-Mcx_N173FYeZxBPhok-:b-mkfu95)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 708 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcud-jaEK-R_lt54vSS:-McxYoc_1mQY_K7Xt90f:b-khelib) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L708)):*

Done.
</blockquote></details>

Sorry to nag on this one, I still think something like —
"Trim the polygon with a collection of polygons that intersect with it in space. Portions of the polygon in the direction of the trimming polygon's normal will be kept, and those on the other side will be clipped."

andrewheumann (commented):
> Reviewed 1 of 3 files at r2, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

andrewheumann (commented):
> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

ikeough:
What would you expect the behavior of the first one to be, with one plane fully occluding another? I would expect the result to be on the polygon fully to the "right".

ikeough:
Do all of the above fail with exceptions from the polygonizer, or in other ways?

andrewheumann (commented):
I think they're all polygonizer exceptions. The polygonizer was very much designed with the assumption that you always have "complete loops" — you can have single edges around the outside, but it expects that all internal edges will be doubled. It was designed to find "all" the polygons, not just one — so that may be why you're running into trouble with it. This is the case it was designed for:



I have no issue with you updating the polygonizer to be more robust to the kinds of scenarios we're hitting here — but it might be easier to produce a case closer to what it expects, and then just pick which resulting polygon is the one you want to keep. 

In the first case, I'd expect this to be the result:



> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605#-:-Mcyu1pPCxac353vR7ln:b7x0box)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

ikeough:
More handled cases:
image
image
image

andrewheumann (commented):
looking WAY better — I had to work a lot harder to break it this time :) 


https://www.dropbox.com/s/w6r0dmcnyl7kqwx/171056.json?dl=0 (exception)

https://www.dropbox.com/s/zmt51spvby0ldgf/171226.json?dl=0 (should have split, didn't)

and feel free to use the test rig: https://dev.hypar.io/workflows/0195965a-7355-4ede-bb09-a7adb7c7d99f

> Reviewed 3 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605#-:-MdIdpc4AU7V7QrFSDXA:b-iywcqw)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

andrewheumann (changes requested):
> Reviewed 5 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 708 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcud-jaEK-R_lt54vSS:-MdNO3Iw9gDUyXuVZpGs:b-zhw551) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L708)):*

Sorry to nag on this one, I still think something like —
"Trim the polygon with a collection of polygons that intersect with it in space. Portions of the polygon in the direction of the trimming polygon's normal will be kept, and those on the other side will be clipped." 
</blockquote></details>

Still think this needs to be more descriptive.
___
*[Elements/src/Geometry/Polygon.cs, line 409 at r4](https://reviewable.io/reviews/hypar-io/elements/605#-MdNM0IL6CZmfddmOuAw:-MdNM0IL6CZmfddmOuAx:bhokmnx) ([raw file](https://github.com/hypar-io/elements/blob/4e698a9e05ebfba2611ebde91ae9d023edb977b5/Elements/src/Geometry/Polygon.cs#L409)):*
> ```Smalltalk
>                     if (unique)
>                     {
>                         if (!xsects.Contains(result))
> ```

This should probably use the overloaded version of `Contains` that takes an EqualityComparer — we already have a `Vector3Comparer` that could be used. 
___
*[Elements/src/Geometry/Polygon.cs, line 773 at r4](https://reviewable.io/reviews/hypar-io/elements/605#-MdNMairChdYGitp3MMU:-MdNMairChdYGitp3MMV:b-fjqnmw) ([raw file](https://github.com/hypar-io/elements/blob/4e698a9e05ebfba2611ebde91ae9d023edb977b5/Elements/src/Geometry/Polygon.cs#L773)):*
> ```Smalltalk
>             foreach (var result in results)
>             {
>                 var skip = result.Count % 2 == 0 ? 2 : 1;
> ```

Leave a comment explaining this skip logic

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 409 at r4](https://reviewable.io/reviews/hypar-io/elements/605#-MdNM0IL6CZmfddmOuAw:-MdO15li3ouMbbg8EO1i:be57b1a) ([raw file](https://github.com/hypar-io/elements/blob/4e698a9e05ebfba2611ebde91ae9d023edb977b5/Elements/src/Geometry/Polygon.cs#L409)):*

This should probably use the overloaded version of `Contains` that takes an EqualityComparer — we already have a `Vector3Comparer` that could be used. 
</blockquote></details>

I don't think it's necessary. Vector3 implements IEquatable, which is the default comparison used. You only need that overload if you want to specify _another_ comparer to use.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 409 at r4](https://reviewable.io/reviews/hypar-io/elements/605#-MdNM0IL6CZmfddmOuAw:-MdO36Cr6f4FABaW1zz5:b-j9c4yf) ([raw file](https://github.com/hypar-io/elements/blob/4e698a9e05ebfba2611ebde91ae9d023edb977b5/Elements/src/Geometry/Polygon.cs#L409)):*

I don't think it's necessary. Vector3 implements IEquatable, which is the default comparison used. You only need that overload if you want to specify _another_ comparer to use.
</blockquote></details>

oh — didn't realize Vector3 Equals already took into account tolerance. ✅

ikeough:
Internal clipping poly intersection is now handled as well.
image

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 708 at r1](https://reviewable.io/reviews/hypar-io/elements/605#-Mcud-jaEK-R_lt54vSS:-MdTgawnBiL4dVus7Njb:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/9cfb7d30dd3b0568bb082fb770d0f800cc45f944/Elements/src/Geometry/Polygon.cs#L708)):*

Still think this needs to be more descriptive.
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Polygon.cs, line 773 at r4](https://reviewable.io/reviews/hypar-io/elements/605#-MdNMairChdYGitp3MMU:-MdTg_oLBzPqSiU17GD4:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/4e698a9e05ebfba2611ebde91ae9d023edb977b5/Elements/src/Geometry/Polygon.cs#L773)):*

Leave a comment explaining this skip logic
</blockquote></details>

Done.

ikeough:
Solved this tricky condition as well where polygons make closed regions as a result of intersections with the polygon and the other polygons.
image

andrewheumann (approved):
> Reviewed 3 of 5 files at r5, 1 of 2 files at r6, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/605)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Polygon.cs, line 753 at r7](https://reviewable.io/reviews/hypar-io/elements/605#-MdTvAS-AEmpg5u9GZPR:-MdTvAS06jN5Ypu68SQp:bfcpjgj) ([raw file](https://github.com/hypar-io/elements/blob/710b9331836a2c39aaec5d8eceb56d9923aa3ad5/Elements/src/Geometry/Polygon.cs#L753)):*
> ```Smalltalk
>         /// <param name="trimEdges">A collection of vertex pairs representing all edges in the timming graph.</param>
>         /// <returns>A collection of polygons resulting from the trim or null if no trim occurred.</returns>
>         public List<Polygon> TrimmedTo(IList<Polygon> polygons, out List<Vector3> intersections, out List<(Vector3 from, Vector3 to)> trimEdges)
> ```

I think the final two outs are likely to be often discarded (as you already do in most of your tests). To avoid the `out _, out _` can we do an overload with your original method signature?
```
        public List<Polygon> TrimmedTo(IList<Polygon> polygons)
        {
            return TrimmedTo(polygons, out _, out _);
        }
``` 
___
*[Elements/test/PolygonTests.cs, line 1455 at r7](https://reviewable.io/reviews/hypar-io/elements/605#-MdTxsi1D8ohS4l-Y4NJ:-MdTxsi1D8ohS4l-Y4NK:baynvrr) ([raw file](https://github.com/hypar-io/elements/blob/710b9331836a2c39aaec5d8eceb56d9923aa3ad5/Elements/test/PolygonTests.cs#L1455)):*
> ```Smalltalk
>             }
> 
>             var result1 = bigPoly.TrimmedTo(splitters, out _, out _);
> ```

if you do take me up on the overload suggestion, clear out these "out outs"  from the tests too

PR #609, opened 2021-06-29
Glb naming + structure adjustments to support content element selection

andrewheumann:
BACKGROUND:
- Element instances of content elements have not been selectable or transformable in explore, because there was no way to map from the mesh in three.js to its corresponding element. 
- Designed to work in tandem with the changes in https://github.com/hypar-io/Explore/pull/1463

DESCRIPTION:
- Makes some changes to the way instances of content elements are merged into the GLB — changing the node structure slightly (so that there is a single parent node whose transform is the instance's transform) and naming the parent node with the element instance's Id.

TESTING:
- I built a function with this version of elements locally, created some content elements, and verified that I could select them and override their transforms in Explore. 
- I tested normal and nested content elements (where the GLB itself has internal instancing) and verified they appeared correctly in the output. 
  
FUTURE WORK:
- We'll need to update the version of elements in functions that want their content elements to be selectable.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/609)


cdriesler (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/609)* status: :shipit: complete! 1 of 1 approvals obtained

PR #615, opened 2021-07-02
Support non XY plane graph polygonization.

ikeough:
BACKGROUND:
When I integrated polygon trimming into the allowable envelope calculations, the first problem that it ran into was intersecting vertical surfaces. Digging into the `HalfEdgeGraph2d` code, I found that the plane angle calculation that we use during graph construction was assuming that the edges were in the XY plane. This resulted in a failure to construct an appropriate edge graph, and correspondingly, the inability to trim vertical polygons. I suspect that trim accuracy would also fall off as planes approach the vertical.

DESCRIPTION:
- This PR adds a new parameter to the `HalfEdgeGraph2d.Polygonize` method, that allows the user to specify a `normal` which will be used during edge traversal. If one is not provided, `Vector3.ZAxis` will continue to be used.

TESTING:
- See the test `PolygonTests.EnclosingPlanesTrimToFormVolume` whose result is shown below.
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/615)


andrewheumann (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/615)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Spatial/HalfEdgeGraph2d.cs, line 174 at r2](https://reviewable.io/reviews/hypar-io/elements/615#-MdcdSNE7XUTLzb-vwJ3:-MdcdSNE7XUTLzb-vwJ4:b-ce1alo) ([raw file](https://github.com/hypar-io/elements/blob/46d8d537bd5d057a4b63ae3897b71b400c45a04b/Elements/src/Spatial/HalfEdgeGraph2d.cs#L174)):*
> ```Smalltalk
>         /// valid.</param>
>         /// <param name="normal">The normal of the plane in which graph traversal for polygon construction will occur.
>         /// If no angle is provided, the +Z axis is used.</param>
> ```

If no angle => If no normal

PR #620, opened 2021-07-13
Update Content Element Schema to use Symbols

andrewheumann:
BACKGROUND:
- The schema on SchemaManager had been updated with the concept of "Symbols," but this was not reflected in Elements code.

DESCRIPTION:
- Pulls in the updated schema and re-generates classes.
- Makes some changes to the .cs half of the ContentElement class to fix compilation errors and ensure backward compatibility with the old constructor.
- several other codegenned classes saw minor changes to their constructors thanks to #593 

TESTING:
- all tests pass. Not much else to test yet.
  
FUTURE WORK:
- Our current internal type codegen strategy doesn't play well with schemas on the schema manager. We should add support for schemas that live on the schema manager. 
- Our current internal type codegen strategy is fragile. If running `.build-types.sh` results in code that doesn't compile, it can't be run again (because the project depends on codegen, which depends on elements) 

 
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/620)


mattc (approved):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/620)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #612, opened 2021-07-01
Investigate .net core DXF reading/writing lib.


ikeough:
The library that we use for DXF reading/writing, netDXF has a .net framework dependency. Although it works with .net core, it emits build warnings. It also has not been updated since 2018. Can we investigate moving to https://github.com/ixmilia/dxf, which is .net core compliant and has more recent updates?

andrewheumann:
We now use IxMilia instead of netDXF.

PR #621, opened 2021-07-13
Migrate from NetDxf to IxMilia.DXF

andrewheumann:
BACKGROUND:
- #612 - there is a more recently maintained, .net core compatible library we could use for DXF writing.

DESCRIPTION:
- Changes the Elements.Serialization.DXF project to utilize IxMilia.DXF instead of NetDxf

TESTING:
- I verified that existing tests produced valid DXFs that were equivalent to those produced via the old method.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/621)


andrewheumann:
I fucked up this branch. will resubmit

PR #622, opened 2021-07-13
use IxMilia instead of netDXF

andrewheumann:
BACKGROUND:
- #612 - there is a more recently maintained, .net core compatible library we could use for DXF writing.

DESCRIPTION:
- Changes the Elements.Serialization.DXF project to utilize IxMilia.DXF instead of NetDxf

TESTING:
- I verified that existing tests produced valid DXFs that were equivalent to those produced via the old method.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/622)

PR #624, opened 2021-07-14
DXF from Content

andrewheumann:
BACKGROUND:
- We need to be able to produce DXF drawings from Models.

DESCRIPTION:
- builds on #620 and #622
- Fixes some issues w/ codegen for the Symbols property 
- Adds DXF converters for `ContentElement`s + `ElementInstance`s
- Removes the DXF converter for Floors
- Adds a generic `GeometricElement` converter
- Refactors `IRenderDxf.TryAddDxfEntity` to take a copy of the document and return void, so that a single element can create multiple entities, and access other properties of the document (like its block table.)

TESTING:
- This version of Elements / DXF is already running on this function and on the export server on my stack:
https://dev.hypar.io/workflows/40d35cec-4dbe-45c0-befe-985b5a60c11b?api=ah-sand
  
FUTURE WORK:
- None of these conversions take into account camera range or orientation. All assume a top view and render regardless of position. 
- Support for layer settings
- More special converters for cases like WallByProfile

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/624)


cdriesler (approved):
> Reviewed 2 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/624)* status: :shipit: complete! 1 of 1 approvals obtained

PR #626, opened 2021-07-19
Dxf layer support

andrewheumann:
BACKGROUND:
- We want to be able to create DXFs with layers. This is a continuation of #624, which adds this capability

DESCRIPTION:
- Creates a new `Elements.Serialization.MappingConfiguration` class that can contain information about mapping to other formats, and adds this as a property to the DXF Renderer
- Fixes some compatibility issues with Autocad to ensure documents render correctly. 
- For now, `MappingConfiguration`s are stored on a server, and there's no mechanism to specify which one you want for a DXF export — we use a hard-coded configuration.

TESTING:
- The example workflow is updated to use this along with an updated API stack
 https://dev.hypar.io/workflows/40d35cec-4dbe-45c0-befe-985b5a60c11b?api=ah-sand
  
FUTURE WORK:
- Only certain values for the Lineweight property seem to be valid — we should maybe restrict the values to those that can be used. 
- Extend `MappingConfiguration` to have details about other formats, like IFC
- Update API to use this information
- Create API for setting/editing MappingConfigurations
- Update UI to enable setting user, workflow, or function-level mapping configurations

REQUIRED:
- [X] All changes are up to date in `CHANGELOG.md`. (leaving this one lumped under DXF support)

COMMENTS:
- There's a bit of a time pressure here — We've promised them a first version of this they can use by tuesday.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/626)


cdriesler (approved):
> Reviewed 9 of 11 files at r1, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/626)* status: :shipit: complete! 1 of 1 approvals obtained

PR #633, opened 2021-07-23
Set Operations

ikeough:
BACKGROUND:
When working with sets of edges we would like a way to classify them according to their set inclusion. An example of this is the processing of two overlapping bounded spaces like a polygon or a brep face.

DESCRIPTION:
- Introduces set classifications and a couple of exemplary methods for working on polygon splitting. 
- Several helper methods to aid in random planar ray casting are added to classify linear segments. 
- A method for building a `HalfEdgeGraph2d` from a classified set is added.

TESTING:
A small set of tests has been included in `SetOperationsTests` that demonstrates the use of these new APIs.
  
FUTURE WORK:
Currently this work assumes `Vector3` and `Polygon` types. They should be made generic so that they can support line segments, solid face edges, or cell complex edges. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/633)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 - Support for DXF from many basic elements.
 - `SetClassification`
-- `SetOperationsClassifySegments2d(Polygon a, Polygon b, Func<(Vector3 from, Vector3 to, SetClassification classification), bool> filter = null)`
+- `SetOperations.ClassifySegments2d(Polygon a, Polygon b, Func<(Vector3 from, Vector3 to, SetClassification classification), bool> filter = null)`
+- `SetOperations.BuildGraph(List<(Vector3 from, Vector3 to, SetClassification classification)> set, SetClassification shared)`
 - `RandomExtensions.NextRayInPlane(this Random random, Vector3 origin, Vector3 normal)`
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Added
 
 - Support for DXF from many basic elements.
+- `SetClassification`
+- `SetOperationsClassifySegments2d(Polygon a, Polygon b, Func<(Vector3 from, Vector3 to, SetClassification classification), bool> filter = null)`
+- `RandomExtensions.NextRayInPlane(this Random random, Vector3 origin, Vector3 normal)`
 
 ### Changed
 
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -255,6 +255,38 @@ namespace Elements.Geometry
         }
 
         /// <summary>
+        /// Intersect this polygon with the provided polygon in 2d.
+        /// </summary>
+        /// <param name="polygon">The target polygon.</param>
+        /// <param name="results">The points resulting from the intersection of
+        /// the two polygons.</param>
+        /// <param name="includeEnds">Should intersection with segment ends be included?</param>
+        /// <returns>True if this polygon intersects the provided polygon,
+        /// otherwise false.</returns>
+        internal bool Intersects2d(Polygon polygon,
+                                   out List<(Vector3 result, int aSegmentIndices, int bSegmentIndices)> results,
+                                   bool includeEnds = false)
+        {
+            var aSegs = this.Segments();
+            results = new List<(Vector3, int, int)>();
+
+            for (var i = 0; i < aSegs.Length; i++)
+            {
+                var a = aSegs[i];
+                var bSegs = polygon.Segments();
+                for (var j = 0; j < bSegs.Length; j++)
+                {
+                    var b = bSegs[j];
+                    if (a.Intersects(b, out Vector3 result, includeEnds: includeEnds))
+                    {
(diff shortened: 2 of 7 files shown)


andrewheumann (changes requested):
> Reviewed 4 of 5 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/633)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/RandomExtensions.cs, line 32 at r1](https://reviewable.io/reviews/hypar-io/elements/633#-MfJheLQDWBZIbVE4nSY:-MfJheLQDWBZIbVE4nSZ:b8kb4ni) ([raw file](https://github.com/hypar-io/elements/blob/a0a4fdcdc804324c52aacb9295ba9fb41acbffa4/Elements/src/RandomExtensions.cs#L32)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Generate a 
> ```

Generate a what? :stuck_out_tongue: 
___
*[Elements/src/SetOperations.cs, line 37 at r2](https://reviewable.io/reviews/hypar-io/elements/633#-MfSiD3JD4_MJm5nvcbo:-MfSiD3JD4_MJm5nvcbp:b-kh4ccx) ([raw file](https://github.com/hypar-io/elements/blob/35db6667ffa4d8a8c81fb6a95366673db5c53e69/Elements/src/SetOperations.cs#L37)):*
> ```Smalltalk
>     {
>         /// <summary>
>         /// Classify a segment against a polygon by shooting a random
> ```

This description makes it sound like the method takes a single segment.
___
*[Elements/src/SetOperations.cs, line 73 at r2](https://reviewable.io/reviews/hypar-io/elements/633#-MfShPt01Zkw9ZeVOiNG:-MfShPt01Zkw9ZeVOiNH:b2ychir) ([raw file](https://github.com/hypar-io/elements/blob/35db6667ffa4d8a8c81fb6a95366673db5c53e69/Elements/src/SetOperations.cs#L73)):*
> ```Smalltalk
>                 var v1 = a.Vertices[i];
>                 var v2 = i == a.Vertices.Count - 1 ? a.Vertices[0] : a.Vertices[i + 1];
>                 var ray = r.NextRayInPlane(v1.Average(v2), p.Normal);
> ```

I know that this works in theory with any random vector, but the idea of a method's behavior potentially changing from one execution to the next makes me nervous. Wouldn't a constant vector suffice? (or rather, some constant vector V cross Normal, with a check to pick a different V if normal == V)
___
*[Elements/src/SetOperations.cs, line 128 at r2](https://reviewable.io/reviews/hypar-io/elements/633#-MfSffTc66cHgSFujGbR:-MfSffTc66cHgSFujGbS:bp82w1c) ([raw file](https://github.com/hypar-io/elements/blob/35db6667ffa4d8a8c81fb6a95366673db5c53e69/Elements/src/SetOperations.cs#L128)):*
> ```Smalltalk
>         {
>             var vertices = new List<Vector3>();
>             var epv = new List<List<(int from, int to, int? tag)>>();
> ```

can we call this variable edgesPerVertex for clarity?

andrewheumann (changes requested):
This can go into "future work" but it seems like there's a missing `AIntersectsB` classification case that would be useful. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/633#-:-MfTOZHM7kvRI-oDvJ9o:bs0b700)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/SetOperations.cs, line 37 at r2](https://reviewable.io/reviews/hypar-io/elements/633#-MfSiD3JD4_MJm5nvcbo:-MfTO9E64iWfW8rdiHom:b-vnft3m) ([raw file](https://github.com/hypar-io/elements/blob/35db6667ffa4d8a8c81fb6a95366673db5c53e69/Elements/src/SetOperations.cs#L37)):*

This description makes it sound like the method takes a single segment.
</blockquote></details>

The description should also probably reflect that the expectation is that segments of these polygons are already atomic / non-intersecting. This threw me for a loop when I was initially reading the code.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/633)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/SetOperations.cs, line 73 at r2](https://reviewable.io/reviews/hypar-io/elements/633#-MfShPt01Zkw9ZeVOiNG:-MfTRW2YAyDaHuSZICae:bnnr16f) ([raw file](https://github.com/hypar-io/elements/blob/35db6667ffa4d8a8c81fb6a95366673db5c53e69/Elements/src/SetOperations.cs#L73)):*

I know that this works in theory with any random vector, but the idea of a method's behavior potentially changing from one execution to the next makes me nervous. Wouldn't a constant vector suffice? (or rather, some constant vector V cross Normal, with a check to pick a different V if normal == V)
</blockquote></details>

you could even use the segment's own direction, I think — it's guaranteed to be in the plane. or maybe segment's direction cross normal, if we're worried the segment's own direction would lead to too many weird edge cases.

PR #643, opened 2021-08-11
Add a texture lookup for repeated strings and text direction.

ikeough:
BACKGROUND:
When using `ModelText` in Hypar, I noticed that drawing times were quite slow for a model with many model text objects with the same text and that text could not align with a curve.

DESCRIPTION:
This PR adds a texture lookup keyed on a hash of the text string and the text color, the two properties which control the appearance of the generated texture. This enables us to lookup an existing instance of a label with the correct text and color in the atlas and use that instead of writing a new label to the file. 

This PR adds a new `lineDirection` to the text data tuple, which is used to direct the line of text.

TESTING:

After editing the model text tests to output text labels with the same string and **different colors**:

image

After editing the model text tests to output text labels with the same string and **same color**:

image

Text can now be oriented in the direction of a curve, as in these grid heads:
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/643)


serenayl (approved):
It's beautiful, I love this.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/643#-:-MhFnlCgCFlBBD0idHUp:bynttm)* status: :shipit: complete! 1 of 1 approvals obtained

PR #645, opened 2021-08-13
Solid plane intersection

ikeough:
BACKGROUND:
While working on allowable envelopes, I was hacking around the fact that we didn't have solid intersection with planes, by creating a really big polygon and clipping it with all the faces of a solid. This works, but will inevitably run into the case where the "big" plane isn't big enough, and the method requires bi-directional intersection of polygons. Having a solid/plane intersection method removes this hack and improves performance.

DESCRIPTION:
This PR adds the `Solid.Intersects(...)` method, which enables intersection of solids with a plane resulting in zero->many polygons.

TESTING:
- See `SolidTests.SolidIntersectsWithPlane()`
  
FUTURE WORK:
- Like all of our solids work, this does not support anything other than planar faces.

REQUIRED:

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)


andrewheumann (changes requested):
In general this works extremely well. I think there's at least one edge case we should capture up-front (where the plane is coplanar with a face) and we should also try to fail more gracefully if we hit an exception when intersecting — rather than a user of the method needing to wrap it in a try/catch, it would be good if it could return partial results if it does run into trouble at some step.

I've created a branch called `solid-plane-intersection-tests` that includes some failing cases that I found experimentally. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645#-:-MgzvcsyEg831GnQcl36:b-3e0aep)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Solids/Solid.cs, line 567 at r1](https://reviewable.io/reviews/hypar-io/elements/645#-MgzkzsUC3j0KQ75dIlX:-MgzkzsUC3j0KQ75dIlY:b-y55o0r) ([raw file](https://github.com/hypar-io/elements/blob/a2d7eaf31e53fc5829a63c5ea8b9791d317b2113/Elements/src/Geometry/Solids/Solid.cs#L567)):*
> ```Smalltalk
>                 }
> 
>                 var d = p.Normal.Cross(f.Value.Plane().Normal).Unitized();
> ```

We operate on the assumption here that we're not intersecting a face parallel to the plane. We should maybe try to handle this case — either by assuming this is never an intersection, or (my preference) if the plane and the face are coplanar, to return the entire face's boundary. This will come up in cases where we're taking levels through an envelope and we start at z=0 — it would be nice if the bottom face of the envelope was returned. Currently I believe this throws an exception, which should be avoided. 
___
*[Elements/src/Geometry/Solids/Solid.cs, line 597 at r1](https://reviewable.io/reviews/hypar-io/elements/645#-MgzjR2i1IHyJwaQ0QeK:-MgzjR2jDkVebTFF-wu_:b-oznkuh) ([raw file](https://github.com/hypar-io/elements/blob/a2d7eaf31e53fc5829a63c5ea8b9791d317b2113/Elements/src/Geometry/Solids/Solid.cs#L597)):*
> ```Smalltalk
>                     // var e2 = (b, a, 0);
>                     if (!graphEdges[a].Contains(e1)) graphEdges[a].Add(e1);
>                     // if (!graphEdges[b].Contains(e2)) graphEdges[b].Add(e2);
> ```

nix commented code

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Solids/Solid.cs, line 567 at r1](https://reviewable.io/reviews/hypar-io/elements/645#-MgzkzsUC3j0KQ75dIlX:-Mh02GC_7maWx7d0q8Hh:blfp2wl) ([raw file](https://github.com/hypar-io/elements/blob/a2d7eaf31e53fc5829a63c5ea8b9791d317b2113/Elements/src/Geometry/Solids/Solid.cs#L567)):*

We operate on the assumption here that we're not intersecting a face parallel to the plane. We should maybe try to handle this case — either by assuming this is never an intersection, or (my preference) if the plane and the face are coplanar, to return the entire face's boundary. This will come up in cases where we're taking levels through an envelope and we start at z=0 — it would be nice if the bottom face of the envelope was returned. Currently I believe this throws an exception, which should be avoided. 
</blockquote></details>

With the change to fix this, the entire plane's boundary will be returned, but it will be returned as the result of intersection with the adjacent faces, because we skip co-planar faces.
___
*[Elements/src/Geometry/Solids/Solid.cs, line 597 at r1](https://reviewable.io/reviews/hypar-io/elements/645#-MgzjR2i1IHyJwaQ0QeK:-Mh02PSx9AvPThlAPW0i:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/a2d7eaf31e53fc5829a63c5ea8b9791d317b2113/Elements/src/Geometry/Solids/Solid.cs#L597)):*

nix commented code
</blockquote></details>

Done.

andrewheumann (approved):
> Reviewed 2 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (changes requested):
There's another case that's failing unexpectedly. The plane is not coplanar with any face, edge, or vertex of the solid: 
```
{"Solid":[{"type":"elements.geometry.solids.solid","vertices":[{"id":100000,"x":42.246505773750755,"y":52.169272887596,"z":0.0},{"id":100001,"x":59.52485890416793,"y":52.169272887596,"z":0.0},{"id":100002,"x":55.786116385571184,"y":58.68108438034021,"z":0.0},{"id":100003,"x":68.82209518950515,"y":57.99459940731279,"z":0.0},{"id":100004,"x":62.26077139722092,"y":52.169272887596,"z":0.0},{"id":100005,"x":72.15124624123484,"y":53.3120127779968,"z":0.0},{"id":100006,"x":72.15124624123484,"y":7.637547078464806,"z":0.0},{"id":100007,"x":42.246505773750755,"y":7.637547078464806,"z":0.0},{"id":100008,"x":42.246505773750755,"y":52.169272887596,"z":10.0},{"id":100009,"x":59.52485890416793,"y":52.169272887596,"z":10.0},{"id":100010,"x":55.786116385571184,"y":58.68108438034021,"z":10.0},{"id":100011,"x":68.82209518950515,"y":57.99459940731279,"z":10.0},{"id":100012,"x":62.26077139722092,"y":52.169272887596,"z":10.0},{"id":100013,"x":72.15124624123484,"y":53.3120127779968,"z":10.0},{"id":100014,"x":72.15124624123484,"y":7.637547078464806,"z":10.0},{"id":100015,"x":42.246505773750755,"y":7.637547078464806,"z":10.0}],"faces":[{"id":0,"outer":[{"vertex_id":100000,"edge_id":10000,"edge_side":"left"},{"vertex_id":100001,"edge_id":10001,"edge_side":"left"},{"vertex_id":100002,"edge_id":10002,"edge_side":"left"},{"vertex_id":100003,"edge_id":10003,"edge_side":"left"},{"vertex_id":100004,"edge_id":10004,"edge_side":"left"},{"vertex_id":100005,"edge_id":10005,"edge_side":"left"},{"vertex_id":100006,"edge_id":10006,"edge_side":"left"},{"vertex_id":100007,"edge_id":10007,"edge_side":"left"}],"inner":[]},{"id":1,"outer":[{"vertex_id":100001,"edge_id":10000,"edge_side":"right"},{"vertex_id":100000,"edge_id":10008,"edge_side":"left"},{"vertex_id":100008,"edge_id":10016,"edge_side":"left"},{"vertex_id":100009,"edge_id":10009,"edge_side":"right"}]},{"id":2,"outer":[{"vertex_id":100002,"edge_id":10001,"edge_side":"right"},{"vertex_id":100001,"edge_id":10009,"edge_side":"left"},{"vertex_id":100009,"edge_id":10017,"edge_side":"left"},{"vertex_id":100010,"edge_id":10010,"edge_side":"right"}]},{"id":3,"outer":[{"vertex_id":100003,"edge_id":10002,"edge_side":"right"},{"vertex_id":100002,"edge_id":10010,"edge_side":"left"},{"vertex_id":100010,"edge_id":10018,"edge_side":"left"},{"vertex_id":100011,"edge_id":10011,"edge_side":"right"}]},{"id":4,"outer":[{"vertex_id":100004,"edge_id":10003,"edge_side":"right"},{"vertex_id":100003,"edge_id":10011,"edge_side":"left"},{"vertex_id":100011,"edge_id":10019,"edge_side":"left"},{"vertex_id":100012,"edge_id":10012,"edge_side":"right"}]},{"id":5,"outer":[{"vertex_id":100005,"edge_id":10004,"edge_side":"right"},{"vertex_id":100004,"edge_id":10012,"edge_side":"left"},{"vertex_id":100012,"edge_id":10020,"edge_side":"left"},{"vertex_id":100013,"edge_id":10013,"edge_side":"right"}]},{"id":6,"outer":[{"vertex_id":100006,"edge_id":10005,"edge_side":"right"},{"vertex_id":100005,"edge_id":10013,"edge_side":"left"},{"vertex_id":100013,"edge_id":10021,"edge_side":"left"},{"vertex_id":100014,"edge_id":10014,"edge_side":"right"}]},{"id":7,"outer":[{"vertex_id":100007,"edge_id":10006,"edge_side":"right"},{"vertex_id":100006,"edge_id":10014,"edge_side":"left"},{"vertex_id":100014,"edge_id":10022,"edge_side":"left"},{"vertex_id":100015,"edge_id":10015,"edge_side":"right"}]},{"id":8,"outer":[{"vertex_id":100000,"edge_id":10007,"edge_side":"right"},{"vertex_id":100007,"edge_id":10015,"edge_side":"left"},{"vertex_id":100015,"edge_id":10023,"edge_side":"left"},{"vertex_id":100008,"edge_id":10008,"edge_side":"right"}]},{"id":9,"outer":[{"vertex_id":100008,"edge_id":10023,"edge_side":"right"},{"vertex_id":100015,"edge_id":10022,"edge_side":"right"},{"vertex_id":100014,"edge_id":10021,"edge_side":"right"},{"vertex_id":100013,"edge_id":10020,"edge_side":"right"},{"vertex_id":100012,"edge_id":10019,"edge_side":"right"},{"vertex_id":100011,"edge_id":10018,"edge_side":"right"},{"vertex_id":100010,"edge_id":10017,"edge_side":"right"},{"vertex_id":100009,"edge_id":10016,"edge_side":"right"}],"inner":[]}]}],"Plane":{"Origin":{"X":73.3279549937009,"Y":23.674585366059862,"Z":1.0},"Normal":{"X":-27.56613411658539,"Y":28.49468752153614,"Z":0.0}},"Exception":"Object reference not set to an instance of an object."}
```


> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645#-:-MhAHXHh1woOLwRkH4g-:ba11el7)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann (changes requested):
If it helps, this is the graph that's being constructed: 

the intersection with the bottom face is pointing a different way than all the other edges. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645#-:-MhALEVHDW0IGC5IfGv7:bjtyday)* status: 1 change requests, 0 of 1 approvals obtained

ikeough:
@andrewheumann Does the half edge graph require that edges orient into consistently wound loops?

andrewheumann:
Yes, it requires that either you have two half edges per edge (in which case you always have complete winding in both directions for the "inside" and "outside" loops) or that you make sure the edges are wound consistently.

ikeough:
Yes, noticing that this creates a graph where one vertex has no edges assigned, which always fails in the HEG.

ikeough (commented):
The provided test case was unexpectedly great. The winding error that you recognized was because our normal calculation for faces in XY was wrong for faces with a normal of -Z, causing the plane intersection points for the face to get sorted across the face incorrectly. This made me realize that our `Plane.IsCoplanar()` method was wrong as well (in addition to being overly complicated). Cleaning up those two things gives us predictable intersection winding and clean half edge graphs.

The provided test was added to the test suite.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645#-:-MhCKpP34RgwIi3FQ0gP:b-svzxyw)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann (changes requested):
> Reviewed 1 of 1 files at r4, 7 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Plane.cs, line 96 at r5](https://reviewable.io/reviews/hypar-io/elements/645#-MhE-ymx0AZhmJuR5WOo:-MhE-ymx0AZhmJuR5WOp:blncsy3) ([raw file](https://github.com/hypar-io/elements/blob/e288ac73b7371e569b427957a9bacd476491bfb6/Elements/src/Geometry/Plane.cs#L96)):*
> ```Smalltalk
>         public bool IsCoplanar(Plane plane)
>         {
>             return this.Normal.Cross(plane.Normal).IsZero();
> ```

This is no longer a sufficient test for coplanarity — It only tests parallel-ness, not position in space. 
___
*[Elements/src/Geometry/Vector3.cs, line 1029 at r5](https://reviewable.io/reviews/hypar-io/elements/645#-MhE2YjP0RWO3Cj6Csbz:-MhE2YjS39-bNFPgPqWq:b-d8kk5r) ([raw file](https://github.com/hypar-io/elements/blob/e288ac73b7371e569b427957a9bacd476491bfb6/Elements/src/Geometry/Vector3.cs#L1029)):*
> ```Smalltalk
>                 var p0 = points[i];
>                 var p1 = points[(i + 1) % points.Count];
>                 normal.X += (p0.Y - p1.Y) * (p0.Z + p1.Z);
> ```

Moving this out of polygon, we should make it explicit in the comments that this function is dependent on the order of the points supplied. Moreover, it will return a result even for non-planar sets of points, which may not be meaningful. I know we're keeping it internal, but even for our own use we should make it clear when it's appropriate to use this method.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Plane.cs, line 96 at r5](https://reviewable.io/reviews/hypar-io/elements/645#-MhE-ymx0AZhmJuR5WOo:-MhEpFsd0z0MDzlLrwV1:bt6gqr3) ([raw file](https://github.com/hypar-io/elements/blob/e288ac73b7371e569b427957a9bacd476491bfb6/Elements/src/Geometry/Plane.cs#L96)):*

This is no longer a sufficient test for coplanarity — It only tests parallel-ness, not position in space. 
</blockquote></details>

The old logic has been replaced in addition to the unitization of the plane vectors during deconstruction.
___
*[Elements/src/Geometry/Vector3.cs, line 1029 at r5](https://reviewable.io/reviews/hypar-io/elements/645#-MhE2YjP0RWO3Cj6Csbz:-MhEorGm1pY5bHkPf_Hr:bp9yqe8) ([raw file](https://github.com/hypar-io/elements/blob/e288ac73b7371e569b427957a9bacd476491bfb6/Elements/src/Geometry/Vector3.cs#L1029)):*

Moving this out of polygon, we should make it explicit in the comments that this function is dependent on the order of the points supplied. Moreover, it will return a result even for non-planar sets of points, which may not be meaningful. I know we're keeping it internal, but even for our own use we should make it clear when it's appropriate to use this method. 
</blockquote></details>

The name of the method is updated to reflect what is required.
___
*[Elements/src/Validators/Validators.cs, line 212 at r6](https://reviewable.io/reviews/hypar-io/elements/645#-MhEoyEo0htTISSga1U_:-MhEoyEo0htTISSga1Ua:b-5s2lko) ([raw file](https://github.com/hypar-io/elements/blob/6a0c96c3a94ca092d51c7f8a3e6565f7e502896a/Elements/src/Validators/Validators.cs#L212)):*
> ```Smalltalk
>         {
>             var p = (Plane)obj;
>             if (!p.Normal.IsUnitized())
> ```

This was causing the non co-planarity error. Planes provided in the sample JSONs didn't have unit length normals. I think it's a safe assumption that we want to make planes have unit length normals.

andrewheumann (approved):
> Reviewed 6 of 7 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/645)* status: :shipit: complete! 1 of 1 approvals obtained

PR #646, opened 2021-08-14
Add Transform.Inverse()

andrewheumann:
BACKGROUND:
- For a great many of our fluent-ish APIs, we supply both `.verb()` (e.g. `Move()`) methods to modify a thing in place, and `.participle-adjective()` (e.g. `Moved()`) to return a new version of the thing with a change applied. Transform had `.Invert()` but not `.Inverse()`.
DESCRIPTION:
- Adds `Transform.Inverse()`

TESTING:
- Added a `Transform_Inverse` test to `TransformTests`.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/646)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@
 
 - `ModelArrows`
 - `ModelText`
+- `Transform.Inverse()`
+
 ### Changed
 
 ### Fixed
+
 - Deduplicate catalog names during code generation.
 
 ## 0.9.2
@@ -43,7 +46,6 @@
   - `Line.ExtendTo(Polygon)`
   - `Line.ExtendTo(Profile)`
 - Support for DXF from many basic elements.
-
 
 ### Changed
 
--- a/Elements/src/Geometry/Transform.cs
+++ b/Elements/src/Geometry/Transform.cs
@@ -308,6 +308,14 @@ namespace Elements.Geometry
         }
 
         /// <summary>
+        /// Return a new transform which is the inverse of this transform.
+        /// </summary>
+        public Transform Inverse()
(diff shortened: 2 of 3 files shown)


andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/646)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Geometry/Transform.cs, line 313 at r1](https://reviewable.io/reviews/hypar-io/elements/646#-MhE9GbD0czFhZqE7CaV:-MhEjjQL8sNBGIjb7eNs:b6zwt45) ([raw file](https://github.com/hypar-io/elements/blob/6683c02536ae7b960decf73745f82fc89e8f8d7a/Elements/src/Geometry/Transform.cs#L313)):*

seems like we use the past tense a lot for these, should this be `Inverted()` ?
</blockquote></details>

I see the argument about Inverted/Inverse, but I think I still prefer Inverse. it's much more common to talk about the Inverse Transform than the "Inverted" transform. The intention behind the "Moved" etc methods is that it's clear that the thing you're getting back has had X done to it — "Give me the moved object" as opposed to "move" which operates on the object in-place "Move this object". I think Invert/Inverse still satisfies this: "This is the inverse transform" vs "invert this transform"
We also already have the Matrix.Inverse() and it would be weird to use two terms for the same thing

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/646)* status: :shipit: complete! 1 of 1 approvals obtained

PR #653, opened 2021-08-19
version retrieval accoring to semver (ignoreing v*)

wynged:
I've found a more reliable way of retrieving the latest version tag as long as we keep doing semver.  Had to exclude 'v' prefix since we don't use that lately.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/653)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/653)* status: :shipit: complete! 1 of 1 approvals obtained

PR #655, opened 2021-08-22
Binary tree and adjacency graph.

ikeough:
BACKGROUND:
- For building elements connected as a graph in a plane, like wall segments, we need a way of building a traversable graph of discrete edge sections.

DESCRIPTION:
This PR adds the following classes:
`Elements.Search.BinaryTree` - A binary tree which allows the user to provide a custom comparer which the tree uses during element addition.
`Elements.Search.AdjacencyList` - A minimal adjacency list implementation which describes a graph as a set of pointers to connected node indices.

This PR adds one method `IList<Line>.Intersections`, which calculates the intersections of an arbitrary set of lines returning the intersection locations and the adjacency graph describing the line connectivity.

`208ms for finding 1581 intersections.`
image

TESTING:
- New tests are added in `LineSweepTests`.
  
FUTURE WORK:
- This implementation is not as fast as it could be for two reasons:
  - For line sets which include lines with coincident start or end points, we need to do a two pass event creation, then group events by the event's X coordinate and store multiple edges per event. This is O(n).
  - We are avoiding duplicate intersection result creation by using `pts.IndexOf(somePt)` to find an existing point. This is also O(n). This could be improved by using a more efficient lookup for existing intersections, like an octree, but the lookup would have to also have a small insertion time to make a good tradeoff.
- The `AdjacencyList` class is very similar to the `HalfEdgeGraph` created by @andrewheumann with the benefit of being smaller (edges are stored as one  `int` pointer on the target's collection instead of a two `Vector3` tuple). Additionally the `AdjacencyList` class is generic, it stores a graph as indices pointing to a separate collection, so any type of connected data can be referenced.

REQUIRED:

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/655)


Elements/src/Geometry/Vector3.cs
```diff
@@ -842,218 +842,4 @@ internal static List<Vector3> AttemptPostClipperCleanup(IList<Vector3> vertices)
             return new Vector3(vector.X, vector.Y);
         }
     }
-
-    /// <summary>
-    /// Extension methods for Vector3.
-    /// </summary>
-    public static class Vector3Extensions
```
ikeough, line 849: This class was moved to its own file.

serenayl (changes requested):
> Reviewed 5 of 14 files at r1, 1 of 2 files at r2, 4 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/655)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[CHANGELOG.md, line 7 at r3](https://reviewable.io/reviews/hypar-io/elements/655#-MiT_Ixm4IY49PQOT_IS:-MiT_Ixm4IY49PQOT_IT:b-lnvopx) ([raw file](https://github.com/hypar-io/elements/blob/6bba99ab57ec7310cdfae2535d48f11ff083ed0a/CHANGELOG.md#L7)):*
> ```Markdown
> - `Elements.Search.BinaryTree<T>`
> - `Elements.Search.BinaryTreeNode<T>`
> ```

This is missing the AdjacencyList?
___
*[Elements/src/Search/BinaryTree.cs, line 203 at r3](https://reviewable.io/reviews/hypar-io/elements/655#-MiTUk2583eUC_iYYZl4:-MiTUk2583eUC_iYYZl5:be0ec0r) ([raw file](https://github.com/hypar-io/elements/blob/6bba99ab57ec7310cdfae2535d48f11ff083ed0a/Elements/src/Search/BinaryTree.cs#L203)):*
> ```
> parent.Right
> ```
For my own reference, this is just an arbitrary decision at this point, right? It could just as easily be left.
___
*[Elements/src/Search/BinaryTree.cs, line 225 at r3](https://reviewable.io/reviews/hypar-io/elements/655#-MiTZZtZBCONz6SYwP7W:-MiTZZt_8wlsc5wDxi6v:bf94odu) ([raw file](https://github.com/hypar-io/elements/blob/6bba99ab57ec7310cdfae2535d48f11ff083ed0a/Elements/src/Search/BinaryTree.cs#L225)):*
> ```
> if (parent != null)
> ```
Non-blocking style thing, I have a tendency to prefer early returns to keep nesting more minimal. Happy to hear other approaches too!

```
if(parent == null) { return null; }
// one less deeply nested logic block
```
___
*[Elements/test/BinaryTreeTests.cs, line 90 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiT_fLa8sdl1ZwwohKz:-MiT_fLbDWlpZLvfjRqT:b-vlnzd8) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/test/BinaryTreeTests.cs#L90)):*
> ```
> int
> ```
For my own education, why do we use a tree of indices rather than lines as the data directly?
___
*[Elements/test/LineSweepTests.cs, line 45 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiTbvfe0gyM9rL8fVhY:-MiTbvfe0gyM9rL8fVhZ:b-dzc7w2) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/test/LineSweepTests.cs#L45)):*
> ```
> with coincident left-most points
> ```
I think this is just one vertical and one horizontal line that intersect at the origin?

serenayl (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/655)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/LineSegmentExtensions.cs, line 44 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiTj0CG8TFst95ovSUJ:-MiTj0CHE36beccf82cY:b-quh9m5) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Geometry/LineSegmentExtensions.cs#L44)):*
> ```
> Find all intersections of the provided collection of lines.
> ```
I think unintuitively for me, this also includes endpoints that are not intersections but that are simply endpoints of the existing lines, if I understand this correctly? This makes sense for creating a complete adjacency, but feels like not the behavior I would anticipate as a user. Is there a way to make that clear in the description?

serenayl (changes requested):
> Reviewed 4 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/655)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/LineSegmentExtensions.cs, line 44 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiTj0CG8TFst95ovSUJ:-MiWhODk8yB9e66BG7Oe:bdkf09o) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Geometry/LineSegmentExtensions.cs#L44)):*

> ```
> Find all intersections of the provided collection of lines.
> ```
I think unintuitively for me, this also includes endpoints that are not intersections but that are simply endpoints of the existing lines, if I understand this correctly? This makes sense for creating a complete adjacency, but feels like not the behavior I would anticipate as a user. Is there a way to make that clear in the description?
</blockquote></details>

After looking through file some more, it feels like `Intersections` should be its own function that takes in as an option whether we want to include the endpoints of the bag of lines, and then we want another function "AdjacencyList" that calls to `Intersections` first with the specification to include endpoints, and then does its own thing and optionally has the out param as the list of intersections.
___
*[Elements/src/Geometry/LineSegmentExtensions.cs, line 83 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiWetyOC_dnz7sGCxia:-MiWetyOC_dnz7sGCxib:b-7u06oi) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Geometry/LineSegmentExtensions.cs#L83)):*
> ```
> var segments = items.Select(item => { return getSegment(item); }).ToArray();
> ```
I think this is better for terseness, but we did also call `getSegment` on every item in the loop once already. Could we have built this array then?
___
*[Elements/src/Geometry/LineSegmentExtensions.cs, line 137 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiWgmGWE48-ZbaXO6Pz:-MiWgmGWE48-ZbaXO6Q-:b-fg49bi) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Geometry/LineSegmentExtensions.cs#L137)):*
> ```
> if (pre != null && suc != null)
> ```
To make sure I'm understanding, this happens if this is a vertical line?
___
*[Elements/src/Search/AdjacencyList.cs, line 10 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiWljNu2xOSOOzSn-4I:-MiWljNu2xOSOOzSn-4J:b-r56b4h) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Search/AdjacencyList.cs#L10)):*
> ```
> .
> ```
connected items/nodes?
___
*[Elements/src/Search/AdjacencyList.cs, line 24 at r4](https://reviewable.io/reviews/hypar-io/elements/655#-MiWmZix64-sHbXaDWXt:-MiWmZix64-sHbXaDWXu:b-3s4o6s) ([raw file](https://github.com/hypar-io/elements/blob/97d857fd71f2be02a8687adc118a30d0450979ee/Elements/src/Search/AdjacencyList.cs#L24)):*
> ```
> ++i
> ```
For consistency, why `++i` here instead of `i++`? We seem to only use `++i` syntax here and in the clipper code.

ikeough:
After discussion with @serenayl, I've added a `Network<T>` class which wraps `AdjacencyList` (which is now made private). The `Network<T>` class provides high level methods for adding nodes and edges to a network, as well as traversal methods which delegate traversal behavior to the caller. I've also made `BinaryTree<T>` private for now as it is an implementation detail and not yet generally useful in the larger Elements context.

serenayl (approved):
:lgtm:

> Reviewed 7 of 14 files at r5, 2 of 3 files at r6, 1 of 1 files at r7, 8 of 8 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/655#-:-Mifnhte02hi1G6dDyga:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough and @serenayl)

___
*[Elements/src/Search/LeftMostPointComparer.cs, line 19 at r8](https://reviewable.io/reviews/hypar-io/elements/655#-Mic-lyE6yaBUCUINj6S:-Mic-lyE6yaBUCUINj6T:bdx2x31) ([raw file](https://github.com/hypar-io/elements/blob/e506f24fa05946d25556d87c908b1521a48abc4a/Elements/src/Search/LeftMostPointComparer.cs#L19)):*
> ```
> xLeft
> ```
Non-blocking super-nit: can these be `aLeft` and `bLeft`?

PR #656, opened 2021-08-27
GlTF export performance improvements

ikeough:
During discussions of compiling Elements to wasm, we began to investigate sources of slowness in the gltf conversion path. There were a number of opportunities for small improvements, like reducing allocation of new polygons, and removing unnecessary creation of `Vector3` when converting from another vector type. 

- Removes the use of octree for merge vertex lookup during CSG output to graphic buffers.
- Removes use of `polygon.Transformed()` and `polygon.Reversed()` in `Solid`. Those methods both allocate new polygons which is not required as the transforms and winding can be handled while building face loops.
- Adds a couple of method to `Vector3` to allow dot and cross products to be calculated using components of an incoming vector.

Writing all HSS members to gltf before:

|                          Method |     Mean |   Error |  StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|--------:|--------:|-----------:|----------:|------:|----------:|
| 'Draw all beams without merge.' | 235.7 ms | 4.99 ms | 2.97 ms | 29000.0000 | 1000.0000 |     - |  95.18 MB |
|    'Draw all beams with merge.' | 232.3 ms | 2.69 ms | 1.60 ms | 30000.0000 | 3000.0000 |     - |  95.18 MB |

After:
|                          Method |      Mean |     Error |    StdDev |      Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|-------------------------------- |----------:|----------:|----------:|-----------:|----------:|----------:|----------:|
| 'Draw all beams without merge.' | 119.72 ms | 29.135 ms | 19.271 ms | 25000.0000 | 2000.0000 | 1000.0000 |  83.01 MB |
|    'Draw all beams with merge.' |  95.55 ms |  8.368 ms |  5.535 ms | 25000.0000 | 2000.0000 | 1000.0000 |  83.01 MB |

Observations:
- The "before" benchmarks are run against master and don't have the API which allows to merge vertices, so both numbers are similar.
- The "after" benchmarks are run with and without the new API to merge vertices, which shows a nearly 20ms speed up even with the merge. My assumption here is that the reduced number of vertices after merge helps to accelerate everything downstream.
 
Creating a beam with a varying number of holes before:
|           Method | NumberOfHoles |     Mean |    Error |   StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------- |-------------- |---------:|---------:|---------:|--------:|------:|------:|----------:|
| 'Tesselate CSG.' |             1 | 756.0 us | 459.1 us | 25.17 us | 81.0547 |     - |     - | 250.77 KB |
| 'Tesselate CSG.' |            10 | 811.7 us | 472.5 us | 25.90 us | 81.0547 |     - |     - | 250.77 KB |
| 'Tesselate CSG.' |            20 | 774.2 us | 168.2 us |  9.22 us | 81.0547 |     - |     - | 250.77 KB |
| 'Tesselate CSG.' |            50 | 769.9 us | 399.9 us | 21.92 us | 81.0547 |     - |     - | 250.77 KB |

After:
|           Method | NumberOfHoles |     Mean |     Error |  StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------- |-------------- |---------:|----------:|--------:|--------:|------:|------:|----------:|
| 'Tesselate CSG.' |             1 | 143.7 us |  59.86 us | 3.28 us | 55.1758 |     - |     - | 169.75 KB |
| 'Tesselate CSG.' |            10 | 149.3 us | 169.65 us | 9.30 us | 55.1758 |     - |     - | 169.75 KB |
| 'Tesselate CSG.' |            20 | 158.4 us |  38.93 us | 2.13 us | 55.1758 |     - |     - | 169.75 KB |
| 'Tesselate CSG.' |            50 | 137.9 us | 106.55 us | 5.84 us | 55.1758 |     - |     - | 169.75 KB |

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/656)

### DIFF
--- a/Elements.Benchmarks/Elements.Benchmarks.csproj
+++ b/Elements.Benchmarks/Elements.Benchmarks.csproj
@@ -10,6 +10,12 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Csg">
+      <HintPath>../Elements/lib/Csg.dll</HintPath>
+    </Reference>
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.12.1" />
   </ItemGroup>
 
--- a/Elements.Benchmarks/Program.cs
+++ b/Elements.Benchmarks/Program.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.IO;
 using System.Linq;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Running;
@@ -10,22 +9,23 @@ using Elements.Serialization.glTF;
 
 namespace Elements.Benchmarks
 {
+    [MemoryDiagnoser]
     [SimpleJob(launchCount: 1, warmupCount: 1, targetCount: 3)]
-    public class Csg
+    public class CsgBenchmarks
     {
         private WideFlangeProfileFactory _profileFactory = new WideFlangeProfileFactory();
+        private Beam _beam;
+        private Csg.Solid _csg;
 
-        [Params(1, 10, 20)]
-        public int Samples { get; set; }
+        [Params(1, 10, 20, 50)]
+        public int NumberOfHoles { get; set; }
 
-        [Benchmark(Description = "Compute csg of beam.")]
-        public void CSG()
+        public CsgBenchmarks()
         {
-            var profile = _profileFactory.GetProfileByType(WideFlangeProfileType.W10x100);
-
             var line = new Line(new Vector3(0, 0, 0), new Vector3(10, 0, 5));
-            var beam = new Beam(line, profile, BuiltInMaterials.Steel);
-            for (var i = 0.0; i <= 1.0; i += 1.0 / (double)Samples)
+            var profile = Polygon.Rectangle(Units.InchesToMeters(10), Units.InchesToMeters(20));
+            _beam = new Beam(line, profile, BuiltInMaterials.Steel);
+            for (var i = 0.0; i <= 1.0; i += 1.0 / (double)NumberOfHoles)
             {
                 var t = line.TransformAt(i);
                 var lt = new Transform(t.Origin, t.ZAxis, t.XAxis.Negate());
@@ -34,11 +34,15 @@ namespace Elements.Benchmarks
                 {
                     LocalTransform = lt
                 };
-                beam.Representation.SolidOperations.Add(hole);
+                _beam.Representation.SolidOperations.Add(hole);
             }
-            var model = new Model();
-            model.AddElement(beam);
-            GltfExtensions.InitializeGlTF(model, out var buffers, false);
+            _csg = _beam.GetFinalCsgFromSolids();
+        }
+
+        [Benchmark(Description = "Tesselate CSG.")]
+        public void CsgToGraphicsBuffers()
+        {
+            _csg.Tessellate();
         }
     }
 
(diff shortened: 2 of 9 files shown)


mattc (changes requested):
> Reviewed 1 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/656)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/CsgExtensions.cs, line 152 at r2](https://reviewable.io/reviews/hypar-io/elements/656#-MiT7pL0CX1YyVxWeJzB:-MiT7pL0CX1YyVxWeJzC:b-5djz9u) ([raw file](https://github.com/hypar-io/elements/blob/4ed7547ade1b87b05f0f103bad345936a561b12b/Elements/src/Geometry/CsgExtensions.cs#L152)):*
> ```Smalltalk
>         }
> 
>         private static int GetOrCreateVertex(double x, double y, double z, double nx, double ny, double nz, double u, double v,
> ```

Can we pass around each vertex as a tuple like we started to other places, rather than fully destructuring x/y/z?
___
*[Elements/src/Geometry/Solids/Solid.cs, line 773 at r2](https://reviewable.io/reviews/hypar-io/elements/656#-MiT83XeAdeWtLyL4DDP:-MiT83XeAdeWtLyL4DDQ:bgidkrq) ([raw file](https://github.com/hypar-io/elements/blob/4ed7547ade1b87b05f0f103bad345936a561b12b/Elements/src/Geometry/Solids/Solid.cs#L773)):*
> ```Smalltalk
>         }
> 
>         private void FindOrCreateVertex(Vector3 pt, int i, Transform transform, bool mergeVerticesAndEdges, Vertex[] verts)
> ```

What are the cases where this would be a Find rather than a Create? Is it not possible to know whether the point is expected to be found or not? Is this specifically for handling short edges or non-standard polygons?

Stray thought triggered here (not totally relevant): it seems like we should move more operations away from equality comparison on vertices and move towards operations on vertex IDs.

mattc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/656)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Solids/Solid.cs, line 773 at r2](https://reviewable.io/reviews/hypar-io/elements/656#-MiT83XeAdeWtLyL4DDP:-MicGD7k304UxKIfIw8E:biqezt4) ([raw file](https://github.com/hypar-io/elements/blob/4ed7547ade1b87b05f0f103bad345936a561b12b/Elements/src/Geometry/Solids/Solid.cs#L773)):*

The case for which this was originally created was that of adding a new face to a solid, knowing nothing about the original solid. We don't use this method when we created extrudes or sweeps because we know where all the vertices are. We do use this in the case of booleans where faces from a booleaned solid will need to be "attached" to existing faces.

Looking at this code for the first time in a while, it's obvious that we're doing "find" for every vertex of a face's loop every time we add a new face, which is unnecessary in the case of our canonical solid operations because we know where all the vertices are. I was so heavily thinking about things in terms of polygons at the time that I defaulted to a single `AddFace` method that is overkill, and not performant, for most cases. I'd recommend that we re-write this when we support curved surfaces because so much of what we're doing right now, with regards to splitting curves in flat facets, won't be required.
</blockquote></details>

Sounds good, could you capture the important parts of your reply as comments/TODOs?

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/656)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mattc)

___
*[Elements/src/Geometry/Solids/Solid.cs, line 773 at r2](https://reviewable.io/reviews/hypar-io/elements/656#-MiT83XeAdeWtLyL4DDP:-MigrP759UHGwZkR20FT:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/4ed7547ade1b87b05f0f103bad345936a561b12b/Elements/src/Geometry/Solids/Solid.cs#L773)):*

Sounds good, could you capture the important parts of your reply as comments/TODOs?
</blockquote></details>

Done.

mattc (approved):
> Reviewed 4 of 8 files at r1, 2 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/656)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mattc)

PR #659, opened 2021-09-01
Static method on Model to get elements of type within a model dict

serenayl:
BACKGROUND:
- This is a utility function used in the function execution queries work for cross-function overrides.
- People need to do this all the time in functions.

DESCRIPTION:
- Creates `Model.GetItemsOfType` to grab the relevant elements from a model dictionary.

TESTING:
- Try calling `Model.GetItemsOfType<YourTime>(inputModels)` in a function
  
FUTURE WORK:
- N/A

REQUIRED:

COMMENTS:
- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/659)


andrewheumann (changes requested):
I like this, but I think we should keep naming consistent — I'd still call it `AllElementsOfType`. 

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/659#-:-MiaEMFl4WrNJNiiLuyh:b-8393jv)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Model.cs, line 363 at r2](https://reviewable.io/reviews/hypar-io/elements/659#-MiaE3ZPBp9MnHqgsXBw:-MiaE3ZPBp9MnHqgsXBx:b3v8fim) ([raw file](https://github.com/hypar-io/elements/blob/39d53ea9e33c7c4e1c5891ecbb6405cb2a84d052/Elements/src/Model.cs#L363)):*
> ```Smalltalk
>         /// <typeparam name="T">The type of element we want to retrieve</typeparam>
>         public static List<T> GetItemsOfType<T>(Dictionary<string, Model> models, string modelName)
> ```

I would suggest making this an extension method by putting the "this" keyword in front of `Dictionary<string, Model>` — so you can just call 
`inputModels.AllElementsOfType<Envelope>("Envelope")`

To do this you might also have to move this method to its own static class — something like `ModelExtensions` which you might have to create. 

I'd also constrain `T`, so people don't go looking for non-element types in models:
```
 public static List<T> AllElementsOfType<T>(this Dictionary<string, Model> models, string modelName) where T : Element
{
...
}
```

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/659)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/659)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/Model.cs, line 364 at r4](https://reviewable.io/reviews/hypar-io/elements/659#-MifTTr40a7w07BcvdJK:-MifTTr40a7w07BcvdJL:br476xt) ([raw file](https://github.com/hypar-io/elements/blob/65dc60256c68a1196afd0a6c19f8e7a2f1dfb578/Elements/src/Model.cs#L364)):*
> ```Smalltalk
>         /// <param name="models">Dictionary of models keyed by string</param>
>         /// <param name="modelName">The name of the model</param>
>         /// <typeparam name="T">The type of element we want to retrieve</typeparam>
> ```

nit: periods at the end of these comments

PR #662, opened 2021-09-03
Remove validators and merge generated classes.

ikeough:
BACKGROUND:
During work on solid booleans, I noticed that every point, plane, polygon, etc. that we were constructing was hitting validation, which was doing a lookup in a type cache. This has a small effect locally, but for operations where many geometry are being created it has a large cumulative effect.

DESCRIPTION:
- This PR merges all generated classes into their partial classes and merges validation logic into constructors as necessary. It maintains the `Validator.DisableValidationOnConstruction` static property and wraps the newly merged validation logic in a check of that property. 
- Move csg creation out of solid operation construction. In many cases csg creation was being done multiple times: during initial object construction, and during subsequent calls to `UpdateRepresentation`. Csg creation only needs to be done _once_ during serialization to gltf. Now it is.
- **POSSIBLY CONTROVERSIAL** We add the `Polygon.Plane` and `Polygon.Normal` properties, set during construction, and adjust the `Polygon` class to use these properties instead of the methods to retrieve cached values for normals and planes.
- A trace project is added to the benchmarks which enables outputting a speedscope file for tracing the creation and serialization of all HSS types.
  - Tracing resulted in several micro-optimizations like the pre-sizing of arrays. 

|                Method | SkipValidation |      Mean |     Error |    StdDev |     Gen 0 | Allocated |
|---------------------- |--------------- |----------:|----------:|----------:|----------:|----------:|
| 'Draw all HSS beams.' |          False | 13.982 ms | 0.2696 ms | 0.6510 ms | 1000.0000 |      7 MB |
| 'Draw all HSS beams.' |           True |  8.452 ms | 0.8165 ms | 2.4075 ms | 1000.0000 |      5 MB |

OTHER WORK:
- [ ] Update CLI to use newest version of `Elements.CodeGeneration` which no longer generates validators in constructors.
 

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/662)


wynged (changes requested):
can we make deleting all of the json files a separate brand and PR so that we can revert quickly if we need to?  I suspect there is code in the wild that references these schemas, and having to deal with all of them being deleted all at once might be a headache.

> Reviewed 114 of 118 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662#-:-MjoRdoLBkTBwUu0zn20:b-lf8k8q)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[README.md, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoVMEP6FOLIsYPGL4J:-MjoVMEQ6h-a7CgKoDTi:bp3m86) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/README.md#L51)):*
> ```Markdown
>   - Elements have a unique identifier and a name.
> - Elements is a C# library presently, but we expect that Element types will be used in other languages in the future. Therefore, we shouldn't rely on capabilities of C# (ex: attributes) to convey meaning of the types or their properties. 
> - The core Element types will be defined in exactly the same way that third-party types will be defined. 
> ```

also delete this line.
Can we include some notes about how user generated types are still possible and widely supported on the platform?  A lot of the ## Code Generation comments below still apply.
___
*[Elements/src/Element.cs, line 37 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjogVX78RE02GeguMl9:-MjogVX78RE02GeguMlA:b5mdgim) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Element.cs#L37)):*
> ```Smalltalk
>         {
>             get { return _id; }
>             set
> ```

This seems new... but even if not, this seems like a good time to change.  I don't think element ids should be mutable.
___
*[Elements/src/Model.cs, line 64 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjojMwb53Dy5Hduz5s9:-MjojMwb53Dy5Hduz5sA:b25o8d2) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Model.cs#L64)):*
> ```Smalltalk
>         /// <param name="elements">The model's elements.</param>
>         public Model(Transform @transform, System.Collections.Generic.IDictionary<Guid, Element> @elements)
>         {
> ```

why did this validation just go away entirely?
___
*[Elements/src/Geometry/BBox3.cs, line 10 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjokYew2wbHLvwpLxwi:-MjokYew2wbHLvwpLxwj:b9tn7i0) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/BBox3.cs#L10)):*
> ```Smalltalk
>     /// An axis-aligned bounding box.
>     /// </summary>
>     [Newtonsoft.Json.JsonConverter(typeof(Elements.Serialization.JSON.JsonInheritanceConverter), "discriminator")]
> ```

other classes don't have this JsonConverter declaration, why is it needed here?
___
*[Elements/src/Geometry/CsgExtensions.cs, line 177 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjorKTACiso4eV_iJHe:-MjorKTACiso4eV_iJHf:b8iowg) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/CsgExtensions.cs#L177)):*
> ```Smalltalk
>             });
>         }
> ```

I presume this method got moved because it is actually an extension on Solid, but that's true of the two methods below as well, maybe we should move them too and remove this file?
___
*[Elements/src/Geometry/Curve.cs, line 9 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjomEGM5Xd8W-xCEl56:-MjomEGM5Xd8W-xCEl57:b-gbbzul) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Curve.cs#L9)):*
> ```Smalltalk
>     /// The abstract base class for all curves.
>     /// </summary>
>     [Newtonsoft.Json.JsonConverter(typeof(Elements.Serialization.JSON.JsonInheritanceConverter), "discriminator")]
> ```

Does this need a special call to the JsonInheritanceConverter as well?
___
*[Elements/src/Geometry/Profile.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjofhtX0oUpc1m4P-9t:-MjofhtX0oUpc1m4P-9u:bswc394) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Profile.cs#L12)):*
> ```Smalltalk
>     /// A polygonal perimeter with zero or more polygonal voids.
>     /// </summary>
>     public class Profile : Element, IEquatable<Profile>
> ```

do we actually want to remove the `partial`?  It's not removed for Polygon for example, and I don't know that it should be.
___
*[Elements/src/Geometry/Solids/SolidOperation.cs, line 8 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-Mjor6nqA0Osa9pNtyWH:-Mjor6nqA0Osa9pNtyWI:bwzg5dn) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Solids/SolidOperation.cs#L8)):*
> ```Smalltalk
>     /// The base class for all operations which create solids.
>     /// </summary>
>     [Newtonsoft.Json.JsonConverter(typeof(Elements.Serialization.JSON.JsonInheritanceConverter), "discriminator")]
> ```

this also has that inheritance converter.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 76 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoS4K5EWdvsqgNGad_:-MjoS4K6ArKMClLQZ2Cv:b-awvpsl) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L76)):*
> ```Smalltalk
> 
>             failedAssemblyErrors = new List<string>();
>             foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies().Where(a =>
> ```

can we make this check based on an Array of these prefixes rather than the multiple && logic
___
*[Elements/src/Validators/Validator.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoojPY14HbbI5COFRW:-MjoojPY14HbbI5COFRX:bkaxyyu) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Validators/Validator.cs#L12)):*
> ```Smalltalk
>         /// Note: Disabling validation can have unforeseen consequences. Use with caution.
>         /// </summary>
>         public static bool DisableValidationOnConstruction { get; set; } = false;
> ```

Removing the `Instance.GetFirstValidatorForType<T>()` entirely means this version of elements is incompatible with the Hypar.CLI.  can we leave that path in place, and mark it obsolete with a suggestion to upgrade your CLI, so that we can have this in Elements w/out breaking all developers who migrate to it?
___
*[Elements.Benchmarks/Program.cs, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-Mjoq6tS3ZHgjsj7dzB7:-Mjoq6tTDK2torWSVKOo:b-xd5zsx) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements.Benchmarks/Program.cs#L51)):*
> ```Smalltalk
>     [EventPipeProfiler(EventPipeProfile.CpuSampling)]
>     [SimpleJob(launchCount: 1, warmupCount: 0, targetCount: 1)]
>     public class Trace
> ```

seems like each of these classes and the `Program` entry point should be in their own file.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[README.md, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoVMEP6FOLIsYPGL4J:-MjzAdCf698yCajQ3H3S:bcfkhas) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/README.md#L51)):*

also delete this line.
Can we include some notes about how user generated types are still possible and widely supported on the platform?  A lot of the ## Code Generation comments below still apply.
</blockquote></details>

What platform ? ;) If we want to provide documentation for code generation from types because we think it's important for developers to know in relation to Hypar, we should provide that documentation in the Hypar docs.
___
*[Elements/src/Element.cs, line 37 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjogVX78RE02GeguMl9:-MjzFfXw5IQPLoeUekl7:b-1q5598) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Element.cs#L37)):*

This seems new... but even if not, this seems like a good time to change.  I don't think element ids should be mutable.
</blockquote></details>

It's not new. See https://github.com/hypar-io/Elements/blob/master/Elements/src/Generate/Element.g.cs#L55. Although I agree that Id mutation is probably bad, it's outside the scope of this PR to change it.
___
*[Elements/src/Model.cs, line 64 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjojMwb53Dy5Hduz5s9:-Mjz2Kte0gYwgI6iMlDV:bxljfm2) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Model.cs#L64)):*

why did this validation just go away entirely?
</blockquote></details>

Because it didn't exist. Many types didn't have validators, but all types were generated as if they might. Where no validator logic was present for a type, no validation was carried over.
___
*[Elements/src/Geometry/BBox3.cs, line 10 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjokYew2wbHLvwpLxwi:-Mjz2n67DHIXwEtaM_eD:bpd04zn) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/BBox3.cs#L10)):*

other classes don't have this JsonConverter declaration, why is it needed here?
</blockquote></details>

Classes which inherit from Element don't have it because the base class already has it. Classes which don't inherit from element, like `BBox3` have it by default because of the code generation template. I don't think it's required in this case, but I carried it over, where it existed already in the generated classes, for consistency.
___
*[Elements/src/Geometry/CsgExtensions.cs, line 177 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjorKTACiso4eV_iJHe:-Mjz3XND1s4o76FgDB1F:b-ofs5vi) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/CsgExtensions.cs#L177)):*

I presume this method got moved because it is actually an extension on Solid, but that's true of the two methods below as well, maybe we should move them too and remove this file?
</blockquote></details>

`CsgExtensions` has a number of extensions methods in it that, as the name suggests, extend Csg. We don't own Csg, so we need an extension class to bolt on methods. It would be a fair argument to say that we should move `ToMesh(this Mesh mesh)` to a mesh extensions class if our logic is that other type extensions should not live here. The solid method was moved only because I noticed it as part of one performance tweak in this PR _and_ we already had a solid extensions class.
___
*[Elements/src/Geometry/Curve.cs, line 9 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjomEGM5Xd8W-xCEl56:-Mjz4GUI1YysOHY2sbcX:btc2pke) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Curve.cs#L9)):*

Does this need a special call to the JsonInheritanceConverter as well?
</blockquote></details>

This class is an abstract base class for all of our curve types. Any function with a property of type `Curve` will make use of this when the actual curve type in the JSON needs to be converted to a concrete class. So yes, in this case this is required, and isn't just boiler plate carry-over from code generation (although I did copy it over from there).
___
*[Elements/src/Geometry/Profile.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjofhtX0oUpc1m4P-9t:-Mjz52_f1WVztqWK1J6D:blt4m9z) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Profile.cs#L12)):*

do we actually want to remove the `partial`?  It's not removed for Polygon for example, and I don't know that it should be.
</blockquote></details>

This does not need the partial any more because it has no counterpart. `Polygon` does. The class which contains our static constructor methods like `Polygon.Rectangle(...)` is the `partial` of `Polygon`.
___
*[Elements/src/Geometry/Solids/SolidOperation.cs, line 8 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-Mjor6nqA0Osa9pNtyWH:-Mjz5J_MElm5Hqx9XJ3i:b-h2wb5l) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Geometry/Solids/SolidOperation.cs#L8)):*

this also has that inheritance converter.
</blockquote></details>

See both notes above related to carrying over converters.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 76 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoS4K5EWdvsqgNGad_:-Mjz8uic7pPfAmTDiGfy:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L76)):*

can we make this check based on an Array of these prefixes rather than the multiple && logic
</blockquote></details>

Done.
___
*[Elements/src/Validators/Validator.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoojPY14HbbI5COFRW:-Mjz7f3l-mX8hwRgCpgC:bwggvwn) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Validators/Validator.cs#L12)):*

Removing the `Instance.GetFirstValidatorForType<T>()` entirely means this version of elements is incompatible with the Hypar.CLI.  can we leave that path in place, and mark it obsolete with a suggestion to upgrade your CLI, so that we can have this in Elements w/out breaking all developers who migrate to it?
</blockquote></details>

Correct. After these changes there is no validator. The only thing left is a static property. The singleton would have nothing to provide. I have a note in the PR to update the CLI and we can use our version notification system to message to the developer that they should upgrade.
___
*[Elements.Benchmarks/Program.cs, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-Mjoq6tS3ZHgjsj7dzB7:-Mjz6r4uAISBMSV5K-tO:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements.Benchmarks/Program.cs#L51)):*

seems like each of these classes and the `Program` entry point should be in their own file.
</blockquote></details>

Done.

wynged (changes requested):
> Reviewed 14 of 14 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[README.md, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoVMEP6FOLIsYPGL4J:-Mk29LTnEgpReb3OWrkH:b-q0rtc4) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/README.md#L51)):*

What platform ? ;) If we want to provide documentation for code generation from types because we think it's important for developers to know in relation to Hypar, we should provide that documentation in the Hypar docs.
</blockquote></details>

but couldn't you could use a generated type anywhere you use elements? And the bulk of the logic/code for code generation is in this repo.  We should also document that in the Hypar docs.
___
*[Elements/src/Validators/Validator.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoojPY14HbbI5COFRW:-Mk2AYFk9I_DylcTgKMa:bat7qet) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Validators/Validator.cs#L12)):*

Correct. After these changes there is no validator. The only thing left is a static property. The singleton would have nothing to provide. I have a note in the PR to update the CLI and we can use our version notification system to message to the developer that they should upgrade.
</blockquote></details>

We could make a gentler transition by making Validator a class, and providing a instance...  but if you think this is worth it then that's ok.

wynged (changes requested):
cool.

> Reviewed 31 of 31 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662#-:-MkAMFJk7I15WeRyY2hX:b-7s4s80)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[README.md, line 51 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoVMEP6FOLIsYPGL4J:-MkALhl0-Iu99o0GlHRI:bb7f2mf) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/README.md#L51)):*

I've restored the schemas (which are referenced in places like our GH docs) and moved the entirety of the comments related to code generation into a `README.md` in the `Elements.CodeGeneration` project.
</blockquote></details>

done
___
*[Elements/src/Validators/Validator.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoojPY14HbbI5COFRW:-MkALlsE1OHdfyC2oInh:bpxc7bn) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Validators/Validator.cs#L12)):*

I've reintroduces the Validator and Validators files and marked everything therein obsolete.
</blockquote></details>

oh sorry, I don't think we actually need any of the validators themselves, I just want to prevent breakage by leaving the Instance around and leaving methods that do nothing (and are obsoleted) around.  Do you think that if we do this we also need to leave Validators around that might try and do actual work?  maybe we need to chat about this.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Validators/Validator.cs, line 12 at r4](https://reviewable.io/reviews/hypar-io/elements/662#-MjoojPY14HbbI5COFRW:-MkDq-Io9HA6EGJysMv-:bpfp52b) ([raw file](https://github.com/hypar-io/elements/blob/720d7fe8eeda91bd4a8b596ab5755c0f0595610d/Elements/src/Validators/Validator.cs#L12)):*

oh sorry, I don't think we actually need any of the validators themselves, I just want to prevent breakage by leaving the Instance around and leaving methods that do nothing (and are obsoleted) around.  Do you think that if we do this we also need to leave Validators around that might try and do actual work?  maybe we need to chat about this. 
</blockquote></details>

In looking at where this would cause a problem, it's in generated classes that called `Validator.Instance.GetValidatorForType<T>()`. So that method needs to exist be deprecated to send a warning. And if the assumption is that the method will continue to do some validation, then the underlying methods need to continue to exist as well, but also be deprecated in case anyone is trying to use them in other situations.

I don't think either of these things will be a big deal because of the low number of developers, and we're free to remove all of this as soon as we're absolutely sure that no one is using it. You were right to suggest that we shouldn't break this, and this change actually does the right thing marking stuff as deprecated instead of just yanking it out.

wynged (commented):
Cool, LGTM just waiting to approve until after Elements release occurs.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662#-:-MkDw6B8Ala3JPr60n-P:b9deq5i)* status: 1 change requests, 0 of 1 approvals obtained

wynged (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/662#-:-MkHwa6QCntX5Hqvgbdo:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #666, opened 2021-09-16
fix ExtendTo when a segment crosses the origin

andrewheumann:
BACKGROUND:
- There was a bug in ExtendTo where it would return the wrong line if one of the segments being extended to crossed the origin. 

DESCRIPTION:
- verifies that `intersects` is true (rather than assuming it is) because the `out` variable from Intersects will contain `default(Vector3)` = `(0,0,0)` if intersects is false.

TESTING:
- I added a new test which used to fail and now passes. I also verified that all other tests are passing

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/666)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/666)* status: :shipit: complete! 1 of 1 approvals obtained

PR #667, opened 2021-09-20
Graceful failures for Profile Operations

andrewheumann:
BACKGROUND:
- Our clipper-based operations sometimes produce geometry that is not valid by hypar's standards — polygons with only 2 vertices, very very small polygons, or self-intersections. This causes issues in HyparSpace, where a Profile Difference might have *mostly* succeeded, but throws an exception if any of its resulting shapes is invalid. 

DESCRIPTION:
- Instead of throwing an exception inside Profile conversion operations, return a null value
- in code that calls these methods repeatedly (ie to populate a list), discard those nulls

TESTING:
- Two new tests are added which fail w/ current elements and now pass.
  
FUTURE WORK:
- We might need a mechanism to report out these failures when they've occurred — but an Exception is far too blunt an instrument.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/667)


cdriesler (approved):
safe and sound, agreed that this kind of thing shouldn't stop the whole show

also agreed with your future work note - maybe we add a `TODO: Indicate warning here` or something in all the

```
                if (profile != null)
                {
                    joinedProfiles.Add(profile);
                }
```

blocks? Not a strong opinion here though unsure what the sentiment around todos are in Elements.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/667#-:-Mk34r0a-yz0HQHQlMlS:b-m0hx47)* status: :shipit: complete! 1 of 1 approvals obtained

PR #670, opened 2021-09-23
add override identity from override object

andrewheumann:
BACKGROUND:
- In Parallel with https://github.com/hypar-io/Hypar/pull/416, I wanted to make it easier to call `Identity.AddOverrideIdentity` directly off of an override object received by the function.

DESCRIPTION:
- Adds an overrload for `Identity.AddOverrideIdentity()` that takes in the override object one has received into the function. 

TESTING:
- I am using this throughout my test project, as here: https://github.com/hypar-io/AddDeleteOverridesExample/blob/bff71037193e274d59330a3572ba5f76ee50e458/src/AddDeleteOverridesTest.cs#L90

REQUIRED:

COMMENTS:
- because of the use of `dynamic`, this required adding a reference to Microsoft.CSharp. I tried to find ways around this, such as setting up an interface, and linking to that via codegen, but because Hypar.Model doesn't reference Elements, and Elements doesn't reference Hypar.Model, it kept coming down to `dynamic` strategies. This seemed the smallest intervention.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/670)


serenayl (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/670)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Identity.cs, line 20 at r1](https://reviewable.io/reviews/hypar-io/elements/670#-MkO1XqqBQ007DIrjXpH:-MkO1Xqr5shj2_daE13-:b-e2uaso) ([raw file](https://github.com/hypar-io/elements/blob/80e00dac8678b7511eaa29645cfae60b3d57e4fb/Elements/src/Identity.cs#L20)):*
> ```
>  dynamic overrideObject
> ```
I'm not as familiar with what is available in C#, but is there a way that our overrides can all implement some kind of interface, and the interface comes with the ID, Identity, `GetName`, etc? Otherwise nothing is stopping me from doing something like:

`element.AddOverrideIdentity(new Material("red", new Color()));`

which is kind of confusing

serenayl (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/670)* status: :shipit: complete! 1 of 1 approvals obtained

PR #674, opened 2021-09-29
use content name instead of URL as rule name

andrewheumann:
BACKGROUND:
- @serenayl noticed that instances of content elements in hyparspace were named with their URLs

DESCRIPTION:
- Uses the content item's name, instead of its url, if present as the basis for the rule name, which ultimately becomes the instance name.

TESTING:
- I published a version of lounge layout on dev built against this version of elements.components, and verified that the name was correct.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/674)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/674)* status: :shipit: complete! 1 of 1 approvals obtained

PR #685, opened 2021-10-21
make the default useReferenceOrientation be false

wynged:
BACKGROUND:
- While working with Aliaxis content catalogs I realized that a transform error was occuring because the generated code for the ContentElements had transforms, which it turned out were no longer compatible with Revit import being aligned with gltf import.

DESCRIPTION:
- set the default catalog generation behavior to _not_ use referenceOrientations.  

TESTING:
- All transforms in the generated code in a catalog should be the identity transform.  This catalog url can be used to test catalog generation before and after.  use `hypar generate-catalog -u [THE-URL] -o [AN-OUTPUT-PATH]`  current cli (and therefor elements) will have various transforms generated.  A local CLI using this elements change will have all identity transforms.  
https://hypar-content-catalogs.s3.us-west-2.amazonaws.com/63e6e9b6-71c1-4f62-bb8e-daddfc1b8fd4/AkasisonXL_ReducerEccentricLong_INT-63e6e9b6-71c1-4f62-bb8e-daddfc1b8fd4.json
  
FUTURE WORK:
- Since this default has changed it might be the case that no code is using the `useReferenceOrientation` option, we should consider removing it.

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/685)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,15 @@
 # Changelog
+## 0.9.6
+
+### Added
+
+-
+### Changed
+
+- Change default for `useReferenceOrientation` when generating content catalogs.
+### Fixed
+
+-
 
 ## 0.9.5
 


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/685)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #687, opened 2021-10-24
Transform.CreateOrientedAlongCurve name is ambiguous as to expected behavior


Gytaco:
**Describe the bug**
I am trying to rotate a profile to a ZAxis using the following. code.
I was expecting the Transform.Transform.CreateOrientedAlongCurve to rotate the profile however this does not appear to work as expected.

var profile = new Profile(Polygon.Rectangle(50, 50));
Line line = new Line(new Vector3(0, 0), new Vector3(0, 1000));
var transform = Transform.CreateOrientedAlongCurve(line, 0);
var profiletr = profile .Transformed(transform);
var extrude = new Extrude(profiletr ,  Math.Abs(line.Length()), line.Direction(), false);

**To Reproduce**
Run the code above.

**Expected behavior**
I would expect the Transform.CreateOrientedAlongCurve to rotate the profile. So after I can see the profile is correctly updated.

PR #688, opened 2021-10-24
CreateOrientedAlongCurve => CreateHorizontalFrameAlongCurve

andrewheumann:
BACKGROUND:
- the name `CreateOrientedAlongCurve` was ambiguous as to _what_ orientation would be produced, leading to [confusion](https://github.com/hypar-io/Elements/issues/687).

DESCRIPTION:
- renames `CreateOrientedAlongCurve` => `CreateHorizontalFrameAlongCurve`

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/688)


ikeough:
This method has a lot of problems that simply renaming won't help. I probably wrote it :( For starters, this comment is false:
>/// Create a transform that is oriented along 
        /// a curve at parameter t. The transform's +z axis will align with 
        /// the +z world axis, and the +x axis will align with the tangent
        /// of the curve.
This transform has a skew. Internally we just negate two of the axes and set the z axis as the world z axis. This will break in all kinds of interesting ways, like a curve that goes vertically where the tangent is the world z axis. I'd propose that we deprecate this method altogether.

andrewheumann:
Adding @katehryhorenko because I think ian has forgotten about this one :)

katehryhorenko (approved):

andrewheumann (commented):
Oh ignore my comment — Somehow I missed @ikeough's. 

Re: Ian's comment —
I didn't realize that this didn't do what it says on the tin, since the behaviors are equivalent for XY-parallel polygons. To that effect, I've restored the behavior of the original method, and `Obsolete`d it. 
A "Frame oriented consistently horizontally but aligned to the curve tangent" though, is a useful thing — I used it in a training session today, as a matter of fact. So to that effect I've modified the implementation of the new method accordingly:


> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/688#-:-Mphl5MJ8byoJqNR3KzF:b-i2kgnc)* status: :shipit: complete! 1 of 1 approvals obtained

PR #695, opened 2021-11-02
Improvements to AdaptiveGrid

DmytroMuravskyi:
Skip tiny domains near start or end of the range. Line.PointAt doesn't like such domains. 
This would lead to degenerated line and exception.
 Improve performance of adding region by checking against global bounding box first

BACKGROUND:
This PR solves two situations:
- When working with AdaptiveGrid, it's creation was sometimes failing with exception because of small cells near beginning or end of the domain. In this case  Line.PointAt transformed these values near 0 or 1 into exactly 0 or 1. This led to changing something like [0; 5e-6] into [0; 0] - degenerated line.
- AdaptiveGrid.AddEdge was too slow because it checked each existing line for intersection with new one.

DESCRIPTION:
- Add check for tiny border cells and skip any found.
- Add bounding box check before doing any iterations over existing edges.

FUTURE WORK:
- The discussion might be needed about Line.PointAt. It's the only implementation of PointAt that does ApproximatelyEqual check.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/695)

### DIFF
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -347,115 +347,119 @@ namespace Elements.Spatial.AdaptiveGrid
             }
         }
 
-        private List<Edge> AddEdge(ulong startId, ulong endId, IEnumerable<Edge> edgesToIntersect)
+        private List<Edge> AddEdge(ulong startId, ulong endId,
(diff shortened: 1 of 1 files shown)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/695)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #696, opened 2021-11-02
Parametric Profile


ikeough:
Many things in the built world are defined by a profile swept along a path. Concrete bridge sections, structural steel, etc. Elements currently has a `Profile` type that is useful for representing a "perimeter" with one or more "void" curves. There are classes like `WideFlangeProfileServer` which add onto this the idea of a catalog of profiles. 

Adding a new profile type to Elements requires either writing and submitting code into Elements or writing code for your own profile type in your project. Profiles are generally defined as a set of vertices in the XY plane, each vertex offset by x and y from a center line. In the case of structural profiles though, to use one example, the shape of the profile is defined by a set of parameters that are usually listed in the supplier's catalog like: "d" for depth, and "t" for thickness. The code to generate the profile then takes the form of offsetting relative to the previous point, as you move around the profile in the CCW direction by amounts defined by those parameters. 

A general purpose method for describing a parametric profile in terms of parameters to be supplied by a catalogue, would enable users to define new Elements profile types without writing code.

This proposal is for the creation of a `ParametricProfile` type which takes as its parameter a configuration consisting of the following:
- A series of named parameters, which represent numerical values.
- A series of vertex location expressions which represent the location of each vertex in terms of the named parameters as two mathematical expressions, one for x and one for y.
The `ParametricProfile` constructor "compiles" the expressions, emits errors as necessary where properties are unknown or the expressions are incorrectly defined, and constructs a profile. Profile factories, like `WideFlangeProfileFactory` could then implement expression-based constructions of profiles based on databases of parameter values. 

Requirements:
- Expressions must be parsable by C# (the language of Elements).
- Expressions must be serializable/deserializable to/from JSON.

What this does not need to be:
- A constraint-based sketching environment with point constraints defined by the parameters. Parameters are known. Users should be able to create profiles that have incorrect shapes.

cdiggins:
Hi @ikeough could a profile just be an arbitrary closed 2D shape represented as an array of cubic bezier curves? Instead of arbitrary expressions? Each curve would have 4 control points. I think one could get a pretty good accuracy, but for machining/construction, it might not be accurate enough.

ikeough:
@cdiggins Just seeing this now, apologies for the delayed response. Yes, that could work, but `Profile` as it is now only supports one exterior polygon and a number of interior void polygons. We have another type, `Contour`, which is an array of curve types (`Line`, `Bezier`, `Arc`), which has not received much development and could be the target for what you're suggesting.

[...]

PR #697, opened 2021-11-03
Implement parametric profile.

ikeough:
BACKGROUND:
See #696 

DESCRIPTION:
- Adds a new type, `ParametricProfile` which derives from `Profile` and enables the user to specify the profile's vertex locations using expressions for X, and Y.
- Refactors profile servers to implement the `IProfileFactory` interface. "Old" profile servers derive from a base class `ProfileFactoryBase`, while all new profile servers derive from `ParametricProfileFactory`.
- Adds C, L, WT, W, MC, HSS, and ST sections implemented as section catalogues and parametric profile factories.

TESTING:
- See `ParametricProfileTests`
  
FUTURE WORK:
- The original `WideFlangeProfile` is deprecated and should be removed.
- The original `HSSPipeProfile` should be converted to a parametric profile type and the original should be deprecated.

REQUIRED:

image
image
image
image
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/697)


Elements/src/Geometry/Profiles/ParametricProfile.cs
```diff
@@ -0,0 +1,108 @@
...
+                throw new ArgumentException("The parametric profile could not be created. No translation expressions were provided.");
+            }
+
+            Data = data;
+
+            // These references are required to get dynamic to work.
+            _refs = new List<MetadataReference>{
+                                    MetadataReference.CreateFromFile(typeof(Microsoft.CSharp.RuntimeBinder.RuntimeBinderException).GetTypeInfo().Assembly.Location),
+                                    MetadataReference.CreateFromFile(typeof(System.Runtime.CompilerServices.DynamicAttribute).GetTypeInfo().Assembly.Location)};
+            Perimeter = CreatePolygonFromExpressions(data.PerimeterVectorExpressions, data.PropertyValues, _refs).Result;
```
radumg, line 45: using `.Result` in ctors can/will likely result in deadlocks
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#constructors

andrewheumann (changes requested):
> Reviewed 2 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/697)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @radumg)

___
*[Elements/src/Geometry/Profile.cs, line 26 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-Mnjknwv2mcCiDaftnXp:-Mnjknww36zxsQYFHALE:b-haxd4x) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profile.cs#L26)):*
> ```Smalltalk
>         /// and is not intended to be used directly.
>         /// </summary>
>         public Profile() { }
> ```

should this be `internal`, rather than `public`?
___
*[Elements/src/Geometry/Profile.cs, line 266 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-MnjixpA-gAj1yjeB0LF:-MnjixpA-gAj1yjeB0LG:bm73xh0) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profile.cs#L266)):*
> ```Smalltalk
>             // profile which defines its own perimeter and void logic
>             // during construction.
>             if (Perimeter == null || Voids == null)
> ```

are we 100% sure that profiles can't ever have null Voids in non-parametric profile cases? I think this is still possible See the `Profile(Polygon perimeter, Polygon @void)` constructor for one way to construct profiles that doesn't involve checking for a null `Voids` array. (We should probably just fix that instead of this, but maybe a `if this is ParametricProfile` or something would be prudent too?)
___
*[Elements/src/Geometry/Profiles/ParametricProfile.cs, line 23 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-MnjkwD32NuwZXDl1rO8:-MnjkwD32NuwZXDl1rO9:bm40a6p) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profiles/ParametricProfile.cs#L23)):*
> ```Smalltalk
>         /// The data which defines the profile.
>         /// </summary>
>         public ParametericProfileData Data { get; set; }
> ```

Parameteric  => Parametric

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/697)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @radumg)

___
*[Elements/src/Geometry/Profile.cs, line 26 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-Mnjknwv2mcCiDaftnXp:-Mnq7cIK8Wj0ePfuSpDR:bb6un1j) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profile.cs#L26)):*

should this be `internal`, rather than `public`?
</blockquote></details>

Done
___
*[Elements/src/Geometry/Profile.cs, line 266 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-MnjixpA-gAj1yjeB0LF:-Mnq7jXzA_JmCGip0kpn:b3nnnqt) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profile.cs#L266)):*

are we 100% sure that profiles can't ever have null Voids in non-parametric profile cases? I think this is still possible See the `Profile(Polygon perimeter, Polygon @void)` constructor for one way to construct profiles that doesn't involve checking for a null `Voids` array. (We should probably just fix that instead of this, but maybe a `if this is ParametricProfile` or something would be prudent too?)
</blockquote></details>

In any case where voids can be null, this would be necessary. Are you asking if I should just remove the comment? 
___
*[Elements/src/Geometry/Profiles/ParametricProfile.cs, line 23 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-MnjkwD32NuwZXDl1rO8:-Mnq8KGz7cTRVP5HCwKZ:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profiles/ParametricProfile.cs#L23)):*

Parameteric  => Parametric
</blockquote></details>

Done.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/697)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @radumg)

___
*[Elements/src/Geometry/Profile.cs, line 266 at r2](https://reviewable.io/reviews/hypar-io/elements/697#-MnjixpA-gAj1yjeB0LF:-Mnr5VIi0NxnSypUeZFS:blswq4s) ([raw file](https://github.com/hypar-io/elements/blob/21a28b9b2b794c44c81dd063bd8a2cf1cadc39de/Elements/src/Geometry/Profile.cs#L266)):*

In any case where voids can be null, this would be necessary. Are you asking if I should just remove the comment? 
</blockquote></details>

oh, that makes sense — was thrown off by the comment. disregard.

andrewheumann (commented):
> Reviewed 1 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/697)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @radumg)

PR #701, opened 2021-11-12
Grid2d change tolerance used for GetTrimmedGeometry.

DmytroMuravskyi:
BACKGROUND:
- When working on Grid2d based on boundary that is not rectangular - getting trimmed geometry is failed completely sometimes.  Intersection returns quantized distances with step of tolerance.
If intersection region is one EPSILON wide it's highly unstable:
4.00001 - 4 = 9.9999999996214228E-06 but 4 - 3.99999 = 1.0000000000065512E-05.
And so, such region may be cut off by Intersection - in this case the cell is just skipped.
But if number passed the tolerance test - chances are high it will fail after fromGrid transformation is applied and then whole trimming process will fail.

 Also, since polygon doesn't check number of vertices where was other unstable behaviour:
-Polygon has RemoveSequentialDuplicates in constructor that removes points less than tolerance away.
- Then, if only 2 points left Polyline.CheckSelfIntersectionAndThrow may or may not fail. This function uses Vertices.ToTransform() that takes first and second points difference as local X and other points as local Y. Since there are only 2 points → only X axis is initiated in transformation. If 2 given points have only X difference - the code will pass all the checks, otherwise will produce an exception.

DESCRIPTION:
- Added Vertex number check to Polygon constructor to avoid late exceptions.
- Raised tolerance used in Intersects to avoid tolerance fluctuations. As alternative - we could add distance check to ToPolygon that works with Int points but it would be required only in the case if tolerance parameter is equal to EPSILON that makes code clunky.

TESTING:
- Added  RotateGridWithClosePointDoNotThrow test to show the case that failed before due to transformation of the grid

FUTURE WORK:
- Are any other places that uses Intersects need update?
- Other boolean operations?
 
REQUIRED:
- [ x] All changes are up to date in `CHANGELOG.md`. (Do we need to mention something)

COMMENTS:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/701)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@
 - `Position.ToVectorMeters` now requires a `relativeToOrigin` Position, so that it will actually give meaningful measurements in meters.
 - glTF generation now uses material IDs instead of names for material names, to prevent collisions.
 - Line.PointAt does not round input values near 0 or 1 anymore.
+- `Polygon` constructor throws error if there are less than 3 vertices provided.
+- Decrease intersection tolerance for Grid2d polygon splitting.
 
 ### Fixed
 
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -33,7 +33,7 @@ namespace Elements.Geometry
                 }
 
                 this.Vertices = Vector3.RemoveSequentialDuplicates(this.Vertices, true);
-                
+
                 if (this.Vertices.Count < 3)
                 {
                     throw new ArgumentException("The polygon could not be created. At least 3 vertices are required.");
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -33,6 +33,12 @@ namespace Elements.Geometry
                 }
 
                 this.Vertices = Vector3.RemoveSequentialDuplicates(this.Vertices, true);
+                
+                if (this.Vertices.Count < 3)
+                {
+                    throw new ArgumentException("The polygon could not be created. At least 3 vertices are required.");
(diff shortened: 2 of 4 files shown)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/701)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @serenayl)

serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/701)* status: :shipit: complete! 2 of 1 approvals obtained

serenayl (changes requested):
Could you please put a note in the changelog under "Changed":
- 'Polygon' constructor throws error if there are less than 3 vertices provided.
- Decrease intersection tolerance for `Grid2d` polygon splitting.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/701#-:-ModyAW01vLq7tu-gvo-:b-cvblej)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained

serenayl (changes requested):
(except I meant `Polygon` rather than the single quote)

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/701#-:-ModykbnBdl-TegS6JKu:bpw1s3l)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained

serenayl (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/701#-:-MojFt5I0v-ZCY92z7Wf:bnfp4nl)* status: :shipit: complete! 2 of 1 approvals obtained

PR #703, opened 2021-11-15
Fixed Area calculation for Profiles in a plane other than XY

katehryhorenko:
BECKGROUND

Profile area is calculated using clipper. It's 2D tool and we are using only X and Y coordinates to get difference between perimeter and void polygons. As result Area is always 0 for profiles with voids in plane that other than XY

DESCRIPTION

The transformation is added to rotate profile and place it in the XY plane before calculating the difference

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/703)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/703)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[Elements/src/Geometry/Profile.cs, line 295 at r1](https://reviewable.io/reviews/hypar-io/elements/703#-Mo_zi_PEFxa2RNGzaSS:-Mo_zi_Q4cY-dxiyriDl:b-bqcdri) ([raw file](https://github.com/hypar-io/elements/blob/28a1fa7de2a834b7261013f02668894342abbbd2/Elements/src/Geometry/Profile.cs#L295)):*
> ```Smalltalk
> 
>             var clipper = new ClipperLib.Clipper();
>             var tranfrorm = new Transform(Perimeter.Start, Perimeter.Normal());
> ```

transfrorm => transform

also — can we check if the normal is nearly the z axis, and skip the transform if so?

katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/703)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Profile.cs, line 295 at r1](https://reviewable.io/reviews/hypar-io/elements/703#-Mo_zi_PEFxa2RNGzaSS:-Mpmsr2D-KZ-rGusQ6by:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/28a1fa7de2a834b7261013f02668894342abbbd2/Elements/src/Geometry/Profile.cs#L295)):*

transfrorm => transform

also — can we check if the normal is nearly the z axis, and skip the transform if so? 
</blockquote></details>

Done.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/703)* status: :shipit: complete! 1 of 1 approvals obtained

PR #704, opened 2021-11-18
Simplify to graphics buffers

serenayl:
BACKGROUND:
- We need to support GridLine, which is an element type that is made up of multiple curves.

DESCRIPTION:
- Add internal skeleton method `TryToGraphicsBuffers` to `GeometricElement`, which has no functionality.
- Override this method in `GeometricElement`s that do have their own logic.
- Return a list of graphics buffers, rather than a single one, and write them all together in GLTF extensions.

TESTING:
- All GLTFs should continue to look and render and behave the same.
  
FUTURE WORK:
- Gridline class is coming after.

REQUIRED:
- ~~[ ] All changes are up to date in `CHANGELOG.md`.~~ (not updated because changes are internal)

COMMENTS:
- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/704)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/704)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @serenayl)

___
*[Elements/src/GeometricElement.cs, line 211 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCmAbw3pZa1Y72cY1o:-MpCmAbw3pZa1Y72cY1p:bliu8b) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/GeometricElement.cs#L211)):*
> ```Smalltalk
>         /// Out variables should be ignored if the return value is false.
>         /// </returns>
>         internal Boolean TryToGraphicsBuffers(out List<GraphicsBuffers> graphicsBuffers, out string id, out glTFLoader.Schema.MeshPrimitive.ModeEnum? mode)
> ```

This should be marked `virtual` and overridden with the `override` keyword, instead of `new`. The distinction is subtle, but here's why:

```
class Program
{
    static void Main(string[] args)
    {
        Console.WriteLine("Hello World!");
        Foo bar = new Bar();
        bar.DoSomething();
        bar.DoSomethingElse();
    }
}

public class Foo
{
    public virtual void DoSomething()
    {
        Console.WriteLine("Foo.DoSomething");
    }

    public void DoSomethingElse()
    {
        Console.WriteLine("Foo.DoSomethingElse");
    }
}

public class Bar : Foo
{
    public override void DoSomething()
    {
        Console.WriteLine("Bar.DoSomething");
    }

    public new void DoSomethingElse()
    {
        Console.WriteLine("Bar.DoSomethingElse");
    }
}
```


This code will print 
`Bar.DoSomething`
`Foo.DoSomethingElse`

In other words, because we're storing `Bar` in a variable of type `Foo`, the "new" way of overriding its behavior gets ignored (reverts to `Foo` behavior), but the "virtual + override" way preserves the overridden behavior.
___
*[Elements/src/ModelArrows.cs, line 121 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCoSJOComJgj5WHS5m:-MpCoSJP9voxp5LgH5k9:b-c4xnge) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/ModelArrows.cs#L121)):*
> ```Smalltalk
>         }
> 
>         new internal Boolean TryToGraphicsBuffers(out List<GraphicsBuffers> graphicsBuffers, out string id, out glTFLoader.Schema.MeshPrimitive.ModeEnum? mode)
> ```

per above, all these `new`s will become `override`s.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1088 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCoaBVAb-BoVOluTpB:-MpCoaBVAb-BoVOluTpC:b-a57zhp) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/Serialization/glTF/GltfExtensions.cs#L1088)):*
> ```Smalltalk
>             if (e is GeometricElement)
>             {
>                 var eType = e.GetType();
> ```

Using the `override` method above, all this reflection should also not be necessary. It should suffice to call `ge.TryToGraphicsBuffers` directly.

serenayl (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/704)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[Elements/src/GeometricElement.cs, line 211 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCmAbw3pZa1Y72cY1o:-MpD-XveCfJaH0JgxIg8:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/GeometricElement.cs#L211)):*

This should be marked `virtual` and overridden with the `override` keyword, instead of `new`. The distinction is subtle, but here's why:

```
class Program
{
    static void Main(string[] args)
    {
        Console.WriteLine("Hello World!");
        Foo bar = new Bar();
        bar.DoSomething();
        bar.DoSomethingElse();
    }
}

public class Foo
{
    public virtual void DoSomething()
    {
        Console.WriteLine("Foo.DoSomething");
    }

    public void DoSomethingElse()
    {
        Console.WriteLine("Foo.DoSomethingElse");
    }
}

public class Bar : Foo
{
    public override void DoSomething()
    {
        Console.WriteLine("Bar.DoSomething");
    }

    public new void DoSomethingElse()
    {
        Console.WriteLine("Bar.DoSomethingElse");
    }
}
```


This code will print 
`Bar.DoSomething`
`Foo.DoSomethingElse`

In other words, because we're storing `Bar` in a variable of type `Foo`, the "new" way of overriding its behavior gets ignored (reverts to `Foo` behavior), but the "virtual + override" way preserves the overridden behavior.
</blockquote></details>

Done.
___
*[Elements/src/ModelArrows.cs, line 121 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCoSJOComJgj5WHS5m:-MpD-ZVG5yqmJLQl80bo:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/ModelArrows.cs#L121)):*

per above, all these `new`s will become `override`s.
</blockquote></details>

Done.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1088 at r1](https://reviewable.io/reviews/hypar-io/elements/704#-MpCoaBVAb-BoVOluTpB:-MpD-YvDCxMJT2cVl6js:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/d79096b657ef3adfc59a9f414847a32cdb6a1837/Elements/src/Serialization/glTF/GltfExtensions.cs#L1088)):*

Using the `override` method above, all this reflection should also not be necessary. It should suffice to call `ge.TryToGraphicsBuffers` directly. 
</blockquote></details>

Done.

andrewheumann (approved):
LGTM — could use a test as an example, feel free to use [this one](https://reviewable.io/reviews/hypar-io/Elements/711#-) :) 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/704#-:-MpDrf3u21qu-nAelg7M:b-kh464w)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

PR #705, opened 2021-11-18
Add gridline

serenayl:
BACKGROUND:
- We need a selectable gridline to attach overrides to.

DESCRIPTION:
- Creates a "GridLine" element with a start and end point, and options for how long the start and end extensions are, as well as the grid head.

TESTING:
- See https://reviewable.io/reviews/hypar-io/BuildingBlocks/69#-
  
FUTURE WORK:
- We may want a way to also include the render of the name, but we have previously discovered that splitting `ModelText` leads to serious performance problems.

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/705)


andrewheumann (approved):
> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/705)* status: :shipit: complete! 1 of 1 approvals obtained

PR #650, opened 2021-08-17
Add tolerance to Vector3.Unitized

nadiia-volyk:
BACKGROUND:
- `Vector3Extensions.AreCollinear` didn't work for coincident points. 
- A vector with almost zero values will still be unitized

DESCRIPTION:
- Add tolerance to Vector3.Unitized

TESTING:
- Added a new test, which used to fail, and now succeeds

REQUIRED:
- [ ] All changes are up to date in `CHANGELOG.md`.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/650)

PR #714, opened 2021-11-28
Fix intersect exception

andrewheumann:
BACKGROUND:
- In https://github.com/hypar-io/Elements/pull/650, we added a tolerance to `Vector3.Unitized`, in order to fix an exception with `Vector3Extensions.AreCollinear`.
- @Gytaco submitted this PR, https://github.com/hypar-io/Elements/pull/712, aiming to fix another exception in `Line.Intersects`, which was also related to `Vector3Extensions.AreCollinear`. In this case, the three points being checked for collinearity were far apart:
```
[(13540, -1238), (9840, -1238), (9840.0341, -1238.0003)]
```
image (NTS)

While these points are very nearly collinear, they are **not** collinear within the default tolerance of 1e-5. However, the `Plane` constructed through these three points winds up with a nearly zero-length normal, because of tolerance in `Vector3.Unitized` which causes it to return the vector untransformed, rather than correctly normalized. This in turn produces an exception when we try to construct the intersection plane, which @Gytaco's PR fixes by choosing a different plane in case of a zero-length normal. However, the zero-length normal should never have been created in the first place.

There were other recent changes to the `AreCollinear` method in #707, which seem to resolve the issue #650 was intended to address without the need for a `Unitized` tolerance.

DESCRIPTION:
- Incorporates the test from #712 
- Removes the tolerance parameter from `Vector3.Unitized`, effectively reverting #650 (but preserving its tests!)

TESTING:
- All tests pass, including the one brought in from #712  and the original test from #650

REQUIRED:

COMMENTS:
- It's conceivable that I'm missing a case where this tolerance check would be desirable — would love another pair of eyes on this problem to make sure I'm not making a mistake by reverting this.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/714)


mattc (approved):
I think this is the right move. This is equivalent to the pattern of using precise equality checks in the serialization layer.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/714#-:-MphhwkD9L_HkRUfVScn:b-6bqp0s)* status: :shipit: complete! 1 of 1 approvals obtained

PR #717, opened 2021-12-02
Open Polyline Offset

andrewheumann:
BACKGROUND:
- All our polyline offset methods produce closed polygons. We need one that can produce an open polyline from an open polyline.

DESCRIPTION:
- adds a new `Polyline.OffsetOpen` method

TESTING:
- A new `OffsetOpen` test is added:
image
  
FUTURE WORK:
- This is a very naïve method. It does not handle eliminating self-intersections at concave corners, as seen in the test image. It also assumes 2D only. `OffsetOnSide` has better corner handling (but does not join the individual polygonal segments), and it would be nice to shift the logic of this method to take on similar behavior for the sake of consistency.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/717)

### DIFF
--- a/Elements/src/Geometry/Polyline.cs
+++ b/Elements/src/Geometry/Polyline.cs
@@ -388,7 +388,7 @@ namespace Elements.Geometry
             {
                 if (s.Length() == 0)
                 {
-                    throw new ArgumentException("A segment fo the polyline has zero length.");
+                    throw new ArgumentException("A segment of the polyline has zero length.");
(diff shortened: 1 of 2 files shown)


mattc (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/717)* status: :shipit: complete! 1 of 1 approvals obtained

PR #718, opened 2021-12-03
AdaptiveGrid: Remove AddEdge that does intersections. Double the Tolerance.

DmytroMuravskyi:
BACKGROUND:
- After updating Elements to the latest version code that used Adaptive grid didn't traverse it as expected anymore.
- The reason was changing tolerance of intersect from Epsilon to Epsilon * 2. New tolerance is still better because grid doesn't fail randomly anymore but only individual cells with width [0, Epsilon] may appear. 
- Was also discovered that `AddEdge` that should have created new vertices after a lot of intersection operations didn't do so because it has wrong calculations inside of its fast checks that were designed to improve operation performance. Everything worked fine before because split points were aligned between regions of the grid.  Since adding regions to the grid was biggest bottleneck for large grid create from many regions, I decided not to fix broken function but to skip intersections completely and rely only on split points.

DESCRIPTION:
- Tolerance is doubled to ignore tiny cracks  that may be produced by Grid2D.
- AddEdge is removed because it never truly worked: was huge performance bottleneck and simplified version never worked correctly.

TESTING:
- Added test "AdaptiveGridTwoAlignedSections" covering two aligned regions
  
FUTURE WORK:
- The AdaptiveGrid is complex to use at the moment because you need to choose split points carefully. We might need to add something new that simplifies the process without introducing performance bottleneck.

REQUIRED:
- [ x] All changes are up to date in `CHANGELOG.md`.

COMMENTS:
@serenayl  this is performance improvement we discussed together with regression fix

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/718)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 - Improved the logic of `AreCollinear` to utilize perpendicular distance for tolerance checks.
 - `BBox3` constructor now takes an `IEnumerable<Vector3>` instead of a `IList<Vector3>` as input.
 - `Vector3Extensions.Unitized` no longer takes a tolerance for its zero-length check.
+- `AdaptiveGrid` no longer inrsect new edges with all existing edges when new region is added to the grid.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


wynged (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/718)* status: :shipit: complete! 1 of 1 approvals obtained

PR #720, opened 2021-12-06
Permit using Content Configurations dynamically

andrewheumann:
BACKGROUND:
- There were some challenges in utilizing content configurations dynamically within a function.
- The ContentConfiguration class dynamically matches content in referenced configurations to a loaded catalog assumed to be at `./catalog.json`  via the `ContentCatalogRetrival` class.

DESCRIPTION:
- Adds `[JsonIgnore]` attribute to the computed ContentElement property, which is only needed at runtime
- Support specifying a different `catalog.json` path in case the catalog was fetched dynamically in a function and needs to be saved to a `temp` location.

TESTING:
- I created a `Create Content Configuration` function (see dev workflow `acdeaa39-f97a-46fd-856d-50fc76910bb9`) that loads a content catalog on the fly and then instantiates it using the components framework.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/720)

### DIFF
--- a/Elements.Components/src/ContentCatalogRetrieval.cs
+++ b/Elements.Components/src/ContentCatalogRetrieval.cs
@@ -41,16 +41,7 @@ namespace Elements.Components
                     throw new Exception("Catalog not found. To use this class, make sure you have a content catalog stored at catalog.json in the build directory.");
                 }
                 var json = File.ReadAllText(catalogPath);
-                // there are two catalog formats in the wild — one encoded as a model, one with the content catalog encoded bare.
-                try
-                {
-                    var model = Model.FromJson(json);
-                    catalog = model.AllElementsOfType<ContentCatalog>().First();
-                }
-                catch
-                {
-                    catalog = JsonConvert.DeserializeObject<ContentCatalog>(json);
-                }
+                catalog = ContentCatalog.FromJson(json);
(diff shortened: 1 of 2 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/720)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements.Components/src/ContentCatalogRetrieval.cs, line 44 at r1](https://reviewable.io/reviews/hypar-io/elements/720#-MqNGbM0DSJ4cq3WMzN1:-MqNGbM0DSJ4cq3WMzN2:b-o76orn) ([raw file](https://github.com/hypar-io/elements/blob/eca0b321f24de01bf3047fd8b7065f291175175a/Elements.Components/src/ContentCatalogRetrieval.cs#L44)):*
> ```Smalltalk
>                 }
>                 var json = File.ReadAllText(catalogPath);
>                 // there are two catalog formats in the wild — one encoded as a model, one with the content catalog encoded bare.
> ```

any reason this method wouldn't work here?  it handles the two types of catalog https://github.com/hypar-io/Elements/blob/master/Elements/src/ContentCatalog.cs#L53

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/720)* status: :shipit: complete! 1 of 1 approvals obtained

PR #721, opened 2021-12-06
Fix Line Trim when a vertex is very close

andrewheumann:
BACKGROUND:
- a Hypar Space workflow demonstrated an issue where no walls were being generated between offices.  This turned out to trace backwards through `Grid.GetCellSeparators()` => `Line.Trim(Polygon)`.
- Specifically, a line that started just outside of a polygon — almost exactly within tolerance of the edge — was showing as "greater than tolerance" in one check (The perpendicular distance check in Polygon.Contains, Polygon.cs, line 495) but "less than tolerance" in the other (The "duplicate point" check in Line.Trim, Line.cs line 877).
image
- Since the point was seen as outside, but then the second point was ignored as a duplicate, `Line.Trim` was assuming the whole segment lay outside of the trimming polygon. If the first check had been under tolerance, it would have returned a containment of `CoincidentAtEdge` instead of `Outside`, and we would have gotten the right behavior. 

DESCRIPTION:
- If we encounter a duplicate vertex in `Line.Trim`, we have to update the containment check using the "skipped" vertex, so we can correctly classify the segment.

TESTING:
- I added a new test that demonstrates the behavior in question. This test was failing prior to this change.
- I also ran a local version of the function that was failing, and verified that it produced walls as expected.
  
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/721)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.9.7
+
+### Added
+
+### Changed
+
+### Fixed
+
+- Under some circumstances when a line originated nearly within tolerance of a polygon, `Line.Trim` would return the wrong result.
+
 ## 0.9.6
 
 ### Added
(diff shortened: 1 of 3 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/721)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #719, opened 2021-12-05
Support an optional disk cache for content gltf resolution


andrewheumann:
As of #539, we utilize an [in-memory cache](https://github.com/hypar-io/Elements/blob/f5f5e2cb24cc61275fbf08e41bcebe357e62d71d/Elements/src/Serialization/glTF/GltfExtensions.cs#L1132) for GLTFs we're merging from content elements. However, with the new `hypar run` server strategy (Hypar CLI version 0.9.7+) where the process completely restarts, this means GLBs are fetched every time the user makes a code edit, which can be slow. It would be nice to support an optional disk cache (or other similar mechanism) that can remain present through subsequent code edits on `hypar run`.

PR #726, opened 2021-12-09
set up an optional disk cache for content element glbs

andrewheumann:
BACKGROUND:
- https://github.com/hypar-io/Elements/issues/719
- Repeatedly running a test, or using `hypar run` with a function utilizing content elements can be painfully slow, since the referenced GLBs were getting fetched every single time. 

DESCRIPTION:
- Adds a completely optional disk cache for `GltfExtensions.gltfCache`, which is only utilized if `gltfExtensions.GltfCachePath` is set. 

TESTING:
- Running off local elements, I tested the following:
  - running a generated function test with content elements that I modified by setting `GltfCachePath` at the beginning of the test. I verified that each subsequent test ran MUCH faster than the first one, and that files were getting stored at the specified directory.
  - Using the `hypar run` command, with its server project modified to set `GltfCachePath` at the beginning. I verified that it was much faster than previously in after a restart. 
  
FUTURE WORK:
- Once this is released, we should update `HyparServer.StartAsync` and our `generate test` templates to call this automatically with a temp directory, so that the user doesn't even need to know about this option.

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/726)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Content Elements can now use an optional disk cache when running locally for testing purposes, to speed up repeated tests or runs, by setting `GltfExtensions.GltfCachePath`.
+
 ### Changed
 
 ### Fixed
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -35,6 +35,17 @@ namespace Elements.Serialization.glTF
             return _currentId;
         }
 
+        /// <summary>
+        /// In normal function use, this should be set to null.
+        /// If not null and set to a valid directory path, gltfs loaded for 
+        /// content elements will be cached to this directory, and can be
+        /// explicitly loaded by calling LoadGltfCacheFromDisk(). This is used
+        /// by `hypar run` and test capabilities to speed up repeated runs. 
+        /// </summary>
+        public static string GltfCachePath { get; set; } = null;
+
+        private const string GLTF_CACHE_FOLDER_NAME = "elementsGltfCache";
+
         private const string emptyGltf = @"{
     ""asset"": {""version"": ""2.0""},
     ""nodes"": [{""name"": ""empty""}],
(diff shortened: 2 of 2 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/726)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 45 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhEoFDZDubqQUEsV_:-MqVhEoFDZDubqQUEsVa:b-nsoixg) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L45)):*
> ```Smalltalk
>         /// by `hypar run` and test capabilities to speed up repeated runs. 
>         /// </summary>
>         public static string GltfCachePath { get; set; } = null;
> ```

I think we should have some logic in the `set` path here that actually checks if the directory is valid, and console logs if not.  This would also allow us to set a private boolean for "useLocalCache" rather than checking for directory existence every time.  It seems unlikely that the directory would be created during the running of the cache.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1147 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhZ7tCTgaax5I1tm7:-MqVhZ7tCTgaax5I1tm8:b-grl09l) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L1147)):*
> ```Smalltalk
>         private static void WriteGltfCacheForKey(string key, MemoryStream stream)
>         {
>             if (GltfCachePath == null || !Directory.Exists(GltfCachePath))
> ```

see above,  I think we can have a bool that contains this information above rather than checking everytime.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1179 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhmDf5FE7wArMO-3W:-MqVhmDg1r3z9l9j4sr6:b-grl09l) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L1179)):*
> ```Smalltalk
>             {
>                 return;
>             }
> ```

see above,  I think we can have a bool that contains this information above rather than checking everytime.

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/726)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 45 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhEoFDZDubqQUEsV_:-MqZcs7WDX25p4cmJ8sn:bgjhxlk) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L45)):*

I think we should have some logic in the `set` path here that actually checks if the directory is valid, and console logs if not.  This would also allow us to set a private boolean for "useLocalCache" rather than checking for directory existence every time.  It seems unlikely that the directory would be created during the running of the cache.
</blockquote></details>

Good call. I'm using the backing `string` field instead of an additional boolean, and just not setting the string if it's not a valid path. 
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1147 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhZ7tCTgaax5I1tm7:-MqZcWYzCSbwlNB_x6wg:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L1147)):*

see above,  I think we can have a bool that contains this information above rather than checking everytime.
</blockquote></details>

Done.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 1179 at r1](https://reviewable.io/reviews/hypar-io/elements/726#-MqVhmDf5FE7wArMO-3W:-MqZd0lq6HYHn6tcmj1-:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/e741cf99cad3efdfa94558ead54c4ff56a667b9f/Elements/src/Serialization/glTF/GltfExtensions.cs#L1179)):*

see above,  I think we can have a bool that contains this information above rather than checking everytime.
</blockquote></details>

Done.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/726)* status: :shipit: complete! 1 of 1 approvals obtained

PR #713, opened 2021-11-25
Use a single buffer for a gridline

andrewheumann:
BACKGROUND:
- I noticed while testing the new selectable gridlines that because it produces multiple buffers, each section of the grid would highlight / "select" individually. 

DESCRIPTION:
- Uses a single buffer for a gridline, so that it appears as a single selectable object.

TESTING:
- I ran the test and visualized the GLB to verify that it looked correct with a single buffer.
  
FUTURE WORK:
- In a case of multiple buffers, we might want to create some sort of a parent node in the glb so that multiple buffers can be treated as a single selectable object.

PR #727, opened 2021-12-09
Support curved gridlines

serenayl:
BACKGROUND:
- The gridlines as currently specified in the schema manager support arbitrary geometry. We would like to retain this flexibility.

DESCRIPTION:
- Deprecates `GridLine.Line` and uses `GridLine.Curve` instead.
- Adds `GridLine.GetCircleTransform()` method to help with text placement, which is still currently separate from GridLine geometry.

TESTING:
- Try the updated test examples for multiple gridlines with `ModelText`, as well as for a curved GridLine.
  
FUTURE WORK:
- How can we also include the text piece of this?

REQUIRED:

COMMENTS:
- Rolls in and should supersede #713

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/727)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.9.7
 
 ### Added
+- `GridLine.GetCircleTransform`
 
 ### Changed
 - Support non-linear gridlines by deprecating `GridLine.Line` and replacing it with `GridLine.Curve`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.9.7
 
 ### Added
-- `GridLine.GetCircleTransform`
+- `GridLine.GetCircleTransform()`
 
 ### Changed
 - Support non-linear gridlines by deprecating `GridLine.Line` and replacing it with `GridLine.Curve`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- Support non-linear gridlines by deprecating `GridLine.Line` and replacing it with `GridLine.Curve`.
 
 ### Fixed
 
--- a/Elements/src/GridLine.cs
+++ b/Elements/src/GridLine.cs
@@ -13,7 +13,14 @@ namespace Elements
         /// Line that runs from the start of the gridline to its end.
         /// </summary>
         /// <value></value>
+        [Obsolete("We now use 'Curve' instead.")]
         public Line Line { get; set; }
+
+        /// <summary>
+        /// Curve that runs from the start of the gridline to its end.
+        /// </summary>
+        /// <value></value>
+        public Curve Curve { get; set; }
 
         /// <summary>
         /// Radius of the gridline head.
@@ -32,7 +39,7 @@ namespace Elements
 
         internal override Boolean TryToGraphicsBuffers(out List<GraphicsBuffers> graphicsBuffers, out string id, out glTFLoader.Schema.MeshPrimitive.ModeEnum? mode)
         {
-            if (this.Line == null)
+            if (this.Curve == null)
             {
                 return base.TryToGraphicsBuffers(out graphicsBuffers, out id, out mode);
             }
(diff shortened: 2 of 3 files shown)


andrewheumann (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/727)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/GridLine.cs, line 111 at r2](https://reviewable.io/reviews/hypar-io/elements/727#-MqZB1oN3SVf_Q3nhNMm:-MqZB1oN3SVf_Q3nhNMn:b-tvuqvp) ([raw file](https://github.com/hypar-io/elements/blob/cb2b73d72639f2357341495cc82e60ca1cd4b2ed/Elements/src/GridLine.cs#L111)):*
> ```Smalltalk
>         /// <param name="texts">Collection of texts to add to.</param>
>         /// <param name="color">Color for this text.</param>
>         public void AddTextToCollection(List<(Vector3 location, Vector3 facingDirection, Vector3 lineDirection, string text, Color? color)> texts, Color? color = null)
> ```

What do you think about this method accepting a ModelText, instead of a list? Personally I cringe every time I have to write out that full tuple type in function code. We could adjust the ModelText constructor to have defaults for arguments, and then it's just: 
```
var mt = new ModelText {
   FontSize = FontSize.PT72, 
   Scale = 50
};
var gridline = new GridLine(...);
gridline.AddText(mt);
```


(an alternative strategy, maybe for a later PR, would be to make a `ModelTextItem` class/struct to replace the tuple. we could write implicit converts from the tuple to the class, and a List<tuple> => List<ModelTextItem> so we wouldn't even break any code)

((my personal long term vision for this is that we stop using `ModelText` for `GridLine`s in favor of screen-facing special always-on labels. but that's not for today :smile: ))

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/727)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/GridLine.cs, line 111 at r2](https://reviewable.io/reviews/hypar-io/elements/727#-MqZB1oN3SVf_Q3nhNMm:-MqZDDqs3AQ8OsK6XZfo:b6vzbg0) ([raw file](https://github.com/hypar-io/elements/blob/cb2b73d72639f2357341495cc82e60ca1cd4b2ed/Elements/src/GridLine.cs#L111)):*

What do you think about this method accepting a ModelText, instead of a list? Personally I cringe every time I have to write out that full tuple type in function code. We could adjust the ModelText constructor to have defaults for arguments, and then it's just: 
```
var mt = new ModelText {
   FontSize = FontSize.PT72, 
   Scale = 50
};
var gridline = new GridLine(...);
gridline.AddText(mt);
```


(an alternative strategy, maybe for a later PR, would be to make a `ModelTextItem` class/struct to replace the tuple. we could write implicit converts from the tuple to the class, and a List<tuple> => List<ModelTextItem> so we wouldn't even break any code)

((my personal long term vision for this is that we stop using `ModelText` for `GridLine`s in favor of screen-facing special always-on labels. but that's not for today :smile: ))
</blockquote></details>

oh! and the other necessary change to the `ModelText` class to support the above idea is that I think 
```
GenerateTextureAtlas();
GenerateMesh();
```
would have to be moved to an override on `UpdateRepresentations`, otherwise they won't update as the text collection is modified. I *think* that would work.... if not we could put an `AddText` method on ModelText that calls those methods. But I like the idea that we only construct the mesh and texture atlas once at gltf-generation time.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/727)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @serenayl)

___
*[Elements/src/GridLine.cs, line 111 at r2](https://reviewable.io/reviews/hypar-io/elements/727#-MqZB1oN3SVf_Q3nhNMm:-MqZIsEQDHqupU3roRYj:b-l107kj) ([raw file](https://github.com/hypar-io/elements/blob/cb2b73d72639f2357341495cc82e60ca1cd4b2ed/Elements/src/GridLine.cs#L111)):*

oh! and the other necessary change to the `ModelText` class to support the above idea is that I think 
```
GenerateTextureAtlas();
GenerateMesh();
```
would have to be moved to an override on `UpdateRepresentations`, otherwise they won't update as the text collection is modified. I *think* that would work.... if not we could put an `AddText` method on ModelText that calls those methods. But I like the idea that we only construct the mesh and texture atlas once at gltf-generation time.
</blockquote></details>

One last thing! :joy: If you go with this strategy, don't forget to update the `Texture` property on the ModelText's Material to the `_texturePath` after `GenerateTextureAtlas` is run

serenayl (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/727)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/GridLine.cs, line 111 at r2](https://reviewable.io/reviews/hypar-io/elements/727#-MqZB1oN3SVf_Q3nhNMm:-MqodW4R2dO_NEZRTvK3:bj59rzf) ([raw file](https://github.com/hypar-io/elements/blob/cb2b73d72639f2357341495cc82e60ca1cd4b2ed/Elements/src/GridLine.cs#L111)):*

One last thing! :joy: If you go with this strategy, don't forget to update the `Texture` property on the ModelText's Material to the `_texturePath` after `GenerateTextureAtlas` is run
</blockquote></details>

Handling in follow-n PR

PR #730, opened 2021-12-15
Substract timeouts

wynged:
BACKGROUND:
- There have been a couple of conditions that resulted in csg infinite loop bugs that are frustrating to find and debug.

DESCRIPTION:
- create timebound tasks that will throw if a CSG takes more than a given amount of time. currently set to 1 second, but this can be changed.  Seems like all unions we are doing should be sub 1 second though.
- Create two tests, one for union and one for substract that would both cause infinite execution time if we didn't have these fixes.

TESTING:
- Tests pass that would have failed.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- We do pay a speed penalty for having this protection in there, but it feels worth it to me since this has come up so much recently.  Here are benchmarks before and after.  Seems to be a 50ms constant overhead penalty.
### Before this change
image

### After this change
image

- It's possible that we can actually simply return when the operation is taking a long time, and we still get valid results... this feels highly insecure, but this is what an alternative `assert` would look like
 ```
             // var taskFinishedOnTime = Task.Run(() =>
            // {
            //     wallByProfile.UpdateRepresentations();
            //     var solid = wallByProfile.GetFinalCsgFromSolids();
            // }).Wait(2000);
            // Assert.True(taskFinishedOnTime);

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/hypar-io/elements/730)
<!-- Reviewable:end -->


ikeough:
Is this because we can't find or deal with the root problem? Can we instead run CSG under the debugger and see where the error is happening and fix it there? A 50ms penalty for a geometric operation is not acceptable.

wynged (commented):
Ok, this has been updated to use a more modified version of CSG that fixes issues finding normal vectors when polygons had inline vertices.  This is that CSG branch https://github.com/hypar-io/Csg/tree/hypars-branch 

I'm not sure how our polygons have sequential duplicates, I think we run this method on construction. https://github.com/hypar-io/Elements/blob/6bf4ecac19cddacaafe9f3796f272adeae7a939f/Elements/src/Geometry/Polygon.cs#L35

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/730#-:-Mr0pg66DJnwth8qYdfl:bx684io)* status: 0 of 1 approvals obtained (waiting on @katehryhorenko)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Changed
 - Support non-linear gridlines by deprecating `GridLine.Line` and replacing it with `GridLine.Curve`.
+- Add exceptions for timeout during long running csg operations.
```
katehryhorenko, line 12: I believe that change description has changed :)

katehryhorenko (approved):

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/730)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[CHANGELOG.md, line 12 at r2](https://reviewable.io/reviews/hypar-io/elements/730#-Mqy4Yp2z-3IQ80u0GDb-r2-12:-Mr3fDDC1lmCTViW6fDL:b-2koddt) ([raw file](https://github.com/hypar-io/elements/blob/fb0ae9b0d23adf13b959e6d9cd4ff629a7b98b21/CHANGELOG.md#L12)):*

I believe that change description has changed :)
</blockquote></details>

Done. thanks!

PR #731, opened 2021-12-15
Added Cable element;

katehryhorenko:
DESCRIPTION:
This PR contains new Cable element that can be build from the polyline with a given Radius 

TESTING:
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/731)


ikeough:
Is there a way that we can represent this more generally as a "network with radii". This connects to some of the conversations that @andrewheumann and I have had regarding sketches. LIke a network could have vertices that have an associated radius. Then we could draw pipes, cables, roads, etc.

andrewheumann (commented):
@ikeough to my mind the "network with radii" is a general utility but we would still want element types for the actual thing in question (like a cable). just my 2¢! 

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/731#-:-Mr2Vng03sU5Gb3Q7WD_:b-gsetl8)* status: 0 of 1 approvals obtained (waiting on @mattc)

mattc (commented):
+1. We shouldn't use inheritance to indicate that they share this piece of behavior - we should compose the behavior.

My question is whether this belongs in Elements at all. Shouldn't this be a schema-based class with UpdateRepresentation as an extension to the generated partial class? But I'm not clear on where that UpdateRepresentation should go if it'll eventually be used by multiple functions (which it won't be initially).

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/731#-:-Mr2gv320y_Wumyq8QrH:bh01b5t)* status: 0 of 1 approvals obtained

PR #732, opened 2021-12-16
allow user to set a modelCurve as unselectable

andrewheumann:
BACKGROUND:
- In threejs, intersection seems to pretty aggressively prefer selecting model curves over meshes that occupy the same place. Sometimes we want model curves selectable, but other times we want to prevent them from being selected if they're only serving a diagrammatic purpose.

DESCRIPTION:
- Allows a user to call `SetSelectable(false)` on a ModelCurve object to alter its given name in the gltf (thereby preventing it from being selected in the UI)

TESTING:
- I published an `Unselectable Model Curves Test` function on dev. The green circle is selectable, and the red is not. 
  
FUTURE WORK:
- Other diagrammatic types might want to be made selectable/unselectable
- We might want to use a property other than the gltf object name to mark this, 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/732)


wynged (approved):
preciate this, will be helpful on some contract work I think

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/732#-:-Mr3e0YQB5f9vRYRPDbA:b-y4h6yw)* status: :shipit: complete! 1 of 1 approvals obtained

PR #733, opened 2021-12-16
Add a Box type, and expand BBox3

andrewheumann:
BACKGROUND:
- We already have a BBox3 world-axis-aligned bounding box type. But often for architectural use cases it's helpful to have a box that need not be aligned to the world coordinate system. 

DESCRIPTION:
**Adds the following APIs:**
- `Transform.Rotated()`
- `BBox3.PointAt`
- `BBox3.TransformAt`
- `BBox3.UVWAtPoint`
- `BBox3.XSize`, `BBox3.YSize`, `BBox3.ZSize`
- `BBox3.XDomain`, `BBox3.YDomain`, `BBox3.ZDomain`
- `Box` type, representing an oriented 3d box.
  - `Box.PointAt`
  - `Box.TransformAt`
  - `Box.UVWAtPoint`
  - `Box.UVWToBox`
  - `Box.BoxToUVW`
  - `Box.TransformBetween`

TESTING:
- A number of new tests have been added to `BoxTests` and `BBox3Tests`.
  
FUTURE WORK:
- There should be a "Box" input for functions that gives you an interactive box or boxes. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/733)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/BBox3.cs, line 280 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84RAy0S7IojWTc8CZ:-Mr84RAzBsD1yDQYwZHU:b-vfivkl) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L280)):*
> ```Smalltalk
>         /// <param name="uvw">The vector in the box's parametric UVW coordinate space.</param>
>         /// <returns>A point in world coordinates.</returns>
>         public Vector3 PointAt(Vector3 uvw)
> ```

TBH I don't get this, feels confusing to me to ask for a PointAt a Point ( I know it's a vec3 but...) and then get back a different point.  Is this useful?
___
*[Elements/src/Geometry/BBox3.cs, line 297 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84dxqBKtB-T-ighJM:-Mr84dxqBKtB-T-ighJN:bhrd68v) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L297)):*
> ```Smalltalk
>         public Transform TransformAt(double u, double v, double w)
>         {
>             return new Transform(PointAt(new Vector3(u, v, w)));
> ```

this makes a new vector that is not necessary given the `PointAt(u,v,w)` method above.
___
*[Elements/src/Geometry/Box.cs, line 18 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr85oyX7AY-ZyiCyYGy:-Mr85oyX7AY-ZyiCyYGz:bq67r8v) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L18)):*
> ```Smalltalk
>         /// <summary>
>         /// The Transform indicating the local coordinate frame of this box. Do
>         /// not modify this transform directly to modify the box, instead use TransformBox().
> ```

Given that Transform for all GeometryElements _is_ how you transform them, can we change the name of this variable?  `BoxOrientation`?
___
*[Elements/src/Geometry/Box.cs, line 181 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr86LWn3AzN_u7D2M6j:-Mr86LWn3AzN_u7D2M6k:b-1mlwba) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L181)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="transform">The transform to apply.</param>
>         public void TransformBox(Transform transform)
> ```

you say not to modify the Transform.. but isn't this method exactly what the a user would do if they did modify the transform directly?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Geometry/BBox3.cs, line 280 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84RAy0S7IojWTc8CZ:-Mr8oREt74XsPu4aEHD_:b2802em) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L280)):*

TBH I don't get this, feels confusing to me to ask for a PointAt a Point ( I know it's a vec3 but...) and then get back a different point.  Is this useful?
</blockquote></details>

Would it be clearer if the method took separate u,v,w arguments? Personally I prefer the Vector3 because one common use case would be to construct geometry in UVW space. I could draw a mesh in UVW space (say, a morphing facade panel), and just call `PointAt` on all the vertex positions to map it to the box's coordinate space. A broken-out-by-component constructor would add an unnecessary extra step. 
___
*[Elements/src/Geometry/BBox3.cs, line 297 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84dxqBKtB-T-ighJM:-Mr8qGUo9kQHSwhp2RGo:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L297)):*

this makes a new vector that is not necessary given the `PointAt(u,v,w)` method above.
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Box.cs, line 18 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr85oyX7AY-ZyiCyYGy:-Mr8ofr8A9H-u22rxrq6:b-8slyn7) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L18)):*

Given that Transform for all GeometryElements _is_ how you transform them, can we change the name of this variable?  `BoxOrientation`?
</blockquote></details>

I prefer to keep it as `Transform` since in the future it might handle more than orientation (shear, scaling, etc)
___
*[Elements/src/Geometry/Box.cs, line 181 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr86LWn3AzN_u7D2M6j:-Mr8oy4o4hKYLHKi20vM:bibou2k) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L181)):*

you say not to modify the Transform.. but isn't this method exactly what the a user would do if they did modify the transform directly?
</blockquote></details>

No, this calls the setter which also sets the inverse. If you modify the transform in place, the inverse won't update with it.

wynged (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733#-:-MrANSpx0rgs9DVyrXis:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/BBox3.cs, line 280 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84RAy0S7IojWTc8CZ:-MrAJBdX9txHk0ur8_nM:b-yhjmic) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L280)):*

Would it be clearer if the method took separate u,v,w arguments? Personally I prefer the Vector3 because one common use case would be to construct geometry in UVW space. I could draw a mesh in UVW space (say, a morphing facade panel), and just call `PointAt` on all the vertex positions to map it to the box's coordinate space. A broken-out-by-component constructor would add an unnecessary extra step. 
</blockquote></details>

You do already have the method with separate u,v,w arguments, which seems good.  Is this mapping to the boxes coordinate space the same as transforming the Mesh using this boxes Transform?  If both methods are valid, should we have both?
___
*[Elements/src/Geometry/Box.cs, line 18 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr85oyX7AY-ZyiCyYGy:-MrADUsV-PjyQ8DM4gFZ:b9i3knf) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L18)):*

I prefer to keep it as `Transform` since in the future it might handle more than orientation (shear, scaling, etc)
</blockquote></details>

ok, having it be a Transform makes sense, calling it Transform I think I get, although the Origin of this transform won't be the origin of the Box... ?  which is odd.  If the set behavior is important to use by the other method, can we make it a `private set`, or maybe just remove the set and have this logic in the other method?
___
*[Elements/src/Geometry/Box.cs, line 181 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr86LWn3AzN_u7D2M6j:-MrAI5Vt-y6Iae9VN46a:b4iatsy) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L181)):*

No, this calls the setter which also sets the inverse. If you modify the transform in place, the inverse won't update with it. 
</blockquote></details>

ahh, ok.

wynged (changes requested):
note, comment is below, approved by accident.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733#-:-MrAbD33EWcsfFkOErNV:b2mfqp0)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Geometry/BBox3.cs, line 280 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84RAy0S7IojWTc8CZ:-MrDSSxS22TrV1DpcCfa:b-k4r6qq) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L280)):*

You do already have the method with separate u,v,w arguments, which seems good.  Is this mapping to the boxes coordinate space the same as transforming the Mesh using this boxes Transform?  If both methods are valid, should we have both?
</blockquote></details>

This is the Bbox3, method, so there's no relevant transform. In the equivalent method in `Box`, it's still not the same as applying the box's transform. It would be equivalent to applying the box's `UVWToBox` transform.

___
*[Elements/src/Geometry/Box.cs, line 18 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr85oyX7AY-ZyiCyYGy:-MrDTIJdB9zGnYjt-HBD:b-1um5bh) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L18)):*

ok, having it be a Transform makes sense, calling it Transform I think I get, although the Origin of this transform won't be the origin of the Box... ?  which is odd.  If the set behavior is important to use by the other method, can we make it a `private set`, or maybe just remove the set and have this logic in the other method?
</blockquote></details>

It's totally valid to `set` the Transform on the box, so I'd prefer to leave it public. It's just not valid to _mutate_ the transform. so I can't do `box.Transform.Rotate(45)` but I could do `box.Transform = box.Transform.Rotated(45)`. 

And no, the transform's origin is not the origin of the box, it's the origin of its coordinate frame.  So if you want a 10x10x10 box centered on a certain transform, you can construct it with a min of -5,-5,-5 and a max of 5,5,5.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/733)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[Elements/src/Geometry/BBox3.cs, line 280 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr84RAy0S7IojWTc8CZ:-MrNI7uH265AU5hTkSqh:b3gaspv) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/BBox3.cs#L280)):*

This is the Bbox3, method, so there's no relevant transform. In the equivalent method in `Box`, it's still not the same as applying the box's transform. It would be equivalent to applying the box's `UVWToBox` transform.


</blockquote></details>

ahhh, UVW space is the domain space, I'm on board now.  Thanks :)
___
*[Elements/src/Geometry/Box.cs, line 18 at r1](https://reviewable.io/reviews/hypar-io/elements/733#-Mr85oyX7AY-ZyiCyYGy:-MrNIu4o1AP-SjTBcwCV:b-56ck8a) ([raw file](https://github.com/hypar-io/elements/blob/105ddc04facfc363df548767e87ed9c13200a1cd/Elements/src/Geometry/Box.cs#L18)):*

It's totally valid to `set` the Transform on the box, so I'd prefer to leave it public. It's just not valid to _mutate_ the transform. so I can't do `box.Transform.Rotate(45)` but I could do `box.Transform = box.Transform.Rotated(45)`. 

And no, the transform's origin is not the origin of the box, it's the origin of its coordinate frame.  So if you want a 10x10x10 box centered on a certain transform, you can construct it with a min of -5,-5,-5 and a max of 5,5,5.
</blockquote></details>

ahhh, the issue is that transforms are mutable and we have no way of knowing when we need to update `_inverseTransform`.  Well, we'll just have to see if this ever bites us, I don't have any ideas for protecting against that, or people thinking the Origin of Transform is the Origin of the Box.

PR #738, opened 2021-12-22
Updated condition of the Traverse function exit;

katehryhorenko:
BACKGROUND:
Network Traverse function can loop in some cases. For example, if leaf is very close to the edge, so for all calculations it's on the edge, but Network is still "think" that it's a leaf (it's smth we need to fix). So, when we are trying to travers network CCW, we got loop:

image

Because Leaf is almost on nearest edge, algorithm cannot use it and finds other edge that can be CCW. As result in a few step we get loop.

DESCRIPTION:
This PR contains condition inside Traverse  method: exit if we have already passed two elements in the same order

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/738)

### DIFF
--- a/Elements/src/Search/Network.cs
+++ b/Elements/src/Search/Network.cs
@@ -396,6 +396,15 @@ namespace Elements.Search
                 {
                     break;
                 }
+
+                if (path.Contains(currentIndex))
+                {
+                    // if we have already passed two elements in the same order, we've achieved a loop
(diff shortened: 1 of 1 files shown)


ikeough (approved):

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/738)* status: :shipit: complete! 1 of 1 approvals obtained

PR #741, opened 2021-12-24
Discriminator warnings

wynged:
BACKGROUND:
- Actually throwing errors during deserialization can make it very hard to debug an issue if the is a deserialization issue, even if that deserialization issue can be safely ignored.  We eventually handle these exceptions anyways and surface messages, so this simply does that explicitly.
- This really only affects working with local Elements, but this is frequent enough to be annoying.

DESCRIPTION:
- Create a _deserializationDiscriminatorWarnings list that saves a list of deserialization issues encountered, rather than throwing for each one encountered.
- Remove top level nulls from the model at the end of the deserialization

TESTING:
- If you debug the test `SkipsUnknownTypesDuringDeserialization` in the master branch you will hit an error inside `FromJson` even though the test succeeds and there are not . This kind of error could occur dozens of times if there is an unknown type, and so you have to step through all of these errors to get to the error you're actually interested in.  Now, in this branch, if you debug the test you won't hit any errors.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/741)


ikeough (changes requested):
I'm slightly skeptical of creating null elements, just to loop through and 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/741#-:-MskipD0DO0r2zzV1OHv:b-2lm16b)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[Elements/src/Model.cs, line 241 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskiwijBn-dkKR3VP1j:-MskiwijBn-dkKR3VP1k:b-lps74d) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Model.cs#L241)):*
> ```Smalltalk
>             JsonInheritanceConverter.Elements.Clear();
>             // Don't leave null elements in the model.
>             foreach (var e in model.Elements.Where(e => e.Value == null).ToList())
> ```

I'm slightly skeptical of adding nulls just to remove them again. The `Where`, followed by the `Remove` (and the resizing of the collection) is work that the computer doesn't need to do if we can handle this in another way. 

Instead of doing all of this, can we just edit the error handling delegate we already have on the serializer to skip logging these errors? Edit `Model.ToJson(...logWarnings = true)`, and use that parameter to optionally skip logging warnings. Or, throw a specific type of exception and look for that, like a `KeyNotFound` exception.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 233 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskhQ-I3CqZJVdx3EwV:-MskhQ-I3CqZJVdx3EwW:bxjzt9l) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L233)):*
> ```Smalltalk
>                 if (!Elements.ContainsKey(id))
>                 {
>                     _deserializationDiscriminatorWarnings.Add($"Expected to find an element with id {id} but it was not found during deserialization.  Look for other deserialization errors.");
> ```

This was a little confusing to me here because it's called `_deserializationDiscriminatorWarnings` and it has nothing to do with discriminator errors. Can we rename the field to `_deserializationWarnings` because it now contains more than one type of warning.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 233 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskhiX75QsD-c2iOgyV:-MskhiX75QsD-c2iOgyW:b-rm5i5f) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L233)):*
> ```Smalltalk
>                 if (!Elements.ContainsKey(id))
>                 {
>                     _deserializationDiscriminatorWarnings.Add($"Expected to find an element with id {id} but it was not found during deserialization.  Look for other deserialization errors.");
> ```

Recommend: `Element {id} was not found during deserialization. Check for other deserialization errors.` This is a bit more terse.

wynged (commented):
So, the logging of warnings is not the concern, I'm ambivalent about whether we gather a list of errors.   It's the actual `throw` down below when a discriminator was found but the element type was not.  That `throw` makes live very annoying if you are working with local elements and you have a model you are de-serializing where you might not have all types.  When you go to debug the IDE gets caught on all of these throws before you can get to the actual issue you are trying to debug.  This occurs semi-often during function development and a user recently commented on this same issue, so I'm re-invigorating this PR.

So, if we don't want to `throw` then we have to return something, because `ReadJson()` returns `object`, hence the nulls.

To test the performance hit I ran 3 benchmarks.

benchmarking removing nulls vs errors.  First runs are 1000 invalid elements, second run is 1 million invalid elements.

current error strategy


with the strategy previously in the PR, has Dictionary resizing concern.


changed approach to dictionary pre-allocations hoping for speedup (not faster it turns out)


> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/741#-:-N-f1PvkB-FRrKRRhfiP:b-xlu21y)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Model.cs, line 241 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskiwijBn-dkKR3VP1j:-N-fKYGK4wCrGpvfdzUb:b8tw6un) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Model.cs#L241)):*

I'm slightly skeptical of adding nulls just to remove them again. The `Where`, followed by the `Remove` (and the resizing of the collection) is work that the computer doesn't need to do if we can handle this in another way. 

Instead of doing all of this, can we just edit the error handling delegate we already have on the serializer to skip logging these errors? Edit `Model.ToJson(...logWarnings = true)`, and use that parameter to optionally skip logging warnings. Or, throw a specific type of exception and look for that, like a `KeyNotFound` exception.
</blockquote></details>

responded about general situation above.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 233 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskhQ-I3CqZJVdx3EwV:-N-epvQC5yHg3_5syYvw:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L233)):*

This was a little confusing to me here because it's called `_deserializationDiscriminatorWarnings` and it has nothing to do with discriminator errors. Can we rename the field to `_deserializationWarnings` because it now contains more than one type of warning.
</blockquote></details>

Done.
___
*[Elements/src/Serialization/JSON/JsonInheritanceConverter.cs, line 233 at r1](https://reviewable.io/reviews/hypar-io/elements/741#-MskhiX75QsD-c2iOgyV:-N-eq1_E0L-oYsmBFE4_:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/7da195170dfd135f6305d1e9b90420e9cc93cd9f/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L233)):*

Recommend: `Element {id} was not found during deserialization. Check for other deserialization errors.` This is a bit more terse. 
</blockquote></details>

Done.

ikeough:
Sorry I'm just realizing that this has been open for so long @wynged. In a way that's a good thing because I've been working on #776. `System.Text.Json` doesn't even handle deserialization warnings so we have to return nulls there. I'm wondering if we can do better than nulls here _and_ solve for that case, by creating a type like `UnknownElement` which has a `discriminator` property that is set to the value of the deserialized discriminator, and its id is set to the id from the JSON, and maybe its properties are all stowed in `AdditionalProperties`. We'd get all the elements, whether we had a corresponding type or not, and we could avoid throwing exceptions. What do you think?

wynged (commented):
That sounds... pretty sweet?  it would help preserve references for loaded items as well, which would be a nice win.  UnknownElement would also be a place to provide some nice helper methods for dealing with objects that are basically a bag of AdditionalProperties.  I think I might just close this PR, and we can implement something liek that when 776 is in.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/741#-:-N8AU4DPAVbuXFwYxN2t:bzcppbx)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

PR #742, opened 2021-12-28
Skip CSG path fully.

ikeough:
# Requires #743

BACKGROUND:
While working on running Elements in web assembly I wanted to make things run as fast as possible so I used the `SkipCSGUnion` flag. It didn't have the increased performance which one would expect when not passing through `Csg`. It turns out we were converting solids to `Csg` even when CSG operations were not required.

DESCRIPTION:
This PR does the following:
- It make the path hit when using `SkipCsgUnion` avoid `Csg` entirely.
- It unifies tessellation into a new static class `Tessellation` with a number of adapter interfaces and classes to handle tessellation of `Csg` and `Solid` types.

REQUIRED:
- [ ] All changes are up to date in `CHANGELOG.md`. **n/a** (all new methods, classes and interfaces are internal)

The following benchmarks both run the `TraceGltfSerialization` benchmark, which uses the `SkipCsgUnion` flag on the creation of all beams. The benchmarks show that this path is now **30% faster** when creating, tessellating and saving a glTF of all 141 HSS beam types.

Before:
|                                        Method |     Mean |   Error |  StdDev |
|---------------------------------------------- |---------:|--------:|--------:|
| 'Create all HSS beams and serialize to glTF.' | 103.0 ms | 1.65 ms | 2.62 ms |

After:
|                                        Method |     Mean |    Error |   StdDev |
|---------------------------------------------- |---------:|---------:|---------:|
| 'Create all HSS beams and serialize to glTF.' | **71.58 ms** | 1.386 ms | 1.897 ms |

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/742)


mattc (approved):
> Reviewed 3 of 11 files at r1, 13 of 13 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/742)* status: :shipit: complete! 1 of 1 approvals obtained

PR #744, opened 2021-12-30
Return list of errors instead of function failure

katehryhorenko:
BACKGROUND:
ToGltf method always fails if one of the elements has error inside. It can be some geometric condition that we do not handle. 

DESCRIPTION:
This PR contains try-catch around  GetRenderDataForElement method and output list with error messages

### DIFF
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -73,15 +73,17 @@ namespace Elements.Serialization.glTF
         /// </summary>
         /// <param name="model">The model to serialize.</param>
         /// <param name="path">The output path.</param>
+        /// <param name="errors">A collection of serialization errors</param>
         /// <param name="useBinarySerialization">Should binary serialization be used?</param>
         /// <param name="drawEdges">Should the solid edges be written to the gltf?</param>
-        public static void ToGlTF(this Model model, string path, bool useBinarySerialization = true, bool drawEdges = false)
+        public static void ToGlTF(this Model model, string path, out List<string> errors, bool useBinarySerialization = true, bool drawEdges = false)
         {
+            errors = new List<string>();
             if (model.Elements.Count > 0)
             {
                 if (useBinarySerialization)
                 {
-                    if (SaveGlb(model, path, drawEdges))
+                    if (SaveGlb(model, path, out errors, drawEdges))
                     {
                         return;
                     }
@@ -89,7 +91,7 @@ namespace Elements.Serialization.glTF
                 }
                 else
                 {
-                    if (SaveGltf(model, path, drawEdges))
+                    if (SaveGltf(model, path, out errors, drawEdges))
                     {
                         return;
                     }
@@ -102,6 +104,19 @@ namespace Elements.Serialization.glTF
         }
 
         /// <summary>
+        /// Serialize the model to a gltf file on disk.
+        /// If there is no geometry, an empty GLTF will still be produced.
+        /// </summary>
+        /// <param name="model">The model to serialize.</param>
+        /// <param name="path">The output path.</param>
+        /// <param name="useBinarySerialization">Should binary serialization be used?</param>
+        /// <param name="drawEdges">Should the solid edges be written to the gltf?</param>
+        public static void ToGlTF(this Model model, string path, bool useBinarySerialization = true, bool drawEdges = false)
+        {
+            ToGlTF(model, path, out _, useBinarySerialization, drawEdges);
+        }
+
+        /// <summary>
         /// Serialize the model to a byte array.
         /// </summary>
         /// <param name="model">The model to serialize.</param>
@@ -110,7 +125,7 @@ namespace Elements.Serialization.glTF
         /// <returns>A byte array representing the model.</returns>
         public static byte[] ToGlTF(this Model model, bool drawEdges = false, bool mergeVertices = false)
         {
-            var gltf = InitializeGlTF(model, out var buffers, drawEdges, mergeVertices);
+            var gltf = InitializeGlTF(model, out var buffers, out var errors, drawEdges, mergeVertices);
             if (gltf == null)
             {
                 return null;
@@ -135,7 +150,7 @@ namespace Elements.Serialization.glTF
         public static string ToBase64String(this Model model, bool drawEdges = false, bool mergeVertices = false)
         {
             var tmp = Path.GetTempFileName();
-            var gltf = InitializeGlTF(model, out var buffers, drawEdges, mergeVertices);
+            var gltf = InitializeGlTF(model, out var buffers, out var errors, drawEdges, mergeVertices);
             if (gltf == null)
             {
                 return "";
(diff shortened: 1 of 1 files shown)


andrewheumann (changes requested):
This is a great and welcome change — I've tested on a function and it works well. 

My only slight concern is that because you have to explicitly go looking for this error, a function author won't necessarily know that anything has gone wrong — they'll just be stuck wondering why an element is missing. Here are some ideas (which can all be Future Work — you don't need to pursue this right now!) for ways to alleviate this:
- update `hypar run` infrastructure to automatically log these errors
- update `hypar test generate` to generate a test that logs these errors prominently (to console or a text file)
- update normal function execution pathways to include these errors (perhaps by appending them to the output's `warnings` or `errors`, or creating a new top-level `elementErrors` key to contain these) so that we might surface them to the UI at some point. 


It would also be great to add a test to `GltfTests` that would fail prior to this change. (one way I found to trigger this is by creating a `GeometricElement`, and then scaling its representation's local transform to 0:
```
extrude.LocalTransform = new Transform().Scaled(new Vector3());
``` 
) - but feel free to invent any method that might have previously triggered an exception that this change now catches. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744#-:-MsBy6CI9740z5tj3_T_:b-u1fbqe)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 903 at r1](https://reviewable.io/reviews/hypar-io/elements/744#-MsBxZUa4VZ3lvlp0QUw:-MsBxZUa4VZ3lvlp0QUx:bhbanzq) ([raw file](https://github.com/hypar-io/elements/blob/642c336ebada676e35bb8c5f7ccea9bf9ccc73cb/Elements/src/Serialization/glTF/GltfExtensions.cs#L903)):*
> ```Smalltalk
>                 catch (Exception ex)
>                 {
>                     errors.Add($"{e.Id} : {ex.Message}");
> ```

Can we add more detail here? the `Message` alone may not be super helpful in debugging an issue — a stack trace might be helpful too, even if it won't point to a user's own code.

katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 903 at r1](https://reviewable.io/reviews/hypar-io/elements/744#-MsBxZUa4VZ3lvlp0QUw:-MsCgtQvDU_OVXo9Qc8Y:bi85ek9) ([raw file](https://github.com/hypar-io/elements/blob/642c336ebada676e35bb8c5f7ccea9bf9ccc73cb/Elements/src/Serialization/glTF/GltfExtensions.cs#L903)):*

Can we add more detail here? the `Message` alone may not be super helpful in debugging an issue — a stack trace might be helpful too, even if it won't point to a user's own code. 
</blockquote></details>

I replaced string with ElementError that contains message, stack trace and element id

katehryhorenko (commented):
I added one test that expects error element. Agree with all items in TODO list. It would be great log such errors and see them in UI if they appear during/after function execution

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744#-:-MsCh7rK2nkhLhzUW5Pn:bs3kk66)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (changes requested):
> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 128 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlAhnAuI4EhhdiiPJ:-MsFlAhnAuI4EhhdiiPK:bpcg6vx) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/src/Serialization/glTF/GltfExtensions.cs#L128)):*
> ```Smalltalk
>         public static byte[] ToGlTF(this Model model, bool drawEdges = false, bool mergeVertices = false)
>         {
>             var gltf = InitializeGlTF(model, out var buffers, out var errors, drawEdges, mergeVertices);
> ```

Since we're just discarding this, I think it can be `out _`
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 153 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlCdR1vQ7CPgcfpvH:-MsFlCdR1vQ7CPgcfpvI:bnj7hl0) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/src/Serialization/glTF/GltfExtensions.cs#L153)):*
> ```Smalltalk
>         {
>             var tmp = Path.GetTempFileName();
>             var gltf = InitializeGlTF(model, out var buffers, out var errors, drawEdges, mergeVertices);
> ```

same as above
___
*[Elements/test/GltfTests.cs, line 112 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlvRu6QAyqvt5rVo-:-MsFlvRu6QAyqvt5rVo0:b-d8e8ma) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/test/GltfTests.cs#L112)):*
> ```Smalltalk
>             model.AddElement(beam);
>             var modelsDir = Path.Combine(Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location), "models");
>             var gltfPath = Path.Combine(modelsDir, "Beam-with-errro.gltf");
> ```

`Beam-with-errro.gltf` => Beam-with-error.gltf

katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 128 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlAhnAuI4EhhdiiPJ:-MsW0wRq4mRkij_mMGnY:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/src/Serialization/glTF/GltfExtensions.cs#L128)):*

Since we're just discarding this, I think it can be `out _`
</blockquote></details>

Done.
___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 153 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlCdR1vQ7CPgcfpvH:-MsW0wfd2G8Csft4szpK:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/src/Serialization/glTF/GltfExtensions.cs#L153)):*

same as above
</blockquote></details>

Done.
___
*[Elements/test/GltfTests.cs, line 112 at r2](https://reviewable.io/reviews/hypar-io/elements/744#-MsFlvRu6QAyqvt5rVo-:-MsW0ygtEv2UKrGgsSea:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ee868849c6d9bfd383a5008aae46c896d76e8596/Elements/test/GltfTests.cs#L112)):*

`Beam-with-errro.gltf` => Beam-with-error.gltf
</blockquote></details>

Done.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/744)* status: :shipit: complete! 1 of 1 approvals obtained

PR #745, opened 2022-01-02
Add notebook helpers for elements

andrewheumann:
BACKGROUND:
- Dotnet Interactive Notebooks support [extensions configured as scripts](https://github.com/dotnet/interactive/blob/788323abc710272a6fbd6bd9f51e3958e408b71a/docs/extending-dotnet-interactive.md#script-based-extensions). This permits adding special notebook-only functionality, without adding any dependencies on notebook-specific packages to the core library.

DESCRIPTION:
- Adds a helper script that runs when `Hypar.Elements` is imported into a dotnet interactive notebook. This helper creates a formatter for the `Model` type which renders a small threejs window to display a model, so that in code you can just `return model;` and see the model in 3d. It also adds a global `DisplayModel` method which can take a variable width and height.

TESTING:
- I temporarily renamed the package id in the csproj to "Hypar.ElementsTest", built the package locally using `dotnet build -c release`, and then referenced the local package source in a notebook like so:
```
#i "nuget: file:///Users/andrewheumann/Dev/elements/nupkg"
#r "nuget: Hypar.ElementsTest"
```

https://user-images.githubusercontent.com/31935763/147881667-284f27ed-26b0-4e2f-9f23-d1926ae4a051.mp4

FUTURE WORK:
- I'm not sure the helpful message I'm trying to render on load is working correctly. This might be a bug or a local package limitation, or maybe I've made a mistake somewhere.
- It would be nice not to have to bundle the JS inline, but I had trouble getting it to import from an externally hosted script. Probably I'm just missing something obvious.

REQUIRED:
- [N/A] All changes are up to date in `CHANGELOG.md`. (not gonna publicize this until we're sure it works)

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/745)

### DIFF
--- a/Elements/src/extension.dib
+++ b/Elements/src/extension.dib
@@ -1,0 +1,142 @@
+#!markdown
+
+This script is packaged with the NuGet package, and only
+called when the package is imported in a dotnet interactive
+notebook via:
+
+`#r "nuget: Hypar.Elements"`
+
+#!csharp
+
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using Microsoft.DotNet.Interactive;
+using Microsoft.DotNet.Interactive.Formatting;
+using Elements;
+using Elements.Geometry;
+using Elements.Serialization.glTF;
+
+var viewerSrc = @"
+<div id=""main_DIV_ID"" style=""height:400px;width:400px;""></div>
+</div>
+<script type=""module"">
+import * as THREE from 'https://unpkg.com/three@0.126.0/build/three.module.js';
+import { GLTFLoader } from 'https://unpkg.com/three@0.126.0/examples/jsm/loaders/GLTFLoader.js';
+import { OrbitControls } from 'https://unpkg.com/three@0.126.0/examples/jsm/controls/OrbitControls.js';
+const loader = new GLTFLoader();
+const scene = new THREE.Scene();
+
+let loaderScene = new THREE.Scene();
+scene.add(loaderScene);
+
+const width = WIDTH_VAR;
+const height = HEIGHT_VAR;
+const camera = new THREE.PerspectiveCamera(75, width / height, 0.1, 1000);
+const light = new THREE.AmbientLight(0xaaaaaa);
+scene.add(light);
+
+const directionalLight = new THREE.DirectionalLight(0xaaaaaa, 0.5);
+scene.add(directionalLight);
+
+
+const renderer = new THREE.WebGLRenderer();
+renderer.setSize(width, height);
+document.getElementById('main_DIV_ID').appendChild(renderer.domElement);
+
+camera.position.z = 5;
+
+const controls = new OrbitControls(camera, renderer.domElement);
+
+function base64ToArrayBuffer (base64) {
+    var binary_string = window.atob(base64);
+    var len = binary_string.length;
+    var bytes = new Uint8Array(len);
+    for (var i = 0; i < len; i++) {
+        bytes[i] = binary_string.charCodeAt(i);
+    }
+    return bytes.buffer;
+}
+// adapted from looeee's solution https://discourse.threejs.org/t/camera-zoom-to-fit-object/936/3
+const fitCameraToObject = function (scene, offset = 1.25) {
+    offset = offset || 1.25;
+    let boundingBox = null;
+    scene.traverseVisible((child) => {
+        const objectBox = new THREE.Box3().setFromObject(child)
+        if (boundingBox === null) {
+            boundingBox = objectBox
+        } else {
+            boundingBox = boundingBox.union(objectBox)
+        }
+    })
+    const center = boundingBox.getCenter(new THREE.Vector3());
+    const size = boundingBox.getSize(new THREE.Vector3());
+
+    // get the max side of the bounding box (fits to width OR height as needed )
+    const maxDim = Math.max(size.x, size.y, size.z);
+    const fov = camera.fov * (Math.PI / 180);
+    let cameraZ = Math.abs(maxDim / 4 * Math.tan(fov * 2));
+
+    camera.position.copy(center.clone().add(size.clone().multiplyScalar(offset)))
+
+    const minZ = boundingBox.min.z;
+    const cameraToFarEdge = (minZ < 0) ? -minZ + cameraZ : cameraZ - minZ;
+
+    camera.far = cameraToFarEdge * 3;
+    camera.updateProjectionMatrix();
+
+    if (controls) {
+        // set camera to rotate around center of loaded object
+        controls.target = center;
+        // prevent camera from zooming out far enough to create far plane cutoff
+        controls.maxDistance = cameraToFarEdge * 2;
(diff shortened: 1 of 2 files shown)


Elements/src/extension.dib
```diff
@@ -0,0 +1,142 @@
...
+
+
+animate();
+
+</script>
+";
+
+if (KernelInvocationContext.Current is { } currentContext)
+{
+    currentContext.DisplayAs("Just add `return model;` at the end of a cell, or call `DisplayModel(model, width, height)` to display an Elements model.", "text/markdown");
```
ikeough, line 120: Recommend "Add return model..." The "just" makes this sound easy and might frustrate the user when it doesn't _just_ work

Elements/src/extension.dib
```diff
@@ -0,0 +1,142 @@
...
+const loader = new GLTFLoader();
+const scene = new THREE.Scene();
+
+let loaderScene = new THREE.Scene();
+scene.add(loaderScene);
+
+const width = WIDTH_VAR;
+const height = HEIGHT_VAR;
+const camera = new THREE.PerspectiveCamera(75, width / height, 0.1, 1000);
+const light = new THREE.AmbientLight(0xaaaaaa);
```
ikeough, line 35: I would like, if possible, to replicate the lighting/shading setup that's on Hypar currently to minimize surprises. I'm ok with this as v1.

Elements/src/extension.dib
```diff
@@ -0,0 +1,142 @@
...
+
+const renderer = new THREE.WebGLRenderer();
+renderer.setSize(width, height);
+document.getElementById('main_DIV_ID').appendChild(renderer.domElement);
+
+camera.position.z = 5;
+
+const controls = new OrbitControls(camera, renderer.domElement);
+
+function base64ToArrayBuffer (base64) {
```
ikeough, line 50: I assume that this integration doesn't have, like wasm, an "unmarshalled" mode where we can push a byte array instead of serializing to a string?

ikeough (approved):
A couple of very minor questions and suggestions.

Elements/src/extension.dib
andrewheumann: Will leave a TODO!

Elements/src/extension.dib
andrewheumann: Not as far as I know. We're not even utilizing notebook variable sharing — we're just rendering raw HTML, so for now it's probably gotta be a string.

Elements/src/extension.dib
andrewheumann: Done.

Elements/src/extension.dib
```diff
@@ -0,0 +1,143 @@
...
+
+if (KernelInvocationContext.Current is { } currentContext)
+{
+    currentContext.DisplayAs("Add `return model;` at the end of a cell or call `DisplayModel(model, width, height);` to display an Elements model.", "text/markdown");
+}
+
+string GetModelViewerSrc(Model model, double width=400, double height=400) {
+    var gltf = model.ToGlTF();
+    var gltfString = Convert.ToBase64String(gltf, 0, gltf.Length);
+    return viewerSrc
```
iainsproat, line 127: Could these C# kernel variables could be accessed directly from within the javascript? https://github.com/dotnet/interactive/blob/main/docs/javascript-overview.md#accessing-kernel-data-from-client-side-javascript-code

iainsproat:
Hi @andrewheumann - really nice idea to implement this 😄 ! 

I was wondering if there was a particular reason you went for a `.dib` file over doing this via a project where [the static class implements `IKernelExtension`](https://github.com/dotnet/interactive/blob/788323abc710272a6fbd6bd9f51e3958e408b71a/samples/extensions/ClockExtension/ClockKernelExtension.cs#L16)?

This might solve the issue with the inline javascript; you could move the javascript to an embedded resource file within the [compiled dll copied to `interactive-extensions/dotnet`](https://github.com/dotnet/interactive/blob/788323abc710272a6fbd6bd9f51e3958e408b71a/samples/extensions/ClockExtension/ClockExtension.csproj#L23).  The embedded resource file could be read locally using [`GetManifestResourceStream`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?view=net-6.0#System_Reflection_Assembly_GetManifestResourceStream_System_String_). It should be faster and more reliable than importing it from an external location.

andrewheumann:
@iainsproat Thanks for the comment! I went with the `.dib` approach primarily because it allows you to include notebook-specific functionality _without_ adding a package dependency on `dotnet.interactive` libraries — I didn't want to bloat Elements with stuff most clients don't need.

I suppose we could also produce a second `Hypar.Elements.NotebookExtensions` NuGet package or similar to take advantage of `IKernelExtension` — but I also shied away from this approach for a second reason, which is that the `examples` in the dotnet interactive nuget for this simply aren't working for me — I tried to build the `ClockExtension` but it ran into some version mismatch issue so I gave up and went a different way. 

As for the inline javascript, the direction I imagine going is just hosting a `.js` file elsewhere and using it as `<script src="blah">`, rather than having it live in the package at all. That way we can make improvements to the viewer (lighting, antialiasing, interaction, whatever) without having to publish a new version of Elements. As you point out, there's a risk of a performance/reliability hit going this route, but it seems more flexible to me.

Elements/src/extension.dib
andrewheumann: Yes, this works well, and was actually the first version of this I tried:

https://user-images.githubusercontent.com/31935763/147887266-754e461c-4c0b-4485-b76e-be2aa59d0f9a.mp4

Mutating the script string with `Replace` is definitely janky, and I don't think it's the right long-term solution, even if it works fine for right now. But in the end I went this route because I couldn't figure out how to do variable sharing with javascript-in-html generated by the extension... I'm sure there's a way, just haven't gotten that far yet!

andrewheumann: (The reason I went with an all `#!html` cell instead of a mixture of `#!html` and `#!javascript, which would have enabled variable sharing, is just because I couldn't for the life of me get three.js to load via the requirejs pathway for dependencies in a `#!javascript` cell. Bringing it in via `<script type="module">` in the HTML was the only way I could get threejs going. Again, I'm sure there's a way!)

iainsproat:
Makes sense, thanks for the fast response. The `.dib` approach is certaintly simpler.

For external javascript files, did you try loading them via [`interactive.configureRequire`](https://github.com/dotnet/interactive/blob/main/docs/javascript-overview.md#loading-external-javascript-modules-at-runtime)?

andrewheumann:
I did! the `plotly` example worked fine, but when I tried to do it with three.js I just couldn't crack it. I'm not super familiar w/ requirejs (and there was a lot of old/conflicting info out there — to `shim` or not to `shim`, etc) — never got it working. If you know of a minimal example of `interactive.configureRequire` in a notebook working with an import of three.js, I'd love to see it!

PR #755, opened 2022-01-11
Suppress 612 and 618 warnings for Wall properties

iainsproat:
BACKGROUND:
When running `dotnet test`, CS0618 warnings were generated for use of Obsolete Wall.Profile and Wall.Height properties; e.g.:
```
Elements\src\WallByProfile.cs(93,13): warning CS0618: 'Wall.Profile' is obsolete: 'The profile property on the Wall base class is obsolete, check the methods of an inherited class like StandardWall or WallByProfile.'
```

DESCRIPTION:
This suppresses the warning without fixing the underlying cause i.e. the obsolete properties continue to be used.

TESTING:
- The warnings, described above, are no longer present when running `dotnet test`
  
FUTURE WORK:
- It may be possible to remove the root cause of the warnings (i.e. not to use the obsolete properties) rather than suppressing them.

REQUIRED:
    - Change log is assumed not required as this does not change behaviour for the end-user.
   
COMMENTS:
- I am not certain whether the code style of inline disable and enabling pragma directives are correct; elsewhere the pragma is at the top of the file and disables the warning for the entire file.
- The inline pragma warnings are not indented; this is to highlight their presence but may not be compatible with style guidelines.

I'd welcome comments as to whether this PR is the correct approach for this.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/755)

### DIFF
--- a/Elements/src/Wall.cs
+++ b/Elements/src/Wall.cs
@@ -59,9 +59,10 @@ namespace Elements
             {
                 throw new ArgumentOutOfRangeException("The wall could not be created. The height of the wall must be greater than 0.0.");
             }
-
+#pragma warning disable 612, 618
             this.Profile = profile;
             this.Height = height;
+#pragma warning restore 612, 618
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


andrewheumann (approved):

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/755)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #462, opened 2020-12-08
Bezier.Length() is wrong


andrewheumann:
**Describe the bug**
Bezier.Length() reports a value that is very off. 

**To Reproduce**
Compare Bezier.Length() to Bezier.ToPolyline(100).Length() — the difference is for me nearly a factor of 2.

Gytaco:
I'm working on a bezier thing atm I assume no updates?

iainsproat:
I've attempted a fix at this bug: https://github.com/hypar-io/Elements/pull/757

While my PR fixes and adds tests for the existing algorithm, the existing algorithm (linear interpolation) seems relatively inaccurate and this may need further work.  For example, the following results in a length of `9.8` and not the expected `10.0`, an (in)accuracy of >2%.

```
[Fact]
public void Bezier_Length_StraightLine()
{
    var a = Vector3.Origin;
    var b = new Vector3(5.0, 0.0);
    var c = new Vector3(10.0, 0.0);
    var ctrlPts = new List<Vector3>{a,b,c};
    var bezier = new Bezier(ctrlPts);

    var expectedLength = 10.0;
    Assert.Equal(expectedLength, bezier.Length(), 1);
    var divisions = 50;
    var polylineLength = bezier.ToPolyline(divisions).Length();
    Assert.Equal(polylineLength, bezier.Length());
}
```

ikeough:
We can improve that slightly by increasing the number of samples, currently set at 50. But that's only going to get slower. There are other methods that we could try: https://stackoverflow.com/questions/29438398/cheap-way-of-calculating-cubic-bezier-length.

PR #757, opened 2022-01-13
Fix `Bezier.Length()`

iainsproat:
BACKGROUND:
Fixes https://github.com/hypar-io/Elements/issues/462

DESCRIPTION:
When calculating the length of the first segment on the bezier curve, it erroneously used the origin as the start of the curve.  This PR corrects this to use the start point of the bezier curve.

TESTING:
Tests included.
- bezier.Length() should now match polyline.Length() where the number of divisions matches the bezier samples.
  
FUTURE WORK:
- The linear interpolation method is very approximate.

REQUIRED:

COMMENTS:
- tests make allowance for the approximate method of calculating length.  This is noted as an inline comment within the test code.
- tests comparing bezier.Length() to polyline.Length() may be brittle to changes in samples constant within Bezier.  This is noted as an inline comment within the test code.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/757)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- Under some circumstances `Bezier.Length()` would return incorrect results
 
 ## 0.9.8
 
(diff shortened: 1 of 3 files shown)


ikeough:
🥇

PR #764, opened 2022-01-27
add a more default model.tojson

wynged:
BACKGROUND:
- Functions published with latest alphas broke because the function signature of ToJson changed.
  - image
- the Hypar repo was not updated appropriately 
- Breakage occurred even though signature introduced only arguments with defaults, which should be optional, and the only place it is used used passes no arguments.

DESCRIPTION:
- Add an explicit `ToJson()` method that has no arguments for future compatibility.

TESTING:
- publishing a function with latest alphas will result in error during function execution.
  - `dotnet add package Hypar.Elements --version 0.9.9-alpha.7`
  - `dotnet add package Hypar.Functions --version 0.9.11-alpha.3`
  
FUTURE WORK:
- We will update the Hypar.Functions project to reference the latest alpha regardless of the effect of this fix.

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/764)

### DIFF
--- a/Elements/src/Model.cs
+++ b/Elements/src/Model.cs
@@ -201,11 +201,20 @@ namespace Elements
         }
 
         /// <summary>
-        /// Serialize the model to JSON using default arguments
+        /// Serialize the model to JSON using default arguments.
         /// </summary>
         public string ToJson()
         {
             return ToJson();
+        }
+
+        /// <summary>
+        /// Serialize the model to JSON to match default arguments.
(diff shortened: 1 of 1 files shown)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/764)* status: :shipit: complete! 1 of 1 approvals obtained

PR #765, opened 2022-01-27
stop using containsKey because the newtonsoft in revit doesn't have t…

wynged:
…his method

BACKGROUND:
- Catalog deserialization was failing in Revit 2020 due to `JObject.ContainsKey` not existing

DESCRIPTION:
- Uses an older method that seems to exist just fine in Revit.

TESTING:
- save catalogs from revit, they would not recognize old catalogs before this change.
  
FUTURE WORK:
- None

REQUIRED:
- [ ] N/A ~~All changes are up to date in `CHANGELOG.md`.~~

COMMENTS:
- Any other notes.


katehryhorenko (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/765)* status: :shipit: complete! 1 of 1 approvals obtained

PR #774, opened 2022-02-17
Line get overlap extension

mwojcik-sagiton:
BACKGROUND:
- I found two missing methods in Line API:
  - `IsCollinear(Line line)` will check if line is collinear with base line
  - `GetOverlap(Line line)` will return line which is a part of both lines, base and line
  
image

DESCRIPTION:
- `IsCollinear(Line line)` methods checks if `Direction` vectors are parallel
- `GetOverlap(Line line)` method checks if lines are collinear, then it takes all line ends and order them by sum of their coordinates. When 2nd and 3rd vector lies on both lines they are ends of overlapping line. 

TESTING:
- Run `IsCollinear` and `GetOverlap` unit tests 
  
FUTURE WORK:
- There is no future work needed 

REQUIRED:

COMMENTS:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/774)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 - `AdaptiveGrid.AddVertex(Vector3 point, List<Vertex> connections)`
 - `Color.SRGBToLinear(double c)`
 - `Color.LinearToSRGB(double c)`
+- `Line.IsCollinear(Line line)`
+- `Line.GetOverlap(Line line)`
 
 ### Changed
 - Add parameter `removeCutEdges` to `AdaptiveGrid.SubtractBox` that control if cut parts of intersected edges need to be inserted into the graph.
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1,3 +1,5 @@
+using System.Net.Sockets;
+using System.Numerics;
 using Elements.Validators;
 using System;
 using System.Collections.Generic;
@@ -996,7 +998,7 @@ namespace Elements.Geometry
         /// <returns></returns>
         public bool IsCollinear(Line line)
         {
-            return line.Direction().IsParallelTo(Direction());
+            return Vector3.AreCollinear(Start, End, line.Start) && Vector3.AreCollinear(Start, End, line.End);
         }
 
         /// <summary>
@@ -1012,7 +1014,7 @@ namespace Elements.Geometry
             if(!IsCollinear(line))
                 return null;
 
-            //order verticies of lines
+            //order vertices of lines
             var vectors = new List<Vector3>() { Start, End, line.Start, line.End };
             var orderedVectors = vectors.OrderBy(v => v.X + v.Y + v.Z).ToList();
 
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1000,6 +1000,38 @@ namespace Elements.Geometry
         }
 
         /// <summary>
+        /// Get overlapping line of this and other line
+        /// </summary>
+        /// <param name="line">Line to check</param>
+        /// <returns>Returns line or null if lines are not collinear or do not overlap</returns>
+        public Line GetOverlap(Line line)
+        {
+            if(line == null)
+                return null;
+
+            if(!IsCollinear(line))
+                return null;
+
+            //order verticies of lines
+            var vectors = new List<Vector3>() { Start, End, line.Start, line.End };
+            var orderedVectors = vectors.OrderBy(v => v.X + v.Y).ToList();
+
+            //check if 2nd point lies on both lines
+            if (!PointOnLine(orderedVectors[1], Start, End, true) || !PointOnLine(orderedVectors[1], line.Start, line.End, true))
+                return null;
+
+            //check if 3rd point lies on both lines
+            if (!PointOnLine(orderedVectors[2], Start, End, true) || !PointOnLine(orderedVectors[2], line.Start, line.End, true))
+                return null;
+
+            //edge case when lines share only point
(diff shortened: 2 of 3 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/774)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton)

___
*[Elements/src/Geometry/Line.cs, line 999 at r1](https://reviewable.io/reviews/hypar-io/elements/774#-Mw7yzEQ0rTr2ZkSyFE7:-Mw7yzEQ0rTr2ZkSyFE8:b-ggwhyf) ([raw file](https://github.com/hypar-io/elements/blob/62a8d5f2a10688e078fda4e1abfa5bea2b0a361c/Elements/src/Geometry/Line.cs#L999)):*
> ```Smalltalk
>         public bool IsCollinear(Line line)
>         {
>             return line.Direction().IsParallelTo(Direction());
> ```

This isn't quite sufficient to check for collinearity.  I think these two lines would pass this check, but aren't collinear.  We need to also check that they are actually in a line with each other.  mabye this method can help? https://github.com/hypar-io/Elements/blob/5a2c506f2dec9953ca1e2c340e43cd39e5060324/Elements/src/Geometry/Vector3.cs#L818



___
*[Elements/src/Geometry/Line.cs, line 1007 at r1](https://reviewable.io/reviews/hypar-io/elements/774#-Mw7yemm5AxZLu8CQX62:-Mw7yemm5AxZLu8CQX63:b5vehda) ([raw file](https://github.com/hypar-io/elements/blob/62a8d5f2a10688e078fda4e1abfa5bea2b0a361c/Elements/src/Geometry/Line.cs#L1007)):*
> ```Smalltalk
>         /// <param name="line">Line to check</param>
>         /// <returns>Returns line or null if lines are not collinear or do not overlap</returns>
>         public Line GetOverlap(Line line)
> ```

I know I suggested this method name, but could we change to `TryGetOverlap`, return true/false, and the `out Line overlap` the overlapping line?  I think this is more clear than having a null sometimes come out of a method.

mwojcik-sagiton:
[In this commit ](https://github.com/hypar-io/Elements/pull/774/commits/7e27659de2cf798bcba3bd3f85ce7dfb25fd0fcb) I changed collinearity check. I also added adequate unit test of test two parallel lines which are not collinear. 

[In this commit](https://github.com/hypar-io/Elements/pull/774/commits/8f68c970c50067c7b4aa1d7afcd5678db80470e4) I change signature of method for `TryGetOverlap(...)` as you suggested

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/774)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[Elements/src/Geometry/Line.cs, line 999 at r1](https://reviewable.io/reviews/hypar-io/elements/774#-Mw7yzEQ0rTr2ZkSyFE7:-Mwag4vF6MHixTHRLV0J:b7d6vi7) ([raw file](https://github.com/hypar-io/elements/blob/62a8d5f2a10688e078fda4e1abfa5bea2b0a361c/Elements/src/Geometry/Line.cs#L999)):*

This isn't quite sufficient to check for collinearity.  I think these two lines would pass this check, but aren't collinear.  We need to also check that they are actually in a line with each other.  mabye this method can help? https://github.com/hypar-io/Elements/blob/5a2c506f2dec9953ca1e2c340e43cd39e5060324/Elements/src/Geometry/Vector3.cs#L818




</blockquote></details>

done. looks good!
___
*[Elements/src/Geometry/Line.cs, line 1007 at r1](https://reviewable.io/reviews/hypar-io/elements/774#-Mw7yemm5AxZLu8CQX62:-Mwag9kU2f-6jSL8ueIy:b-rnrapz) ([raw file](https://github.com/hypar-io/elements/blob/62a8d5f2a10688e078fda4e1abfa5bea2b0a361c/Elements/src/Geometry/Line.cs#L1007)):*

I know I suggested this method name, but could we change to `TryGetOverlap`, return true/false, and the `out Line overlap` the overlapping line?  I think this is more clear than having a null sometimes come out of a method.
</blockquote></details>

done.

nadiia-volyk (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/774)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @mwojcik-sagiton)

___
*[Elements/src/Geometry/Line.cs, line 1022 at r3](https://reviewable.io/reviews/hypar-io/elements/774#-MwatsYg1B8hMSm01QSS:-MwatsYg1B8hMSm01QST:b-2sq7e8) ([raw file](https://github.com/hypar-io/elements/blob/4d75f647d3e517e69a84fb685f74bcd2e9037c90/Elements/src/Geometry/Line.cs#L1022)):*
> ```Smalltalk
>             //order vertices of lines
>             var vectors = new List<Vector3>() { Start, End, line.Start, line.End };
>             var orderedVectors = vectors.OrderBy(v => v.X + v.Y + v.Z).ToList();
> ```

Ordering points by the sum of coordinates is not reliable. Try this test below. Sum of coordinates for all points  will be 0. Maybe try ordering by dot product of this line direction with each vector?



_Code snippet:_
```Smalltalk
var line1 = new Line(new Vector3(-3, 3), new Vector3(2, -2));
var line2 = new Line(new Vector3(-1, 1), new Vector3(4, -4));
Assert.True(line1.TryGetOverlap(line2, out Line overlapResult1));
Assert.True(line2.TryGetOverlap(line1, out Line overlapResult2));
Assert.True(overlapResult1.IsAlmostEqualTo(overlapResult2, false));
```

PR #777, opened 2022-03-04
Fix polygon creation from vertices with sequential duplicates

nadiia-volyk:
BACKGROUND:
- It's not possible to create polygon from list of vertices where first and last points are equal. In `Polygon` constructor this case is solved, but in base `Polyline` constructor it's not - duplicated last point is not removed and then overridden method `Edges` returns zero length edge and exception is thrown "A segment of the polyline has zero length".

DESCRIPTION:
- Extract vertices validation to virtual method, so the right validation logic will be used for derived classes

TESTING:
- Add `PolygonTests.ConstructWithSequentialDuplicates` test.
  
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/777)

### DIFF
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -31,26 +31,26 @@ namespace Elements.Geometry
         [Newtonsoft.Json.JsonConstructor]
         public Polygon(IList<Vector3> @vertices) : base(vertices)
         {
-            if (!Validator.DisableValidationOnConstruction)
+            _plane = Plane();
+        }
+
+        protected override void ValidateVertices()
+        {
+            if (!Vertices.AreCoplanar())
             {
-                if (!vertices.AreCoplanar())
-                {
-                    throw new ArgumentException("The polygon could not be created. The provided vertices are not coplanar.");
-                }
-
-                this.Vertices = Vector3.RemoveSequentialDuplicates(this.Vertices, true);
-                DeleteVerticesForOverlappingEdges(this.Vertices);
-                if (this.Vertices.Count < 3)
-                {
-                    throw new ArgumentException("The polygon could not be created. At least 3 vertices are required.");
-                }
-
-                CheckSegmentLengthAndThrow(Edges());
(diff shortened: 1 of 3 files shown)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/777)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nadiia-volyk)

___
*[Elements/test/PolygonTests.cs, line 2003 at r1](https://reviewable.io/reviews/hypar-io/elements/777#-MxLRbhq5OCSQjgkUcqW:-MxLRbhr3SavrjsEvsoz:bmw5v7x) ([raw file](https://github.com/hypar-io/elements/blob/2903aa8c0af373c9a9800829803775edd6f5a296/Elements/test/PolygonTests.cs#L2003)):*
> ```Smalltalk
>                 Vector3.Origin,
>             });
>             Assert.True(true);
> ```

Is this just intended to test that the constructor didn't throw an exception? I think this line can be omitted — the test will fail with an exception.

nadiia-volyk (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/777)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/test/PolygonTests.cs, line 2003 at r1](https://reviewable.io/reviews/hypar-io/elements/777#-MxLRbhq5OCSQjgkUcqW:-MxZ7Ls47xqtB42qe90p:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/2903aa8c0af373c9a9800829803775edd6f5a296/Elements/test/PolygonTests.cs#L2003)):*

Is this just intended to test that the constructor didn't throw an exception? I think this line can be omitted — the test will fail with an exception. 
</blockquote></details>

Done.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/777)* status: :shipit: complete! 1 of 1 approvals obtained

PR #778, opened 2022-03-09
Mesh sphere

ikeough:
BACKGROUND:
- Having some mesh primitives like spheres and cubes is useful for various tasks.

DESCRIPTION:
- This PR adds the `Mesh.Sphere` method to create a mesh sphere.
- This PR fixes the `ToPolyline` method, which was creating polylines that were one segment short of the end of the curve being converted. It then also fixes all tests where `ToPolyline` is excercised, and adds tests to ensure that polylines created in this way align with the start and end points of curves.

REQUIRED:

image

Use the `divisions` parameter to change the precision.
Example 10
image

Example 50
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/778)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.0.0
 
 ### Added
+- `MeshPrimitives.Sphere(double radius, int divisions)`
 
 ### Changed
 - Remove ``removeCutEdges` from `AdaptiveGrid.SubtractBox` and always remove cut parts of intersected edges.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 1.0.0
 
 ### Added
-- `MeshPrimitives.Sphere(double radius, int divisions)`
+- `Mesh.Sphere(double radius, int divisions)`
 
 ### Changed
 - Remove ``removeCutEdges` from `AdaptiveGrid.SubtractBox` and always remove cut parts of intersected edges.
--- a/Elements/src/Geometry/Bezier.cs
+++ b/Elements/src/Geometry/Bezier.cs
@@ -91,12 +91,11 @@ namespace Elements.Geometry
         /// </summary>
         public override double Length()
         {
-            var div = 1.0 / _samples;
             Vector3 last = new Vector3();
             double length = 0.0;
-            for (var t = 0.0; t <= 1.0; t += div)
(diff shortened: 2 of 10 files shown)


andrewheumann (approved):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/778)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Geometry/Meshes.cs, line 14 at r1](https://reviewable.io/reviews/hypar-io/elements/778#-MxngAPb8XkACP-ChhZR:-MxngAPb8XkACP-ChhZS:b-obufbs) ([raw file](https://github.com/hypar-io/elements/blob/4b2dd88852c570b6604485e627aba8779a78d7b6/Elements/src/Geometry/Meshes.cs#L14)):*
> ```Smalltalk
>         /// <param name="divisions">The number of tessellations of the sphere.</param>
>         /// <returns>A mesh.</returns>
>         public static Mesh Sphere(double radius, int divisions = 10)
> ```

Should we protect against an invalid "divisions" value? it seems we shouldn't go lower than 2 (could also produce a div by 0 error)

andrewheumann (commented):
one small note, otherwise LGTM

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/778#-:-MxnoNUP0AXEx-nGE8DW:bd6saim)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/778)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Meshes.cs, line 14 at r1](https://reviewable.io/reviews/hypar-io/elements/778#-MxngAPb8XkACP-ChhZR:-MxpvF98BJtoV-fhKOKf:bb6un1j) ([raw file](https://github.com/hypar-io/elements/blob/4b2dd88852c570b6604485e627aba8779a78d7b6/Elements/src/Geometry/Meshes.cs#L14)):*

Should we protect against an invalid "divisions" value? it seems we shouldn't go lower than 2 (could also produce a div by 0 error)
</blockquote></details>

Done

PR #779, opened 2022-03-11
Add emissive texture to material.

ikeough:
BACKGROUND:
The glTF specification allows for emissive textures. The components of an emissive texture's color are multiplied after the affect of lighting and shadows simulating an object which emits light.

DESCRIPTION:
This PR adds the `Material.EmissiveTexture` and `Material.EmissiveFactor` properties and updates our glTF writing to add these values to exported glTFs.

REQUIRED:

This image shows a solid orange sphere with an emissive checkerboard pattern with emissive values of 0.5

image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/779)


andrewheumann:
Is it possible to do this without changing the constructor, so we don't break existing calling code?

andrewheumann:
Or at least add overloads for old constructors?

ikeough:
The new parameters are all default parameters, except for the JSON constructor, in which case the values will be null or 0 after construction without values from the JSON. What am I missing?

andrewheumann:
A call like this, which calls the positional arguments, will break:
`var mat = new Material(Colors.Orange, 0.4, 0.4, false, null, true, false, null, false, Guid.NewGuid());`

However, I looked in building blocks to see how many cases this would break, and it turns out all nearly every example I could find was _already_ broken, since the addition of the  `repeatTexture, normalTexture, interpolateTexture` parameters: 
```
var liftMatl = new Material(new Color(1.0f, 0.9f, 0.4f, 0.8f), 0.0f, 0.0f, false, null, false, Guid.NewGuid(), "lift");
var mat3 = new Material(new Color(1.0f, 0.0f, 0.0f, 0.8f), 0.0f, 0.0f, false, null, false, Guid.NewGuid(), "stair");
var mechMatl = new Material(new Color(0.2f, 0.2f, 0.2f, 0.8f), 0.0f, 0.0f, false, null, false, Guid.NewGuid(), "mech");
var rhMat = new Material(new Color(0.5, 0.5, 0.5, 1), 0.1, 0.1, false, null, true, Guid.NewGuid(), "Default");
```
I know from experience I've frequently had to change calling code with updates like this to the Material constructor. 

In the context of other, code-generated elements, we've added an empty constructor by default, so that it's possible to write code that can't be broken by the addition of new parameters:
`new Foo() { arg1= value, arg2= value }`.

These are the options I see as reasonable:
1. Add a few "backwards compatibility" constructors with older configurations of positional arguments.
2. Assume the backwards compatibility ship has sailed, add an empty constructor, and encourage users to use this, like we already do for other code-genned elements. We should probably also comb through all of BB to make this change, and update reference code snippets in the docs. 
3. Assume that the current configuration of positional arguments is the "frozen" one, and don't add these new parameters to the JSONConstructor. Serialization will still work just fine.

ikeough:
I'd add an empty constructor because those code examples are full of default values, and it will shorten the code.

I'm not entirely understanding the backwards compatibility argument though. As long as we don't break deserialization we're fine to add as many arguments as we like. It's a versioned API after all. BB doesn't have to update.

We don't code-gen Material right?

ikeough:
It seems our definition of "break" is different. Versions of an API are allowed to change constructors. That's what versions are for (especially this one which is moving to 1.0).

The addition of the parameterless constructor will minimize future code changes required to calling code, but it wouldn't protect us from actual "breaking" changes like parameter renaming or removal.

ikeough:
We might also be able to make the full JSON constructor internal, to prevent its use and prevent future user code breakage, and do a one time conversion of all our code to using the parameterless constructor.

andrewheumann (commented):
parameterless constructor works for me :heavy_check_mark: 

> Reviewed 3 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/779#-:-My9-Fld7hXX1dbtBBPT:b69f1zm)* status: 0 of 1 approvals obtained (waiting on @katehryhorenko)

katehryhorenko (approved):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/779)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 311 at r2](https://reviewable.io/reviews/hypar-io/elements/779#-MyKF6qOBGBA-dB_aFzg:-MyKF6qOBGBA-dB_aFzh:b-bupk02) ([raw file](https://github.com/hypar-io/elements/blob/e50075191ed034f1f20301c3e92478bc86817484/Elements/src/Serialization/glTF/GltfExtensions.cs#L311)):*
> ```Smalltalk
>                 if (material.EmissiveTexture != null && File.Exists(material.EmissiveTexture))
>                 {
>                     var ti = new TextureInfo();
> ```

Small comment here. I prefer more understandable names, like textureInfo or material instead of ti and m. It makes code easier to read

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/779)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[Elements/src/Serialization/glTF/GltfExtensions.cs, line 311 at r2](https://reviewable.io/reviews/hypar-io/elements/779#-MyKF6qOBGBA-dB_aFzg:-Myp-WOk61SX9RFoK-WA:b-yrdhxy) ([raw file](https://github.com/hypar-io/elements/blob/e50075191ed034f1f20301c3e92478bc86817484/Elements/src/Serialization/glTF/GltfExtensions.cs#L311)):*

Small comment here. I prefer more understandable names, like textureInfo or material instead of ti and m. It makes code easier to read
</blockquote></details>

Agreed. Fixed throughout.

PR #781, opened 2022-03-17
Added ApproximatelyEquals call as first check to avoid error with acc…

katehryhorenko:
…uracy;

BACKGROUND:
A new node that is parallel to a parent and has almost the same Y coordinates was added to the tree as left node but during traverse, it was looking as right node 

DESCRIPTION:
The ApproximatelyEquals call is added as first check to avoid error with accuracy

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/781)

### DIFF
--- a/Elements/src/Search/LeftMostPointComparer.cs
+++ b/Elements/src/Search/LeftMostPointComparer.cs
@@ -58,15 +58,15 @@ namespace Elements.Search
             }
             else
             {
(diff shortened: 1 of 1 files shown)


ikeough (approved):

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/781)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #739, opened 2021-12-22
Network creation from line segments does not split edges at intersections.


ikeough:
**Describe the bug**
When a line intersects another line at the first line's end, the resulting network should have a vertex at this split location with appropriate split edges. Currently, the second edge is not split.

**Screenshots**
![image](https://user-images.githubusercontent.com/1139788/147151530-a9ef2ce6-cd25-4bfe-b4f7-070ca3330eed.png)

ikeough:
This is super annoying. What's happening is that `LeftMostComparer` is returning 0 for the horizontal line and the middle vertical line because their Y components are the same. This keeps the middle line from ever being added to the `BinarTree` used to find adjacent lines for intersection during the sweep. Therefore, the middle line only gets vertices at its ends, and the horizontal line is never split. This can be fixed by adjusting the left most comparer to use a "fake" point for lines which have the same Y location for their end points, using the middle point. If you do this though, you get an infinite loop in the `FindPredecessorsAndSuccessors` method because it's possible to build cyclic graphs.

ikeough:
This works when you rotate all the lines by a small amount as shown here. This suggests that vertical and horizontal lines are not properly handled by the line sweep algorithm.
![image](https://user-images.githubusercontent.com/1139788/158914307-b0699ff8-2dcd-48dd-ac2c-46b8cc8c1582.png)

ikeough:
This is resolved with #782.

PR #782, opened 2022-03-18
Axis aligned intersections in Networks

ikeough:
BACKGROUND:
See #739 

DESCRIPTION:
- This PR adjusts the line sweep algorithm to fix perpendicular intersections by sorting line sweep events first in Y then in X. This has the effect of tilting the sweep line, allowing line segments to stay in the binary tree for the correct duration to do intersections.
- The linesweep algorithm is also adjusted to handle only predecessor or successor intersections with the active segment in the segment's end event. Previously we only handled the case of the predecessor and the successor intersection each other.
- Several locations where we use `ToList()` were removed in the interest of using the `IEnumerable` provided by linq.

TESTING:
- The `NetworkTests.FigureEight()` test has been added to ensure that all perpendicular intersections are handled.
  
REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/782)


katehryhorenko (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/782)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Search/Network.cs, line 225 at r1](https://reviewable.io/reviews/hypar-io/elements/782#-MzLVl9P0xmw4Oudto2I:-MzLVl9P0xmw4Oudto2J:b-7e2xjl) ([raw file](https://github.com/hypar-io/elements/blob/dcf01daa715b8df9db12b6c156ef013e28d75b8f/Elements/src/Search/Network.cs#L225)):*
> ```Smalltalk
>                         }
> 
>                         if (pre != null)
> ```

                        if (pre != null && s.Intersects(getSegment(pre.Data), out Vector3 result, includeEnds: true))

PR #784, opened 2022-03-23
Add AllElementsDerivedFromType<T>().

ikeough:
BACKGROUND:
It is often the case that we want to get all elements in a model which share a base class. 

DESCRIPTION:
- This PR adds the `Model.AllElementsDerivedFromType<T>()` method.

TESTING:
- See `ModelTests.AllElementsDerivedFromType()`

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/784)


wynged (approved):
One note/thought.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/784#-:-Myqle9eC6nPamhjWvMT:b-xvrbwt)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[Elements/src/Model.cs, line 200 at r1](https://reviewable.io/reviews/hypar-io/elements/784#-Myqksng0I4EVrEBa6KS:-Myqksng0I4EVrEBa6KT:b-r6qayz) ([raw file](https://github.com/hypar-io/elements/blob/69a7d4728bf8bb00c553389e2afddf199f9c6776/Elements/src/Model.cs#L200)):*
> ```Smalltalk
>         public IEnumerable<T> AllElementsDerivedFromType<T>()
>         {
>             return this.Elements.Values.Where(e => typeof(T).IsAssignableFrom(e.GetType())).Cast<T>();
> ```

non-blocking, but I'm curious.

I think this would actually work for interfaces as well?  That makes the name of this method a bit limiting/misleading, but I can't think of a better on that wouldn't be more confusing.  maybe explicitly `AllElementsAssignableFromType` would be understandable and more inclusive?

PR #787, opened 2022-03-23
Fixes for Vector3.AreCollinear and Line.TryGetOverlap

DmytroMuravskyi:
BACKGROUND:
When working with walls that belong to several rooms I realized that some wall that were shared between two rooms were not correctly identified. 
I have found two issues in Elements:
- `Vector3.AreCollinear(Vector3 a, Vector3 b, Vector3 c)` was returning `false` if two points are coincident but not exactly.
Checking IsZero after normalizing is not efficient when two points are almost coincident. In this case even slightest variation between two points start to dominate.
- `Line.TryGetOverlap(Line line, out Line overlap)` would return incorrect results due to wrong internal sorting.

DESCRIPTION:
- In `Vector3.AreCollinear` do IsZero check before normalization.
- In `Line.TryGetOverlap`, since it works on collinear lines, use dot product sorting.

TESTING:
- Added extra test cases to `IsCollinear` and `TryGetOverlap` tests.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/787)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Remove ``removeCutEdges` from `AdaptiveGrid.SubtractBox` and always remove cut parts of intersected edges.
 
 ### Fixed
+- `Vector3.AreCollinear(Vector3 a, Vector3 b, Vector3 c)` would return `false` if two points are coincident but not exactly.
+- `Line.TryGetOverlap(Line line, out Line overlap)` would return incorrect results due to wrong internal sorting.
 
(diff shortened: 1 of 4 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/787)* status: :shipit: complete! 1 of 1 approvals obtained

PR #791, opened 2022-03-31
Polygon: add internal SplitInternalLoops. Use it in ToProfile used by Clipper lib.

DmytroMuravskyi:
BACKGROUND:
While working with HyparSpace workflow a have found that corridors floor has strange profile: while 6 "holes" are visible Profile object has only 3. It appeared that 1 void was embedded into perimeter and other void was, in fact, 3 voids together.
image

DESCRIPTION:
Polygon can't have self intersections but points still can lay on other lines. 
Profile.UnionAll produced such complex loops.
This lead to hidden void, embedded into them and complicates processing the profile.

TESTING:
Added ProfileComplexUnionAll test that has exact shapes of corridors from HyparSpace.

FUTURE WORK:
This changes provide cleanup for Profiles produced by boundary operations but since it affect all boolean operations that use ToProfile, it can have unwanted side effects.

REQUIRED:

COMMENTS:
- Any other notes.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/791)


andrewheumann (changes requested):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[Elements/src/Geometry/Profile.cs, line 648 at r1](https://reviewable.io/reviews/hypar-io/elements/791#-MzZePEW8rDg08k-zZnX:-MzZePEW8rDg08k-zZnY:bhgc95q) ([raw file](https://github.com/hypar-io/elements/blob/b1299f1dac7207f7bc12a1169d7308404c6bcb24/Elements/src/Geometry/Profile.cs#L648)):*
> ```Smalltalk
>                     }
>                 }
>                 perimeter = largest;
> ```

Is there any scenario where the loops produced by `SplitInternalLoops` are not all within the largest loop? Is a situation like this possible, where we'd want to actually route this to produce a second profile (or maybe intentionally discard the loop that's external)?


We don't necessarily have to address this situation in this PR — especially since I've never seen this specific scenario emerge in the wild — but it would be good to capture any assumptions here in a comment, and at least make sure if such a thing were to occur that we wouldn't throw an exception.

DmytroMuravskyi (commented):
Yes, I happened because Polygon constructor itself does cleanup and can throw if there are less than 3 points after that. I reworked how I create Polygons on the function so all exceptions are catched.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791#-:-MzZoRntCuXOgnAaibTZ:bl4uk2)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements/src/Geometry/Polygon.cs, line 2232 at r1](https://reviewable.io/reviews/hypar-io/elements/791#-MzVFsVc79GVdLcU9Lio:-MzZoQ4k6Y3K1-xcl2l6:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/b1299f1dac7207f7bc12a1169d7308404c6bcb24/Elements/src/Geometry/Polygon.cs#L2232)):*

=> 

```
/// A Polygon can't have self intersections, but points can still lay on other lines.
/// This leads to hidden voids embedded in the perimeter.
/// This function checks if any points are on another line of the loop and splits into distinct loops if found.
```
</blockquote></details>

Done.
___
*[Elements/src/Geometry/Profile.cs, line 648 at r1](https://reviewable.io/reviews/hypar-io/elements/791#-MzZePEW8rDg08k-zZnX:-MzZwPQW5KGR4pkNHvMl:b-eye4wj) ([raw file](https://github.com/hypar-io/elements/blob/b1299f1dac7207f7bc12a1169d7308404c6bcb24/Elements/src/Geometry/Profile.cs#L648)):*

Is there any scenario where the loops produced by `SplitInternalLoops` are not all within the largest loop? Is a situation like this possible, where we'd want to actually route this to produce a second profile (or maybe intentionally discard the loop that's external)?


We don't necessarily have to address this situation in this PR — especially since I've never seen this specific scenario emerge in the wild — but it would be good to capture any assumptions here in a comment, and at least make sure if such a thing were to occur that we wouldn't throw an exception. 
 
</blockquote></details>

I think that this should not happen within one Profile produced by Clipper. Added a comment and a test case showing current behavior.

andrewheumann (commented):
Using this version of Elements on existing customer HyparSpace workflows is causing various boolean results to disappear completely. Perhaps this should be handled at the level of the function, but I'd like to understand what's causing it.




I'll try to isolate specific calls within elements that are producing differing results. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791#-:-Mzplyr4ETGLXktoOBoG:b-yfk3f)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (commented):
OK, so have done a little further investigation. This was the condition whose results have changed: (Red minus blue) in `Profile.Difference`:


It all centers around an input condition like this:


`Profile.Difference` used to yield profiles that looked like this (colored by index):


This is clearly not a good solution. That blue profile is terrible. 

With your change, we get this instead (blue is Perimeter, yellow is void): 


This is obviously better, but some shapes get misclassified as voids. This is precisely the case your comment highlights. I think it would be worth addressing this now, to avoid shapes suddenly "disappearing" — can we add a containment check in `ToProfile`, or perhaps in `ToProfiles`, to make sure any voids that lay outside of the perimeter get turned into new profiles? If you would prefer to leave this as a `TODO` and address in a separate, follow-up PR, that would be OK with me too. 

I've created a new branch `misclassified-curves-in-profile` that contains example geometry and a test which passes on `master` and fails on your branch to illustrate, if you'd like to test the specific case I found. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791#-:-Mzq5Ri28SLpA7HSeM89:bovuf5p)* status: 1 change requests, 0 of 1 approvals obtained

DmytroMuravskyi (commented):
Thanks for testing this. Since this scenario does happen, I've changed ToPolygon to support multiply perimeters. I've also included the test you provided as it seems very useful for future regression testing. Can you verify if my changes affected other case you've mentioned?

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791#-:-Mztl1GO7RP-Srv4t755:b8m09zp)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann (approved):
Tested against a wide range of customer cases. Looks great! Thanks for all your work on this — it's a big improvement.

> Reviewed 1 of 3 files at r2, 2 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/791#-:-N--Pv7e3-nRBEBc2Tpl:bclr1qo)* status: :shipit: complete! 1 of 1 approvals obtained

PR #792, opened 2022-04-08
better support for excluding types from codegen

andrewheumann:
BACKGROUND:
- This PR works in tandem with https://github.com/hypar-io/Hypar/pull/465.
- In complex scenarios, code generation would not exclude all types it was expected to. For example:
Type `CirculationSegment`'s schema has it extend `Floor`. We were only checking top-level types from `element_types` against known types, so a `Floor.g.cs` still got created, breaking compilation.
- A function containing within its `element_types` a schema (like `CirculationSegment`) which refers to a type like `ThickenedPolyline` which can ALSO be a input type would result in dual generation of that type: `Elements.ThickenedPolyline` and `Circulation.ThickenedPolyline` in the `Inputs` class, and one could not be converted to the other. 

DESCRIPTION:
- Adds more opportunities to inject a `HashSet` of excluded types, and modifies it on the fly as types are generated. 
- I also added an empty constructor for `Floor`, so that a generated type's empty constructor can call the built-in type's empty constructor. 

TESTING:
- I tested in a modified version of the Circulation function, which has a `ThickenedPolyline` override, as well as a reference to the `CirculationSegment` schema (`https://prod-api.hypar.io/schemas/CirculationSegment`). I verified that it code-genned and built (with the parallel `hypar` CLI PR) correctly.
- I also tested in other function with a complex type scenario: `SpacePlanningZones` from Hypar Space has a mix of extended partial classes, types referenced from external assemblies, types in its `input_schema,` and so on — I verified that it still code-genned correctly and built. 
  
FUTURE WORK:
- We should add empty constructors to all element types that we might inherit from (or be more explicit about prohibiting inheritance in elements)

REQUIRED:

COMMENTS:
- It would be great to test this (and the accompanying CLI PR) against some other complex dependency/codegen scenarios, such as in aliaxis

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/792)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,21 +3,27 @@
 ## 1.0.0
 
 ### Added
+
 - `Mesh.Sphere(double radius, int divisions)`
 - `Model.AllElementsAssignableFromType<T>()`
 - `Material.EmissiveTexture`
 - `Material.EmissiveFactor`
 
 ### Changed
+
 - Remove ``removeCutEdges` from `AdaptiveGrid.SubtractBox` and always remove cut parts of intersected edges.
+- `GenerateUserElementTypeFromUriAsync` now takes an optional `excludedTypes` argument.
 
 ### Fixed
+
 - `Vector3.AreCollinear(Vector3 a, Vector3 b, Vector3 c)` would return `false` if two points are coincident but not exactly.
 - `Line.TryGetOverlap(Line line, out Line overlap)` would return incorrect results due to wrong internal sorting.
 - `Profile.UnionAll, Difference, Intersection, Offset` no longer produce internal loops in `Perimeter` or `Voids`.
+
 ## 0.9.9
 
 ### Added
+
 - `Solid.Union(Solid a, Transform aTransform, Solid b, Transform bTransform)`
 - `Solid.Union(SolidOperation a, SolidOperation b)`
 - `Solid.Difference(Solid a, Transform aTransform, Solid b, Transform bTransform)`
@@ -35,10 +41,12 @@
 - `Line.GetOverlap(Line line)`
 
 ### Changed
+
 - Add parameter `removeCutEdges` to `AdaptiveGrid.SubtractBox` that control if cut parts of intersected edges need to be inserted into the graph.
 - Material colors are now exported to glTF using linear color space. Conversion from sRGB to linear color space happens during glTF export.
 
 ### Fixed
+
 - Under some circumstances `Bezier.Length()` would return incorrect results
 
 ## 0.9.8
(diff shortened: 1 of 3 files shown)


wynged (approved):
have tested against Aliaxis, seems to behave well.
couple of nits but approved.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/792#-:-N-Tv9Ux0rI15f0J_UWx:bphg190)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 170 at r1](https://reviewable.io/reviews/hypar-io/elements/792#-N-PQ18f-kXUNrNBkiAp:-N-PQ18f-kXUNrNBkiAq:b-ezthic) ([raw file](https://github.com/hypar-io/elements/blob/91db5bb84a6f3c3519ec96ef067b16de6342c22c/Elements.CodeGeneration/src/TypeGenerator.cs#L170)):*
> ```Smalltalk
> 
>             var typeName = schema.Title;
>             var excludedTypes = excludedTypeNames ?? new HashSet<string>(_coreTypeNames);
> ```

I think I prefer the var name before, that included `Names`
___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 410 at r1](https://reviewable.io/reviews/hypar-io/elements/792#-N-PQboT4Sh-S-1hMXsI:-N-PQboT4Sh-S-1hMXsJ:b-w8supv) ([raw file](https://github.com/hypar-io/elements/blob/91db5bb84a6f3c3519ec96ef067b16de6342c22c/Elements.CodeGeneration/src/TypeGenerator.cs#L410)):*
> ```Smalltalk
>                 ArrayType = "IList",
>                 ArrayInstanceType = "List",
>                 ExcludedTypeNames = excludedTypes ?? (new string[] { }),
> ```

I don't think these parenthesis are necessary?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/792)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @wynged)

___
*[Elements.CodeGeneration/src/TypeGenerator.cs, line 410 at r1](https://reviewable.io/reviews/hypar-io/elements/792#-N-PQboT4Sh-S-1hMXsI:-N-U0QpG3DEa7XQXLN_7:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/91db5bb84a6f3c3519ec96ef067b16de6342c22c/Elements.CodeGeneration/src/TypeGenerator.cs#L410)):*

I don't think these parenthesis are necessary?
</blockquote></details>

Done.

PR #801, opened 2022-05-11
Make TryToGraphicsBuffers public

andrewheumann:
BACKGROUND:
- Users may want to create custom types that specify their own buffer logic. Previously this was only possible for types within Elements.
 
DESCRIPTION:
- Exposes `TryToGraphicsBuffers` and `ToGraphicsBuffers` APIs as public
- simplifies `ToGraphicsBuffers`. It appears it was doing some unnecessary index manipulation to mimic the behavior of  `GL_LINE_LOOP` — but since we exposed setting the Mode (in #704, I think?), this is no longer needed.

TESTING:
- I checked the GLB output of 
  - `GridLine`
  - `ToGLTF()` with `drawEdges`=`true`
  - other `ModelCurve` tests producing both open and closed curves
  - `ModelLines`
 and verified it was visually unchanged
  
FUTURE WORK:
  - `ToGLTF()` with `drawEdges`=`true` appears somewhat broken. It only works right now for element instances. 

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/801)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Added
 
+### Changed
+
+- Simplified `IEnumerable<Vector3>.ToGraphicsBuffers()`
+- `TryToGraphicsBuffers` is now public
+
 ## 1.0.0
 
 ### Added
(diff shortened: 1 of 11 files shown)


serenayl (approved):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/801)* status: :shipit: complete! 1 of 1 approvals obtained

PR #804, opened 2022-05-11
add support for LineDisplaySettings

andrewheumann:
BACKGROUND:
- Sometimes, it is useful to render diagrammatic elements in a model with display line-weights. Some elements are strictly conceptual, like structural wireframes, and they may not want to be shown as infinitely-thin curves.

DESCRIPTION:
- Adds support for `LineDisplaySettings` as a property on `Material`, for configuring how curve geometry should be rendered in a supporting 3D viewer. 
```csharp
new ModelCurve(pline, new Material("Purple", Colors.Purple) { 
     LineDisplaySettings = new LineDisplaySettings { 
          Width = 10, 
          Mode = LineDisplayWidthMode.ScreenUnits 
     } 
}, new Transform(10, 0, 0));
```
- Utilizes the pattern of a glTF extension, which we're calling `HYPAR_materials_edge_settings`. (Open to suggestions for a better name! once it's in there, it's in there!)
an example from the generated GLTF:
```json
{
      "name": "f6cc304f-5078-42e9-b6c8-07d271d04f82",
      "extensions": {
        "KHR_materials_pbrSpecularGlossiness": {...},
        "HYPAR_materials_edge_settings": {
          "lineWidth": 5,
          "sizeMode": 0
        }
      }
}
```

`lineWidth` is a number determining the width of the line. `sizeMode` can be either 0 or 1 — if 0, the `lineWidth` is interpreted as a pixel width and rounded to the nearest integer. If 1, the `lineWidth` is interpreted as being in model units (meters), and should scale as the user zooms in and out. 

TESTING:
- This will have no effect anywhere except a viewer that supports this draft extension. A forthcoming hypar PR will demonstrate a first front-end implementation consuming this information to render lines as meshes. 
  
FUTURE WORK:
- We should consider registering `HYPAR` as a glTF vendor prefix. 
- We should consider developing a [formal spec](https://github.com/KhronosGroup/glTF/blob/main/extensions/Template.md) for the extension 
- We could add dash settings to `LineDisplaySettings` (and the extension) as well. 

REQUIRED:

COMMENTS:
- Any other notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+`LineDisplaySettings` for materials to control the display of lines in supported viewers (like Hypar.io).
+
 ## 1.0.0
 
 ### Added
--- a/Elements/src/LineDisplaySettings.cs
+++ b/Elements/src/LineDisplaySettings.cs
@@ -1,0 +1,33 @@
+namespace Elements
+{
+    /// <summary>
+    /// Settings for how a curve or line should be displayed. 
+    /// </summary>
+    public class LineDisplaySettings
+    {
+        /// <summary>
+        /// The width of the line. If Mode is set to Screen Units, this will be in pixels (and rounded to the nearest integer). 
+        /// If set to World Units, this will be in meters. 
+        /// </summary>
+        public double Width { get; set; } = 1;
+        /// <summary>
+        /// How the Width should be interpreted. If set to Screen Units, Width is interpreted as a constant pixel width (and rounded to the nearest integer). If set to World Units, Width is interpreted as a constant meter width.
+        /// </summary>
+        public LineDisplayWidthMode Mode { get; set; } = LineDisplayWidthMode.ScreenUnits;
+    }
+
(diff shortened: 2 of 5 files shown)


serenayl (approved):
> materials_edge_settings
We already have a divergence between `LineDisplaySettings` vs `edge_settings`. Should it be something like `HYPAR_materials_line_settings` vs `edge`?

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/804#-:-N2P-SDF4bJ4TG7de5r8:b-t9qszu)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (commented):
Renamed properties for consistency.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/804#-:-N2SH4wZ7YBob2KjMjvj:b-t4ch1)* status: :shipit: complete! 1 of 1 approvals obtained

PR #808, opened 2022-05-20
False collinearity of close lines 

mwojcik-sagiton:
BACKGROUND:
I noticed that `Line.IsCollinear()` method returns true for lines which are not collinear and are close to each other. 

Lines which break:
- Start = x: 458.8830, y:-118.7170, z: 13.8152; End = x:458.8830, y:-80.4465, z:13.8152
- Start = x:458.9005, y:29.6573, z:13.7977; End = x:458.9005, y:33.5632, z:13.7977;

DESCRIPTION:
I changed method of checking collinearity inside `Line` class to that which accepts list of `Vectors`. Hopefully it does not break when points are close to each other.
I also fixed `Vector.AreCollinear()` method to return true collinearity checks for Vectors which are not almost the same. 

TESTING:
I added new unit tests with edge cases I noticed. 
  
FUTURE WORK:

REQUIRED:

COMMENTS:
Currently there are two methods of testing collinearity of `Vectors`. One is in in `Vector3` class and takes 3 Vectors as an argument, second is extension method of list of Vectors. I would suggest to use only one, preferably that one on extension method because it is versatile.

### DIFF
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -999,7 +999,8 @@ namespace Elements.Geometry
         /// <returns></returns>
         public bool IsCollinear(Line line)
         {
-            return Vector3.AreCollinear(Start, End, line.Start) && Vector3.AreCollinear(Start, End, line.End);
+            var vectors = new Vector3[] { Start, End, line.Start, line.End };
+            return vectors.AreCollinear();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/808)* status: :shipit: complete! 1 of 1 approvals obtained

PR #809, opened 2022-05-20
Line GetUValue for point on line

mwojcik-sagiton:
BACKGROUND:
I found `GetUParameter()` method useful and missing in `Line` class. 

DESCRIPTION:
`GetuUParameter` checks if point is on line. If not it returns -1, otherwise it returns U parameter.

TESTING:
Run unit tests 
  
FUTURE WORK:
None

REQUIRED:

COMMENTS:
None

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/809)

### DIFF
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1050,7 +1050,7 @@ namespace Elements.Geometry
         /// </summary>
         /// <param name="point"></param>
         /// <returns>Returns U parameter for point on line</returns>
-        public double GetUValue(Vector3 point)
+        public double GetUParameter(Vector3 point)
         { 
             if (!PointOnLine(point, true))
                 return -1;
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1075,17 +1075,18 @@ namespace Elements.Geometry
                 return -1;
             }
 
-            if(point.IsAlmostEqualTo(start))
+            if (point.IsAlmostEqualTo(start))
             {
                 return 0;
             }
 
-            if(point.IsAlmostEqualTo(end))
+            if (point.IsAlmostEqualTo(end))
             {
                 return 1;
             }
 
             return (point - start).Length() / (end - start).Length();
+        }
 
         /// Projects current line onto a plane
         /// </summary>
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1050,7 +1050,7 @@ namespace Elements.Geometry
         /// </summary>
         /// <param name="point"></param>
         /// <returns>Returns U parameter for point on line</returns>
-        public double GetUParameter(Vector3 point)
+        public double GetParameterAt(Vector3 point)
         { 
             if (!PointOnLine(point, true))
                 return -1;
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1048,16 +1048,43 @@ namespace Elements.Geometry
         /// <summary>
         /// Calculate U parameter for point on line
         /// </summary>
-        /// <param name="point"></param>
+        /// <param name="point">Point on line</param>
         /// <returns>Returns U parameter for point on line</returns>
         public double GetParameterAt(Vector3 point)
-        { 
+        {
             if (!PointOnLine(point, true))
+            {
                 return -1;
+            }
 
-            var distance = Start.DistanceTo(point);
+            return GetParameterAt(point, Start, End);
+        }
 
-            return distance / Length();
+        /// <summary>
+        /// Calculate U parameter for point between two other points
+        /// </summary>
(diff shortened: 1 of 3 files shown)


Elements/src/Geometry/Line.cs
```diff
@@ -1045,6 +1045,21 @@ public bool TryGetOverlap(Line line, out Line overlap)
             return true;
         }
 
+        /// <summary>
+        /// Calculate U parameter for point on line
+        /// </summary>
+        /// <param name="point"></param>
+        /// <returns>Returns U parameter for point on line</returns>
+        public double GetUParameter(Vector3 point)
```
ikeough, line 1053: Could we call this `GetParameterAt`? We have other methods for getting points on lines that have names like `GetPointAt(...)`.

mwojcik-sagiton: done

wynged (changes requested):
Could you look at the pattern used for the PointOnLine method and implement a similar pattern?  There is a static method that will accept a point, and then `start` and `end` Vector3, so we have both a method on a line, but also one that doesn't require instantiating a line if we want better performance.

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809#-:-N2nlzsq-noq2rbjL2FC:b-hppb5g)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mwojcik-sagiton)

___
*[`Elements/src/Geometry/Line.cs` line 1055 at r2](https://reviewable.io/reviews/hypar-io/elements/809#-N2mh4za3-4Mq2VuEXaU:-N2mh4za3-4Mq2VuEXaV:b-3jncpb) ([raw file](https://github.com/hypar-io/elements/blob/966498b25360243c169bc528bb48bed32fcc9c97/Elements/src/Geometry/Line.cs#L1055)):*
> ```Smalltalk
>         public double GetParameterAt(Vector3 point)
>         { 
>             if (!PointOnLine(point, true))
> ```

add curley braces around the if expression.

mwojcik-sagiton (commented):
Thank you for suggestion. I added new static method `GetParameterAt(Vector3 point, Vector3 start, Vector3 end)` and reused it in other `GetParameterAt(Vector3 point)`. 

I also added two more unit tests to check if for points almost equal to start or end vectors, method will return rounded values. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809#-:-N2oemtE9sI_ZUh2srRe:beqjy8u)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mwojcik-sagiton)

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @wynged)

___
*[`Elements/src/Geometry/Line.cs` line 1055 at r2](https://reviewable.io/reviews/hypar-io/elements/809#-N2mh4za3-4Mq2VuEXaU:-N2ofdBl3WB-n5jY99y4:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/966498b25360243c169bc528bb48bed32fcc9c97/Elements/src/Geometry/Line.cs#L1055)):*

add curley braces around the if expression.
</blockquote></details>

Done.

wynged (changes requested):
> Reviewed 2 of 2 files at r3, 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mwojcik-sagiton)

___
*[`Elements/src/Geometry/Line.cs` line 1055 at r3](https://reviewable.io/reviews/hypar-io/elements/809#-N2qYMptBsJaAYwSacce:-N2qYMptBsJaAYwSaccf:b-9x5wv5) ([raw file](https://github.com/hypar-io/elements/blob/4d2d96ee3d501fb3a7d8f6956ae9e0a5cc67e2af/Elements/src/Geometry/Line.cs#L1055)):*
> ```Smalltalk
>         public double GetParameterAt(Vector3 point)
>         {
>             if (!PointOnLine(point, true))
> ```

now that we have this check in the method below we don't need it here.

wynged (approved):
> Reviewed 1 of 3 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

wynged (approved):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/809)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

PR #810, opened 2022-05-20
Line merge collinear line

mwojcik-sagiton:
BACKGROUND:
I found `MergeCollinearLine(Line line)` method useful and missing in `Line` class 

DESCRIPTION:
Method `MergeCollinearLine(Line line)` takes all 4 vertices of two lines, sort them and creates line which contains all of them. Finally it keeps direction of original line.

TESTING:
Since there are many similarities to `TryGetOverlap(....)` method I reused unit test cases.  
  
FUTURE WORK:
None

REQUIRED:
- [x] All changes are up to date in `CHANGELOG.md`.

COMMENTS:
None

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/810)

PR #811, opened 2022-05-20
Add Project method to Line

mwojcik-sagiton:
BACKGROUND:
I found `Line.Projected(Plane plane)` useful and missing in `Line` class

DESCRIPTION:
Method `Line.Projected(Plane plane)`  projects `Start` and `End` vectors onto `Plane`. After that it creates new `Line` with projected vectors.

TESTING:
Run unit tests 
  
FUTURE WORK:
None 

REQUIRED:

COMMENTS:
None

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/811)

### DIFF
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -1046,6 +1046,18 @@ namespace Elements.Geometry
         }
 
         /// <summary>
+        /// Projects current line onto a plane
+        /// </summary>
+        /// <param name="plane">Plane to project</param>
+        /// <returns>New line on a plane</returns>
+        public Line Project(Plane plane)
+        {
+            var start = Start.Project(plane);
+            var end = End.Project(plane);
+            return new Line(start, end);
+        }
+
+        /// <summary>
         /// A list of vertices describing the arc for rendering.
         /// </summary>
         internal override IList<Vector3> RenderVertices()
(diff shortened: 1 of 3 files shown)


Elements/src/Geometry/Line.cs
```diff
@@ -1045,6 +1045,18 @@ public bool TryGetOverlap(Line line, out Line overlap)
             return true;
         }
 
+        /// <summary>
+        /// Projects current line onto a plane
+        /// </summary>
+        /// <param name="plane">Plane to project</param>
+        /// <returns>New line on a plane</returns>
+        public Line Project(Plane plane)
```
ikeough, line 1053: Can we call this `Projected`? Similar to my comment in #810, we use past tense in Elements for methods that return a new object.

mwojcik-sagiton: Thank you for suggestion. It was changed in [this commit.](https://github.com/hypar-io/Elements/pull/811/commits/1d9317e07549eb627ffb98912f27e6cadc2e7da5)

wynged (approved):
> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/811)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

PR #812, opened 2022-05-26
16% memory reduction and small speed improvements for glTF creation.

ikeough:
BACKGROUND:

DESCRIPTION:
- This PR removes allocation of new Polygons in the methods that create sweeps for rotated profiles, instead passing the rotation all the way down to the sweep creation to be added to "frame" transforms when they're created. To support this, a new overload of `Transform.Rotate` has been added to rotate around a point.
- This PR adds several `Polygon` and `Polyline` constructors which allow the caller to specify whether they want to disable validation. Places where we have known shapes, like rectangles, L, and structural profiles can then skip validation on construction.
- This PR removes one loop during tessellation which stored all the tessellated vertices.
- This PR removes the `mergeVertices` parameter during tessellation which was not being used.
 
REQUIRED:

Before:
|                                        Method |     Mean |    Error |   StdDev |      Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|---------------------------------------------- |---------:|---------:|---------:|-----------:|----------:|----------:|----------:|
| 'Create all HSS beams and serialize to glTF.' | 74.64 ms | 0.672 ms | 0.561 ms | 12000.0000 | 2000.0000 | 1000.0000 |     45 MB |

After:
|                                        Method |     Mean |    Error |   StdDev |     Gen 0 |     Gen 1 |     Gen 2 | Allocated |
|---------------------------------------------- |---------:|---------:|---------:|----------:|----------:|----------:|----------:|
| 'Create all HSS beams and serialize to glTF.' | 68.47 ms | 1.368 ms | 3.555 ms | 9000.0000 | 2000.0000 | 1000.0000 |     38 MB |

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/812)


mattc (changes requested):
> Reviewed 23 of 24 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/812)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Polygons.cs` line 40 at r2](https://reviewable.io/reviews/hypar-io/elements/812#-N55s52t5JLSSuz4NWwV:-N55s52t5JLSSuz4NWwW:b72mip) ([raw file](https://github.com/hypar-io/elements/blob/6c86a816cb3cacc99abf454bcfa9e0178e043c3d/Elements/src/Geometry/Polygons.cs#L40)):*
> ```Smalltalk
>             var d = new Vector3(min.X, max.Y);
> 
>             return new Polygon(true, new[] { a, b, c, d });
> ```

If using this constructor, why wouldn't this be `new Polygon(true, a, b, c, d)`. (I'm just learning about the `params` keyword).

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/812)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mattc)

___
*[`Elements/src/Geometry/Polygons.cs` line 40 at r2](https://reviewable.io/reviews/hypar-io/elements/812#-N55s52t5JLSSuz4NWwV:-N9-mkxvD6zJkm3T1nhl:b-p89g3u) ([raw file](https://github.com/hypar-io/elements/blob/6c86a816cb3cacc99abf454bcfa9e0178e043c3d/Elements/src/Geometry/Polygons.cs#L40)):*

If using this constructor, why wouldn't this be `new Polygon(true, a, b, c, d)`. (I'm just learning about the `params` keyword).
</blockquote></details>

Yes! All locations where we use this constructor have been updated. I have no recollection why I was calling it that way).

mattc (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/812)* status: :shipit: complete! 1 of 1 approvals obtained

PR #813, opened 2022-05-27
fix profile split with inner voids

andrewheumann:
BACKGROUND:
- `Profile.Split` relies on a half-edge graph, built up from the profile edges and any splitters. We intersect the profile edges and the splitters, and insert half-edges into the graph. 
- However, in some cases, this was failing — throwing an exception on traversal due to hitting a dead end. This dead-end was possible because we only insert a single half-edge for the profile segments, and both half-edges for splitter segments. This is appropriate for the _outer_ boundary of the profile — it helps avoid calculating the outer boundary as a separate polygon — but for the inner boundary, it can fail sometimes.  

DESCRIPTION:
- Stores profile voids as splitters instead of as the outer boundary, so that two half-edges are created for those segments. This can result in the locations within the voids being "found" as polygons, but we use a separate post-processing step to discard those. 

TESTING:
- I added a test which was previously throwing an exception, and verified its results visually: 
image
- All other profile split tests continue to pass. 

REQUIRED:


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/813)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/test/ProfileTests.cs` line 538 at r1](https://reviewable.io/reviews/hypar-io/elements/813#-N3ZMFUr0bv2R5kKo7HT:-N3ZMFUr0bv2R5kKo7HU:bh9e8ex) ([raw file](https://github.com/hypar-io/elements/blob/ca683968b0a7baa88114c3cdac2c5c59efca0cb5/Elements/test/ProfileTests.cs#L538)):*
> ```Smalltalk
>             var profiles = JsonConvert.DeserializeObject<List<Profile>>(profileJson);
>             var segments = JsonConvert.DeserializeObject<List<Line>>(segmentsJson);
>             var splits = Elements.Geometry.Profile.Split(profiles, segments.Select(l => l.ToPolyline(1)));
> ```

even if it's silly, maybe add an Assert about the number of polygons that come out of this to help us catch anything that changes this number in the future?

PR #815, opened 2022-06-02
Loaded types

katehryhorenko:
BACKGROUND:
New types that are inherited from Floor or some other types are exported to the IFC as Objects and are not recognized as Floor or Beam elements. The issue is that during json deserialization inside Exporter class, Exporter generates new classes from schema (e.g. FloorPanal) without inheritance from the Floor class 

DESCRIPTION:
- This PR includes all locally loaded types to the CSharpGeneratorSettings.ExcludedTypeNames. That allows to create proper inheritance. 

TESTING:
I tried:
 - generate IFC
 - CLI init command

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/815)

### DIFF
--- a/Elements.CodeGeneration/src/TypeGenerator.cs
+++ b/Elements.CodeGeneration/src/TypeGenerator.cs
@@ -238,7 +238,7 @@ namespace Elements.Generate
                 try
                 {
                     var schema = await GetSchemaAsync(uri);
-                    var csharpFileContents = GenerateCSharpCodeForSchema(schema);
+                    var csharpFileContents = GenerateCSharpCodeForSchemaExcludingLoadedTypes(schema);
                     foreach (var csharp in csharpFileContents)
                     {
                         if (code.ContainsKey(csharp.Key))
@@ -303,7 +303,7 @@ namespace Elements.Generate
                 try
                 {
                     var schema = await GetSchemaAsync(uri);
-                    var csharpFileContents = GenerateCSharpCodeForSchema(schema);
+                    var csharpFileContents = GenerateCSharpCodeForSchemaExcludingLoadedTypes(schema);
                     foreach (var csharp in csharpFileContents)
                     {
                         if (code.ContainsKey(csharp.Key))
@@ -584,7 +584,7 @@ namespace Elements.Generate
         /// <returns>Returns true if the dll was generated successfully, otherwise false.</returns>
         public static bool GenerateAndSaveDllForSchema(JsonSchema schema, string dllPath, out string[] diagnosticResults, bool frameworkBuild = false)
         {
-            var csharpFileContents = GenerateCSharpCodeForSchema(schema);
+            var csharpFileContents = GenerateCSharpCodeForSchemaExcludingLoadedTypes(schema);
             if (csharpFileContents == null)
             {
                 diagnosticResults = new string[] { };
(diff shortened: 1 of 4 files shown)


wynged (changes requested):
This seems to work well to me.  I tested on one of my functions. It also seems that this code path is not part of normal class code gen, only in other semi-unusual places where we want dynamic types.  In other places the currently available types are normally passed in as excluded types anyways. 
In this case it is used always for in memory/saving dll contexts, so I think this makes sense.  let's just do a little rename.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/815#-:-N3aM_uD33qQoiWswtN1:bwadxtu)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[`Elements.CodeGeneration/src/TypeGenerator.cs` line 597 at r1](https://reviewable.io/reviews/hypar-io/elements/815#-N3aNcx_2vYDjwpNVP4r:-N3aNcx_2vYDjwpNVP4s:b-uwycka) ([raw file](https://github.com/hypar-io/elements/blob/e6dfc9e6991d9d1ce1669073fb49cf3b0af14ce5/Elements.CodeGeneration/src/TypeGenerator.cs#L597)):*
> ```Smalltalk
>         }
> 
>         private static Dictionary<string, string> GenerateCSharpCodeForSchema(JsonSchema schema)
> ```

can we rename this to imply that it inspects the AppDomain for loaded types.  `GenerateCSharpCodeForSchemaExcludingLoadedTypes`  makes it a little more clear how this is differfent than other methods

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/815)* status: :shipit: complete! 1 of 1 approvals obtained

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/815)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[`Elements/src/Floor.cs` line 119 at r2](https://reviewable.io/reviews/hypar-io/elements/815#-N3dqB6oEs-s6IaQA3Uf:-N3dqB6oEs-s6IaQA3Ug:b6v2dla) ([raw file](https://github.com/hypar-io/elements/blob/e5aaee59ac38af5f47d9c8e5288ad01be347bb99/Elements/src/Floor.cs#L119)):*
> ```Smalltalk
>         {
>             if (this.Representation.SolidOperations.Count == 0)
>             {
> ```

@andrewheumann  brought up a good point on this change.  
If the representation exists, and then the profile changes then UpdateRepresentation won't actually update the profile.  is there anothe way around this?  what is the use case?

katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/815)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Floor.cs` line 119 at r2](https://reviewable.io/reviews/hypar-io/elements/815#-N3dqB6oEs-s6IaQA3Uf:-N3eh-Ay52dtgYrXJV-k:bl7ly7b) ([raw file](https://github.com/hypar-io/elements/blob/e5aaee59ac38af5f47d9c8e5288ad01be347bb99/Elements/src/Floor.cs#L119)):*

@andrewheumann  brought up a good point on this change.  
If the representation exists, and then the profile changes then UpdateRepresentation won't actually update the profile.  is there anothe way around this?  what is the use case?

</blockquote></details>

Replaced with updateRepresentation flag inside ToIfc method :+1:

wynged (approved):
good shtuph!

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/815#-:-N3ejFVO4a2_56iqOIdo:bq2e1v9)* status: :shipit: complete! 1 of 1 approvals obtained

PR #816, opened 2022-06-02
Linear dimensions

ikeough:
BACKGROUND:
- For several projects we'd like to create dimensions of elements. @andrewheumann and I decided to make a new element type with no geometry to represent a dimension. Without geometry, any renderer can draw a dimension in the way that works best for that context. Dimension geometry does not get written to the glb.

DESCRIPTION:
- This PR creates a number of new Element types (`Dimension` `LinearDimension` `ContinuousDimension` `AlignedDimension`)for creating linear dimensions. 
- Dimensions can calculate their own value or the `DisplayValue` property can be used. An additional `Prefix` and `Suffix` can be supplied.

TESTING:
- See `DimensionTests`.
  
FUTURE WORK:
- The `ContinuousDimension` is not really continuous currently. It only allows you to supply a start and end point. In the future, we'll want to support multiple points being projected to a line to create truly continuous dimensions.

REQUIRED:

image
image

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)


ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @serenayl)

___
*[`Elements/src/Dimensions/AlignedDimension.cs` line 4 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3eu8ye8BKlHzEWnDSd:-N3knCAj1VEFVey-egEs:bb6un1j) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/AlignedDimension.cs#L4)):*

I liked @serenayl's suggestion to call the base category `Annotations` instead of dimensions  — I'm OK if these continue to inherit from `Dimension` but I sort of like the idea that the namespace should be `Elements.Annotations` — what do you think? 
</blockquote></details>

Done
___
*[`Elements/src/Dimensions/AlignedDimension.cs` line 12 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3evb3PCN1JhfHwPFiK:-N3knPzt3GWikzXhPDmS:b-7ptx45) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/AlignedDimension.cs#L12)):*

I'm not sure we really need separate classes for `AlignedDimension` and `ContinuousDimension`. They have no unique properties that `LinearDimension` does not have. Can we avoid multiplying types for now, and just make these additional constructors or static methods on `LinearDimension`, at least until they do something substantively different? 
</blockquote></details>

I considered using static constructors like `LinearDimension.CreateContinuous` and `LinearDimension.CreateAligned`, but we don't use static constructors elsewhere in elements, using the semantics of the type to convey meaning in the same way that we do with `Beam` and `Brace`. It's a fair argument that using types in this way makes it more challenging to use the serialized result, because you have to say things like "If a thing is an aligned dimension or a continuous dimension then do...", but consider that after you've deserialized a dimension created using a static constructor you no longer have any idea of the original author's intent, you just have a linear dimension. You can infer the intent from whether the dimension has certain properties but then that inference logic has to live in the client, where it can skew from other renderers.
___
*[`Elements/src/Dimensions/AlignedDimension.cs` line 51 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3ey7mo86N1WZw3mCgv:-N3kp9y93_DqrAQW7mSj:bb6un1j) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/AlignedDimension.cs#L51)):*

I think it's more likely we'll want an offset and not a plane than the other way around. Can we swap the order of these two parameters?
</blockquote></details>

Done
___
*[`Elements/src/Dimensions/Dimension.cs` line 13 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3eyXlvEMrE9aaNSJL5:-N3ksXzJAoX8g54FiPWq:b-nnlrzr) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/Dimension.cs#L13)):*

I'm not convinced about the necessity of this value... I don't think we need to serialize two planes for every dimension. This value is really only used here in various construction strategies — I'm fine with it as an argument, but I think it results in "too much information" in a way that might be confusing or conflicting. 

The plane of the dimension should be the plane passing through the points in question and their projection onto the reference plane. I don't use the `Plane` property at all when drawing this in Explore, and you don't even use it in `ToModelArrowsAndText` except to fetch a normal, which could easily be computed.  



Unless perhaps there is some application of this plane that I am missing — totally willing to be convinced otherwise, but in its current use it seems more confusing than helpful. 

</blockquote></details>

I don't use it in `ToModelArrowsAndText` because it's already been used in the constructor to project the provided points of start, end, and the dimension line, to the plane. It's fair to say for this reason that it doesn't have to be serialized, only passed in through the constructor, but it needs to be provided. Otherwise, how do we determine what plane the annotation will line on? The "implicit plane" that you show above could be twisted because it's defined by four points. You're probably right, but I'm trying to understand how, given a start and end that form one line and dimension line which is not is the same plane, how we decide where the annotation will actually be drawn, without the user specifying a specific plane.
___
*[`Elements/src/Dimensions/LinearDimension.cs` line 23 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3ex32yDwCiByfy7jHK:-N3kpVEZC9dWgjD1Yozr:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/LinearDimension.cs#L23)):*

"The plane on which"
</blockquote></details>

Done.
___
*[`Elements/src/Dimensions/LinearDimension.cs` line 111 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3ew3V-AvVs1XmtG9nc:-N3kpzkLEJT-V20iqSjV:bucgdom) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/LinearDimension.cs#L111)):*

Can we put `dimensions` first and give `color` a default?
</blockquote></details>

I reordered the parameters, but we can't give color a default, because the value needs to be a compile time constant (which structs can't be). However, I have it the value "default", so it'll be black (but also transparent which shouldn't matter in this case).

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @serenayl)

___
*[`Elements/src/Dimensions/AlignedDimension.cs` line 12 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3evb3PCN1JhfHwPFiK:-N3kuO7y5Hcc7uGrRLIv:bd0d1dg) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/AlignedDimension.cs#L12)):*

I considered using static constructors like `LinearDimension.CreateContinuous` and `LinearDimension.CreateAligned`, but we don't use static constructors elsewhere in elements, using the semantics of the type to convey meaning in the same way that we do with `Beam` and `Brace`. It's a fair argument that using types in this way makes it more challenging to use the serialized result, because you have to say things like "If a thing is an aligned dimension or a continuous dimension then do...", but consider that after you've deserialized a dimension created using a static constructor you no longer have any idea of the original author's intent, you just have a linear dimension. You can infer the intent from whether the dimension has certain properties but then that inference logic has to live in the client, where it can skew from other renderers.
</blockquote></details>

I also didn't do this because IFC is littered with enumerations used to convey meaning, like how a beam should be used (lintel, brace, etc.). If types are cheap, why not use them?

andrewheumann (changes requested):
> Reviewed 2 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Dimensions/AlignedDimension.cs` line 12 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3evb3PCN1JhfHwPFiK:-N3nWkgi2U410_tewQyR:b-vdbsx5) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/AlignedDimension.cs#L12)):*

I also didn't do this because IFC is littered with enumerations used to convey meaning, like how a beam should be used (lintel, brace, etc.). If types are cheap, why not use them?
</blockquote></details>

OK! I buy that argument. 
___
*[`Elements/src/Dimensions/Dimension.cs` line 13 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3eyXlvEMrE9aaNSJL5:-N3nV4VHA6QftCmLme-e:b-6xf7ox) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/Dimension.cs#L13)):*

I don't use it in `ToModelArrowsAndText` because it's already been used in the constructor to project the provided points of start, end, and the dimension line, to the plane. It's fair to say for this reason that it doesn't have to be serialized, only passed in through the constructor, but it needs to be provided. Otherwise, how do we determine what plane the annotation will line on? The "implicit plane" that you show above could be twisted because it's defined by four points. You're probably right, but I'm trying to understand how, given a start and end that form one line and dimension line which is not is the same plane, how we decide where the annotation will actually be drawn, without the user specifying a specific plane.
</blockquote></details>

The implicit plane can't be twisted, since the start and end are projected along parallel vectors. That guarantees that the projection of "end" lies in the same plane as the projection of "start", "start", and "end". I'm all for having it as a constructor parameter (in cases where the user doesn't supply the reference plane), I just don't think it should be a top-level property.
___
*[`Elements/src/Dimensions/LinearDimension.cs` line 111 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3ew3V-AvVs1XmtG9nc:-N3nWSvo6SoYGGDckCE0:bhspfse) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/LinearDimension.cs#L111)):*

I reordered the parameters, but we can't give color a default, because the value needs to be a compile time constant (which structs can't be). However, I have it the value "default", so it'll be black (but also transparent which shouldn't matter in this case).
</blockquote></details>

Why doesn't transparent matter in this case?

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[`Elements/src/Dimensions/Dimension.cs` line 13 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3eyXlvEMrE9aaNSJL5:-N3qn5QA09DdT0Lh1_tm:bp0nd5y) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/Dimension.cs#L13)):*

The implicit plane can't be twisted, since the start and end are projected along parallel vectors. That guarantees that the projection of "end" lies in the same plane as the projection of "start", "start", and "end". I'm all for having it as a constructor parameter (in cases where the user doesn't supply the reference plane), I just don't think it should be a top-level property.
</blockquote></details>

Projected along parallel vectors onto what? We only allow the user to supply a reference line, which has an infinite number of planes around it. We can only construct the reference plane by creating it orthogonal to the plane in which the dimension will lie. I think I'm missing something about how you assume the reference plane is created.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Dimensions/LinearDimension.cs` line 111 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3ew3V-AvVs1XmtG9nc:-N3qoNt_2JrkcvDANYnV:bx700n5) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/LinearDimension.cs#L111)):*

Why doesn't transparent matter in this case? 
</blockquote></details>

I assume it doesn't matter because we're drawing lines, but I could be wrong. It's worth a check. In any case, I'll check the argument and force it to black inside the constructor.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Dimensions/Dimension.cs` line 13 at r1](https://reviewable.io/reviews/hypar-io/elements/816#-N3eyXlvEMrE9aaNSJL5:-N3u1jYWEulu6dCZxc26:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/a02be26b437ae366808cbd730751f5b9bda3789e/Elements/src/Dimensions/Dimension.cs#L13)):*

The `Plane` can exist as a constructor parameter — you can construct your `ReferencePlane` exactly as you're doing now, based on a user-supplied plane. But once we have computed the reference plane we don't need the `Plane` anymore — the points + the `ReferencePlane` are sufficient to compute the `Plane`. And having both introduces the possibility of contradiction. 
</blockquote></details>

Done.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[`Elements/src/Annotations/LinearDimension.cs` line 90 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3uIPMd8FAwC64y_Mdg:-N3uIPMd8FAwC64y_Mdh:b-nrq7pk) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/LinearDimension.cs#L90)):*
> ```Smalltalk
>             modelArrowData.Add((dimStart, dimDirection, dimStart.DistanceTo(dimEnd), color));
> 
>             var c = new Material("Red", color, unlit: true);
> ```

Maybe call this something other than "Red" :)

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

___
*[`Elements/src/Annotations/AlignedDimension.cs` line 58 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3xsgJc2bic4sjm3FfJ:-N3xsgJd5Yk8BP801QPJ:b8mpm34) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/AlignedDimension.cs#L58)):*
> ```Smalltalk
>             this.End = end.Project(plane);
>             var vRef = (this.End - this.Start).Unitized();
>             var offsetDirection = vRef.Cross(plane.Normal);
> ```

If `offsetDirection` has a length of zero (the default behavior if no plane is supplied and the line is vertical) we should fall back to some default direction, like the X axis, so that we don't throw an exception at the next step.

andrewheumann (commented):
(and one other note about a fallback for 0-length vector below)

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816#-:-N3xsvoY4wWgk0ce-I29:b-mzflbx)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @ikeough)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Annotations/AlignedDimension.cs` line 58 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3xsgJc2bic4sjm3FfJ:-N4-wmAo8e6Co4WRI6Lw:b-3r3osu) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/AlignedDimension.cs#L58)):*

If `offsetDirection` has a length of zero (the default behavior if no plane is supplied and the line is vertical) we should fall back to some default direction, like the X axis, so that we don't throw an exception at the next step. 
</blockquote></details>

This feels like a case where the user would want to know that we did this. I'm going to throw an exception instead.
___
*[`Elements/src/Annotations/LinearDimension.cs` line 90 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3uIPMd8FAwC64y_Mdg:-N4-mFrCB7OOeCZhzAPg:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/LinearDimension.cs#L90)):*

Maybe call this something other than "Red" :) 
</blockquote></details>

Done.

andrewheumann (changes requested):
> Reviewed 1 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/test/DimensionTests.cs` line 113 at r6](https://reviewable.io/reviews/hypar-io/elements/816#-N42kfxM15NizOEayZ-z:-N42kfxM15NizOEayZ0-:b48g8zb) ([raw file](https://github.com/hypar-io/elements/blob/9f3db8d2bbe80bccd0e05aa4b7a19a006308201b/Elements/test/DimensionTests.cs#L113)):*
> ```Smalltalk
>         public void VerticalLineNoPlaneOffsetDirectionIsX()
>         {
>             Assert.Throws<Exception>(() => new AlignedDimension(Vector3.Origin, new Vector3(0, 0, 1)));
> ```

Again, see my comment above... why should we want this to throw an exception? 
___
*[`Elements/src/Annotations/AlignedDimension.cs` line 58 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3xsgJc2bic4sjm3FfJ:-N42jP_mBmXhinPR45Hq:b-cz4k0m) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/AlignedDimension.cs#L58)):*

This feels like a case where the user would want to know that we did this. I'm going to throw an exception instead.
</blockquote></details>

I'm going to keep being stubborn here — I just want a function author to be able to easily say `new AlignedDimension(ptA, ptB, 2.0)` and always get a sensible result, regardless of where the points are. There's no reason for that set of arguments to throw an exception. I'm fine with throwing an exception if the arguments are bad (if the user has supplied a plane, and the points are perpendicular to it) but if we auto generated it, it shouldn't throw an exception. So maybe the answer is to leave the exception have you have it, but adjust the automatic plane logic above to not choose the Z axis as a normal if the points are in the Z: 

```
var vRef = (this.End - this.Start).Unitized();            
if (plane == null)
   {
      var normal = Math.abs(vref.Z) > 1 - Vector3.EPSILON ? Vector3.XAxis : Vector3.ZAxis;
      plane = new Plane(Vector3.Origin, normal);
   }

```

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/test/DimensionTests.cs` line 113 at r6](https://reviewable.io/reviews/hypar-io/elements/816#-N42kfxM15NizOEayZ-z:-N49BJ_J07NXBB90PC7J:bsmhvmq) ([raw file](https://github.com/hypar-io/elements/blob/1ae4b5b1f7b185214ef00329816d951232d9dcae/Elements/test/DimensionTests.cs#L113)):*

Again, see my comment above... why should we want this to throw an exception? 
</blockquote></details>

Because it's exceptional. 
___
*[`Elements/src/Annotations/AlignedDimension.cs` line 58 at r4](https://reviewable.io/reviews/hypar-io/elements/816#-N3xsgJc2bic4sjm3FfJ:-N49ZBRk-SqWW9KsXg8B:b-cow8qj) ([raw file](https://github.com/hypar-io/elements/blob/ec08025cb02baa3c8977d05e8d39287cb922457d/Elements/src/Annotations/AlignedDimension.cs#L58)):*

I'm going to keep being stubborn here — I just want a function author to be able to easily say `new AlignedDimension(ptA, ptB, 2.0)` and always get a sensible result, regardless of where the points are. There's no reason for that set of arguments to throw an exception. I'm fine with throwing an exception if the arguments are bad (if the user has supplied a plane, and the points are perpendicular to it) but if we auto generated it, it shouldn't throw an exception. So maybe the answer is to leave the exception have you have it, but adjust the automatic plane logic above to not choose the Z axis as a normal if the points are in the Z: 

```
var vRef = (this.End - this.Start).Unitized();            
if (plane == null)
   {
      var normal = Math.abs(vref.Z) > 1 - Vector3.EPSILON ? Vector3.XAxis : Vector3.ZAxis;
      plane = new Plane(Vector3.Origin, normal);
   }

```
</blockquote></details>

Taking another look at this, I think we can simplify towards your original idea. I've removed the `plane` argument from both constructors. For an `AlignedDimension` you can now simply provide an offset vector or, if one is not supplied one will be constructed by picking the binormal appropriately. For a `ContinuousDimension` the implicit plane is constructed using the same logic for picking the binormal, which is then used with the dimension line to construct the reference plane.

Long story short, this removes all of the `Project` calls we were doing earlier. It doesn't throw exceptions when the optional parameters are not included, instead providing a consistently oriented dimension.

andrewheumann (approved):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/816)* status: :shipit: complete! 1 of 1 approvals obtained

PR #821, opened 2022-06-15
More fast AreCollinear from private section of AdaptiveGridRouting. Add infinite to Line.Trim

DmytroMuravskyi:
BACKGROUND:
- AdaptiveGraphRouting has AreCollinearInplace - optimized and cheap collinear check that avoid creating (and validating in constructor) of new Vector3 objects and approximately 6 time faster for the version that uses fit line.
- AreCollinear for 3 point used dot approach as well before but not it also uses fit line approach.
- I also has a function in external project that was basically a copy of Line.Trim but uses infinite line. It was required to find all possible intersection with polygon.
- I think that this might be a good opportunity to make the code I have public so I did this PR as a proposal.

DESCRIPTION:
- I think both versions of AreCollinear are valid if it's clear what they do. AreCollinearByDistance uses fit line and guarantees that all points are no more than tolerance away from a common line. AreCollinearByAngle is simpliear and guarantees that  3 points forms the same unsigned angle within certain tolerance (0.001 degrees by default).
- Infinite parameter fits well in Line.Trim since many functions have this parameter and it was not requiring to much change.  

TESTING:
- Added LineTrimWithPolygonInfinite test for Line.Trim
- Added CollinearByAngle for AreCollinearByAngle

FUTURE WORK:
- There is no AreCollinearByAngle for a list of point. One can be created if required. I did not added it because it's not so trivial as I initially thought.
- I added  COS_ANGLE_EPSILON = 0.99999999984769128; as cos of 0.001, so AreCollinear is bound it some real threshold but it still feels complicated since it can be used only in cases when cos is expected.  

REQUIRED:

COMMENTS:
- Any other notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Changed
 
 - `AdaptiveGraphRouting` how recognizes edges as affected by hint line of the same direction if part of it is close enough.
+- `Vector3.AreCollinear` are renamed into `Vector3.AreCollinearByDistance` and added `tolerance` parameter.
+- `Line.Trim` - added `infinite` for the cases when line needs to be treated as infinite.
 
 ### Fixed
 
@@ -17,6 +19,8 @@
 - `LinearDimension`
 - `AlignedDimension`
 - `ContinuousDimension`
+- `Vector3.AreCollinearByAngle(Vector3 a, Vector3 b, Vector3 c, double tolerance)`
+
 
 ### Fixed
 
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -307,13 +307,13 @@ namespace Elements.Geometry
             // construct a plane through this line and the start or end of the other line
             Plane plane;
             Vector3 testpoint;
-            if (!(new[] { start1, end1, start2 }).AreCollinear())
+            if (!(new[] { start1, end1, start2 }).AreCollinearByDistance())
             {
                 plane = new Plane(start1, end1, start2);
                 testpoint = end2;
 
             } // this only occurs in the rare case that the start point of the other line is collinear with this line (still need to generate a plane)
-            else if (!(new[] { start1, end1, end2 }).AreCollinear())
+            else if (!(new[] { start1, end1, end2 }).AreCollinearByDistance())
             {
                 plane = new Plane(start1, end1, end2);
                 testpoint = start2;
@@ -705,7 +705,7 @@ namespace Elements.Geometry
                 // We want to extend only to the first corner of the other lines,
                 // not all the way through to the other end
                 if (segment.Direction().IsParallelTo(testLine.Direction(), tolerance) && // if the two lines are parallel
-                    (new[] { segment.End, segment.Start, testLine.Start, testLine.End }).AreCollinear())// and collinear
+                    (new[] { segment.End, segment.Start, testLine.Start, testLine.End }).AreCollinearByDistance())// and collinear
                 {
                     if (!this.PointOnLine(segment.End, true))
                     {
@@ -838,19 +838,26 @@ namespace Elements.Geometry
         /// <param name="polygon">The polygon to trim with.</param>
         /// <param name="outsideSegments">A list of the segment(s) of the line outside of the supplied polygon.</param>
         /// <param name="includeCoincidenceAtEdge">Include coincidence at edge as inner segment.</param>
+        /// <param name="infinite"></param>
         /// <returns>A list of the segment(s) of the line within the supplied polygon.</returns>
-        public List<Line> Trim(Polygon polygon, out List<Line> outsideSegments, bool includeCoincidenceAtEdge = false)
+        public List<Line> Trim(Polygon polygon, out List<Line> outsideSegments, bool includeCoincidenceAtEdge = false, bool infinite = false)
         {
             // adapted from http://csharphelper.com/blog/2016/01/clip-a-line-segment-to-a-polygon-in-c/
             // Make lists to hold points of intersection
             var intersections = new List<Vector3>();
 
-            // Add the segment's starting point.
-            intersections.Add(this.Start);
-            polygon.Contains(this.Start, out var containment);
-            var StartsOutsidePolygon = containment == Containment.Outside;
+            var startsOutsidePolygon = false;
+            var hasVertexIntersections = false;
+            var containment = Containment.Outside;
 
-            var hasVertexIntersections = containment == Containment.CoincidesAtVertex;
+            if (!infinite)
+            {
+                // Add the segment's starting point.
+                intersections.Add(this.Start);
+                polygon.Contains(this.Start, out containment);
+                startsOutsidePolygon = containment == Containment.Outside;
+                hasVertexIntersections = containment == Containment.CoincidesAtVertex;
+            }
 
             // Examine the polygon's edges.
             for (int i1 = 0; i1 < polygon.Vertices.Count; i1++)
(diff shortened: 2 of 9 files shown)


wynged (changes requested):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/821)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Geometry/Line.cs` line 843 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4hB0cf0Mkccq_K4zY4:-N4hB0cg3oT6Sgmo1xzy:bv33jxy) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Line.cs#L843)):*
> ```Smalltalk
>         /// <param name="infinite">Treat the line as infinite?</param>
>         /// <returns>A list of the segment(s) of the line within the supplied polygon.</returns>
>         public List<Line> Trim(Polygon polygon, out List<Line> outsideSegments, bool includeCoincidenceAtEdge = false, bool infinite = false)
> ```

I'm not sure about making Trim behave this way, we might want a different method to handle this case just for  the same of naming.  Asking @andrewheumann  to chime in here.  
___
*[`Elements/src/Geometry/Vector3.cs` line 829 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h8TcUA7l-C4ylg-7w:-N4h8TcUA7l-C4ylg-7x:b22ppo0) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3.cs#L829)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Check whether three points are on the same line within certain distance.
> ```

Add a note that if you want very high performance use the AreCollinearByAngle method instead.
___
*[`Elements/src/Geometry/Vector3.cs` line 836 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h7c744Fdemqnw0weo:-N4h7c744Fdemqnw0wep:b-lgcpho) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3.cs#L836)):*
> ```Smalltalk
>         /// <param name="tolerance">Distance tolerance.</param>
>         /// <returns>True if the points are on the same line, false otherwise.</returns>
>         public static bool AreCollinearByDistance(Vector3 a, Vector3 b, Vector3 c, double tolerance = Vector3.EPSILON) 
> ```

For backwards compatibility I think we should add a method just called AreCollinear that called this AreCollinearByDistance.  Also, mark that method as deprecated and suggest that the user use one of these new methods.
___
*[`Elements/src/Geometry/Vector3.cs` line 845 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h7u5H2z4B8gm1YTV0:-N4h7u5H2z4B8gm1YTV1:b671kwr) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3.cs#L845)):*
> ```Smalltalk
>         /// Check whether three points are on the same line within certain angle.
>         /// Order is important since unsigned abc angle is checked.
>         /// This function is much faster than AreCollinearByDistance but angle deviation 
> ```

change this sentence a little bit. I suggest: 
"This function is much faster than AreCollinearByDistance but angle deviation scales with the distance of points being compared.  If points are far away from each other they might appear collinear even if there are large distance offsets between them."
___
*[`Elements/src/Geometry/Vector3.cs` line 853 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h8chYAYmN2ExNluue:-N4h8chYAYmN2ExNluuf:bbrd2gy) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3.cs#L853)):*
> ```Smalltalk
>         /// <param name="tolerance">Angle tolerance as cos.</param>
>         public static bool AreCollinearByAngle(Vector3 a, Vector3 b, Vector3 c, double tolerance = Vector3.COS_ANGLE_EPSILON)
> ```

can we name the argument `cosAngleTolerance` ?
___
*[`Elements/src/Geometry/Vector3.cs` line 853 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4hA6_u2N_mmeZunA3V:-N4hA6_u2N_mmeZunA3W:b-7quiwd) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3.cs#L853)):*
> ```Smalltalk
>         /// <param name="tolerance">Angle tolerance as cos.</param>
>         public static bool AreCollinearByAngle(Vector3 a, Vector3 b, Vector3 c, double tolerance = Vector3.COS_ANGLE_EPSILON)
> ```

In a test below you do Math.Cos(Units.RadToDeg() to input an angle tolerance.  Why can't we do that here as well?  Is it because that will hurt our performance and this is supposed to be the fast method?
If we had a different process for setting tolerance that might be ok.  Maybe a method called SetAngleTolerance() that sent an internal variable and if this was present it was used instead of the Const tolerance.  
___
*[`Elements/src/Geometry/Vector3Extensions.cs` line 46 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h8iIb5icQRUEhmb05:-N4h8iIb5icQRUEhmb06:bn3buam) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3Extensions.cs#L46)):*
> ```Smalltalk
>         /// <param name="points">List of points to check. Order is not important.</param>
>         /// <param name="tolerance">Distance tolerance.</param>
>         public static bool AreCollinearByDistance(this IList<Vector3> points, double tolerance = Vector3.EPSILON)
> ```

Same note as above about adding a deprecated method `AreCollinear` that simply calls this method. for backwards compabibility
___
*[`Elements/src/Geometry/Vector3Extensions.cs` line 67 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h91mv8PAzo02tImRe:-N4h91mv8PAzo02tImRf:b6nipic) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Vector3Extensions.cs#L67)):*
> ```Smalltalk
>                 var lengthSquared = d.LengthSquared();
>                 // By Pythagoras' theorem d.Length()^2 = dot^2 + distance^2. 
>                 // if it's less than tolerance squared - point is close enough to the fit line.
> ```

Nice additional note.
capatalize "If" and replace "-" with "then the"  
___
*[`Elements/test/LineTests.cs` line 499 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4h9eOK2G6LSKhvRgIA:-N4h9eOK2G6LSKhvRgIB:b-pvqs0f) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/test/LineTests.cs#L499)):*
> ```Smalltalk
>                   (-6, Vector3.EPSILON * 2, 0)
>             );
>             Assert.False(points1.AreCollinearByDistance());
> ```

just a note that even though we'll add deprecated methods for backwards compatibility I think we should do this change and use the new method names in tests.

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/821)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Geometry/Line.cs` line 843 at r1](https://reviewable.io/reviews/hypar-io/elements/821#-N4hB0cf0Mkccq_K4zY4:-N4hBlxA36HCMFHKM6Zp:byal8mn) ([raw file](https://github.com/hypar-io/elements/blob/beb6f5e91f248810c15caaff7ada760ddd49efc0/Elements/src/Geometry/Line.cs#L843)):*

I'm not sure about making Trim behave this way, we might want a different method to handle this case just for  the same of naming.  Asking @andrewheumann  to chime in here.  
</blockquote></details>

Andrew has informed me that this is consistent with other methods, so it works for me.

PR #833, opened 2022-07-19
Public Edge and Vertex constructors 

mwojcik-sagiton:
BACKGROUND:
I wanted to optimize performance and use `Elements.Geometry.Solids.Edge` class instead of `Line`, unfortunately there is not any public constructor for it. 

DESCRIPTION:
Change `Elements.Geometry.Solids.Edge` and `Elements.Geometry.Solids.Vertex` constructors from internal to public.

TESTING:
Run unit tests 
  
FUTURE WORK:
No additional work is required 

REQUIRED:

COMMENTS:
There isn't any

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/833)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 - `Vector3.AreCollinear` are renamed into `Vector3.AreCollinearByDistance` and added `tolerance` parameter.
 - `Line.Trim` - added `infinite` for the cases when line needs to be treated as infinite.
 - `Vector3.ClosestPointOn` - added `infinite` for the cases when line needs to be treated as infinite.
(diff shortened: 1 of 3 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/833)* status: :shipit: complete! 1 of 1 approvals obtained

ikeough:
In what scenario does this improve performance? In other locations we're moving to tuples of vectors for edge-like functionality. I'm wondering if we can do the same here. Access to these class constructors was previous restricted because they didn't make sense to use outside the context of a solid.

PR #836, opened 2022-07-21
Change Line.PointOnLine

DmytroMuravskyi:
BACKGROUND:
Vector3.PointOnLine is based on dot product and because of that answer is angle based. The longer the line is - the bigger deviation is allowed. For example Line((0, 0, 0), (150, 0, 0) will return true for (100, 0.1, 0) point that can lead to unwanted results.
Because how includeEnds parameter is coded it's unclear what it means: when it's false exactly end point will always return true, but when numeric noise is present: (1e-6, 0, 0) will return True for the line above but (1e-6, 1e-6, 0) will return false.

DESCRIPTION:
PointOnLine is now distance based and is based on calculations from Vector3.DistanceTo(Line): return true if point is less than tolerance away from the line.
New implementation is not using Sqrt so it's quicker than old one although it uses extra checks.
New implementation has similar behavior for end points: when includeEnds is false the exact same point will return false but both (1e-6, 0, 0) and (1e-6, 1e-6, 0) will not return true because they both less than tolerance away.

TESTING:
- Add PointOnLine test to LineTest.
- Adjust results of SplitComplexProfileWithInnerVoids test
- Increase precision of BestFitLine test.

FUTURE WORK:
- There may be other functions based on dot product that are error prone based on inputs size.
 
REQUIRED:

COMMENTS:
- Active section of `CHANGELOG.md` had wrong label 1.0.2 although target version is 1.1.0. It happened because 1.0.2 was skipped in changelog when it was developed. This is fixed now.
- I resolved leftover minor warnings, so the only one left is about using pre-selease version of SixLabours.Fonts. If we could resolve it as well we would be warnings free.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/836)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.0.2
+## 1.1.0
 
 ### Added
 
@@ -22,6 +22,7 @@
 - `Vector3.ClosestPointOn` - added `infinite` for the cases when line needs to be treated as infinite.
 - `Elements.Geometry.Solids.Edge` public constructor
 - `Elements.Geometry.Solids.Vertex` public constructor
+- `Line.PointOnLine` now uses distance to line instead of dot product. 
 
 ### Fixed
 
--- a/Elements/src/Beam.cs
+++ b/Elements/src/Beam.cs
@@ -59,6 +59,9 @@ namespace Elements
                     string name = null) : base(curve, profile, material, startSetback, endSetback, rotation, transform, representation, isElementDefinition, id, name)
         { }
 
+        /// <summary>
+        /// Construct an empty beam.
+        /// </summary>
         public Beam() { }
     }
 }
--- a/Elements/src/Beam.cs
+++ b/Elements/src/Beam.cs
@@ -60,7 +60,7 @@ namespace Elements
         { }
 
         /// <summary>
-        /// Construct an empty beam.
+        /// Construct a beam.
         /// </summary>
         public Beam() { }
     }
--- a/Elements/src/Column.cs
+++ b/Elements/src/Column.cs
@@ -83,6 +83,9 @@ namespace Elements
             this.Height = height;
         }
 
+        /// <summary>
+        /// Construct empty Column.
+        /// </summary>
         public Column()
         {
         }
--- a/Elements/src/Column.cs
+++ b/Elements/src/Column.cs
@@ -84,7 +84,7 @@ namespace Elements
         }
 
         /// <summary>
-        /// Construct empty Column.
+        /// Construct a column.
         /// </summary>
         public Column()
         {
--- a/Elements/src/Geometry/Line.cs
+++ b/Elements/src/Geometry/Line.cs
@@ -480,22 +480,26 @@ namespace Elements.Geometry
         }
 
         /// <summary>
-        /// Test if a point lies within this line segment
+        /// Test if a point lies within a tolerance distance from given line segment
         /// </summary>
         /// <param name="point">The point to test.</param>
-        /// <param name="includeEnds">Consider a point at the endpoint as on the line.</param>
+        /// <param name="includeEnds">Consider a point at the endpoint as on the line.
+        /// When true - point  tolerance away from end points considered on line.
+        /// When false - point should be inside the line, but not exactly above end points.</param>
         public bool PointOnLine(Vector3 point, bool includeEnds = false)
         {
             return Line.PointOnLine(point, Start, End, includeEnds);
         }
 
         /// <summary>
-        /// Test if a point lies within a given line segment
+        /// Test if a point lies within a tolerance distance from given line segment
         /// </summary>
         /// <param name="point">The point to test.</param>
         /// <param name="start">The start point of the line segment.</param>
         /// <param name="end">The end point of the line segment.</param>
-        /// <param name="includeEnds">Consider a point at the endpoint as on the line.</param>
+        /// <param name="includeEnds">Consider a point at the endpoint as on the line.
+        /// When true - point  tolerance away from end points considered on line.
+        /// When false - point should be inside the line, but not exactly above end points.</param>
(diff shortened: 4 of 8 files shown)


andrewheumann (changes requested):
Looks great — thanks for all the additional warnings fixes as well! Just a few minor wording adjustments. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/836#-:-N7b6XxG9SdqmE6uZl9U:b-mw0t4k)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Beam.cs` line 63 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b3wCX2ti0Zq9M6x9C:-N7b3wCX2ti0Zq9M6x9D:bbog1l0) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Beam.cs#L63)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Construct an empty beam.
> ```

`Construct a beam.`
___
*[`Elements/src/Column.cs` line 87 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b4-PcBAdHmekh1On0:-N7b4-PcBAdHmekh1On1:b-vg8ic5) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Column.cs#L87)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Construct empty Column.
> ```

`Construct a column.`
___
*[`Elements/src/StructuralFraming.cs` line 74 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b41CRCaKAYyd1gcRB:-N7b41CRCaKAYyd1gcRC:b-1n0a17) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/StructuralFraming.cs#L74)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Construct empty beam.
> ```

`Construct a framing element.`
___
*[`Elements/src/Geometry/Line.cs` line 483 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b4jZvBetPDHsREbWg:-N7b4jZvBetPDHsREbWh:btvtk8e) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Geometry/Line.cs#L483)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Test if a point lies within a tolerance distance from given line segment
> ```

`Test if a point lies within tolerance of this line segment`
___
*[`Elements/src/Geometry/Line.cs` line 487 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b4HPN6zoaJb3CMsF-:-N7b4HPN6zoaJb3CMsF0:byvahyw) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Geometry/Line.cs#L487)):*
> ```Smalltalk
>         /// <param name="point">The point to test.</param>
>         /// <param name="includeEnds">Consider a point at the endpoint as on the line.
>         /// When true - point  tolerance away from end points considered on line.
> ```

```
When true, any point within tolerance of the end points will be considered on the line.
When false, points precisely at the ends of the line will not be considered on the line. 
```
___
*[`Elements/src/Geometry/Line.cs` line 495 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b5ea6DI9KUlTu4JkZ:-N7b5ea6DI9KUlTu4Jk_:bb7ehfk) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Geometry/Line.cs#L495)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Test if a point lies within a tolerance distance from given line segment
> ```

`Test if a point lies within tolerance of a given line segment`
___
*[`Elements/src/Geometry/Line.cs` line 502 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b5gAZ8OxZHOUjieeY:-N7b5gAZ8OxZHOUjieeZ:byvahyw) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/src/Geometry/Line.cs#L502)):*
> ```Smalltalk
>         /// <param name="includeEnds">Consider a point at the endpoint as on the line.
>         /// When true - point  tolerance away from end points considered on line.
>         /// When false - point should be inside the line, but not exactly above end points.</param>
> ```

```
When true, any point within tolerance of the end points will be considered on the line.
When false, points precisely at the ends of the line will not be considered on the line. 
```
___
*[`Elements/test/LineTests.cs` line 748 at r1](https://reviewable.io/reviews/hypar-io/elements/836#-N7b6Ft3COE3C0sAzv-S:-N7b6Ft3COE3C0sAzv-T:b-uwu7rb) ([raw file](https://github.com/hypar-io/elements/blob/3402d8c1b2cf9f8cabdc20f8b409fed7602be74e/Elements/test/LineTests.cs#L748)):*
> ```Smalltalk
>             Assert.False(Line.PointOnLine(start, start, end));
>             Assert.True(Line.PointOnLine(start, start, end, true));
>             Assert.True(Line.PointOnLine(start, start, end, true));
> ```

I think there are some duplicate lines in these tests

andrewheumann (approved):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/836)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @DmytroMuravskyi)

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/836)* status: :shipit: complete! 1 of 1 approvals obtained

PR #837, opened 2022-07-21
Add DoubleToleranceComparer

DmytroMuravskyi:
BACKGROUND:
- AdaptiveGrid needed a way to group items by Z coordinate within certain tolerance.

DESCRIPTION:
- Added DoubleToleranceComparer that treats all numbers within tolerance as the same.
This comparer doesn't use hash code as it is *impossible* to create a hashing algorithm that consistently returns identical values for any two points within tolerance of each other.
Note that any item that is not within tolerance of other key is set as a new key.
This mean that keys are not always whole number in the middle of a range.

TESTING:
- Added DoubleToleranceComparerTests test.
 
REQUIRED:

COMMENTS:
- These is Vector3Comparer class that is similar but it uses GetHashCode that make it volatile in behavior. It may be considered to make it also always use 0 as hash code.

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/837)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - `Vector3Extensions.BestFitLine(this IList<Vector3> points)`
 - `Polygon.FromAlignedBoundingBox2d(IEnumerable<Vector3> points, Vector3 axis, double minSideSize = 0.1)`
 - `Transform.RotateAboutPoint` and `Transform.RotatedAboutPoint` convenience methods.
+- `DoubleToleranceComparer`
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/837)* status: :shipit: complete! 1 of 1 approvals obtained

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/837)* status: :shipit: complete! 1 of 1 approvals obtained

PR #838, opened 2022-07-22
Fix tolerance for AreCoplanar

andrewheumann:
BACKGROUND:
- The `AreCoplanar` extension method for collections of vertices had an error that caused it to be potentially several orders of magnitude more sensitive to deviation than it was intended to be. Specifically, checking the dot product with the un-unitized cross product meant that the deviation was multiplied by the area of the parallelogram between the first two edges, which could be quite a large number, effectively tightening the tolerance of the method.

DESCRIPTION:
- Fixes the error and adds a test

TESTING:
- A new test deserializing a polygon with a deviation in the z of 6.3e-7 is added
- All other tests are passing
  
FUTURE WORK:

REQUIRED:

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/838)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.0.2
+## 1.1.0
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


cdriesler (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/838)* status: :shipit: complete! 1 of 1 approvals obtained

PR #846, opened 2022-08-01
Add GetSharedSegments method for Polygon

mwojcik-sagiton:
BACKGROUND:
I didn't find method which will return `Polyline` sub segments which are inside of `Polygon`

DESCRIPTION:
Method finds intersections between `Polygon` and `Polyline` and based on that result it creates sub segments of `Polyline`. I added new method to `Polyline` creating sub segment between two points. It will always respect direction of `Polyline` even if points are in reversed order. 

TESTING:
Run unit tests 
  
FUTURE WORK:
None

REQUIRED:

COMMENTS:
None

---
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/846)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/846)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton)

___
*[`Elements/src/Geometry/Polygon.cs` line 2383 at r1](https://reviewable.io/reviews/hypar-io/elements/846#-N8TcQxY3ulq4rP5EnmB:-N8TcQxY3ulq4rP5EnmC:b-kbb2nk) ([raw file](https://github.com/hypar-io/elements/blob/fdabb3d3834be90cae34a1a625ad275440d72606/Elements/src/Geometry/Polygon.cs#L2383)):*
> ```Smalltalk
>         /// <param name="polyline">Polyline to check</param>
>         /// <returns>List of shared polylines</returns>
>         public IEnumerable<Polyline> GetSharedSegments(Polyline polyline)
> ```

I think we should rename this to intersects, and it would be more clear if we kind of inverted it.  can we put this method on the polyline class instead and have it accept a polygon?  basically same functionality but a cleaner API I think.  `IEnumerable<Polyline> polyline.Intersects(polygon)`
___
*[`Elements/test/PolygonTests.cs` line 2060 at r1](https://reviewable.io/reviews/hypar-io/elements/846#-N8TeqAN4q3Ginq1_D4h:-N8TeqAN4q3Ginq1_D4i:bripcjj) ([raw file](https://github.com/hypar-io/elements/blob/fdabb3d3834be90cae34a1a625ad275440d72606/Elements/test/PolygonTests.cs#L2060)):*
> ```Smalltalk
> 
>         [Fact]
>         public void GetSharedSegmentsEmptyWhenDoesNotIntersect()
> ```

move tests to polyline when you move method.

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/846)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Geometry/Polygon.cs` line 2383 at r1](https://reviewable.io/reviews/hypar-io/elements/846#-N8TcQxY3ulq4rP5EnmB:-N8XSYYMCR7gJ6hP8avM:blrk1o5) ([raw file](https://github.com/hypar-io/elements/blob/fdabb3d3834be90cae34a1a625ad275440d72606/Elements/src/Geometry/Polygon.cs#L2383)):*

I think we should rename this to intersects, and it would be more clear if we kind of inverted it.  can we put this method on the polyline class instead and have it accept a polygon?  basically same functionality but a cleaner API I think.  `IEnumerable<Polyline> polyline.Intersects(polygon)`
</blockquote></details>

I applied your suggestion. I also changed method structure into  `bool Intersects(Polygon polygon, out var List<Polyline> sharedSegments)` to mach other `Intersects` methods 
___
*[`Elements/test/PolygonTests.cs` line 2060 at r1](https://reviewable.io/reviews/hypar-io/elements/846#-N8TeqAN4q3Ginq1_D4h:-N8XT4wxC1xqA_9l84XJ:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/fdabb3d3834be90cae34a1a625ad275440d72606/Elements/test/PolygonTests.cs#L2060)):*

move tests to polyline when you move method. 
</blockquote></details>

Done.

wynged (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/846)* status: :shipit: complete! 1 of 1 approvals obtained

PR #849, opened 2022-08-05
Mesh performance

andrewheumann:
BACKGROUND:
- Mesh-Ray intersection used a very brute-force method, looping over all triangles and checking them all for intersection. We were already building an octree for the mesh when adding vertices, but not using it.

DESCRIPTION:
- Uses the mesh's octree to find nearby triangles before trying to compute an intersection
- Exposes a new `Ray.NearbyPoints` method.
- Exposes a public `Elements.Search.PointOctree<T>` method as a thin wrapper around NetOctree's `PointOctree<T>`
- Enable turning the mesh's `BuildOctree` behavior on and off. With it on, mesh construction takes ~2x as long. We leave it on by default since it's still pretty fast, and we want the capabilities enabled by the octree to work "out of the box".

TESTING:
- Added a test for `Ray.NearbyPoints`
- added a new benchmark for `Ray.Intersects(mesh)`:

Before:

|                           Method |    Mean |    Error |   StdDev |
|--------------------------------- |--------:|---------:|---------:|
| 'Intersect 1000 rays with mesh.' | 2.698 s | 0.0365 s | 0.0341 s |

After:

|                           Method |     Mean |   Error |  StdDev |
|--------------------------------- |---------:|--------:|--------:|
| 'Intersect 1000 rays with mesh.' | 148.1 ms | 1.34 ms | 1.25 ms |

- added benchmarks for mesh construction at various sizes with and without the octree:

|           Method | BuildOctree | VertexCount |       Mean |     Error |    StdDev |
|----------------- |------------ |------------ |-----------:|----------:|----------:|
| 'Construct Mesh' |       False |        1000 |   1.156 ms | 0.0120 ms | 0.0112 ms |
| 'Construct Mesh' |       False |        5000 |  12.364 ms | 0.2312 ms | 0.2162 ms |
| 'Construct Mesh' |       False |       10000 |  30.029 ms | 0.6001 ms | 1.3171 ms |
| 'Construct Mesh' |       False |       30000 |  80.432 ms | 1.0678 ms | 0.8916 ms |
| 'Construct Mesh' |        True |        1000 |   3.629 ms | 0.0509 ms | 0.0476 ms |
| 'Construct Mesh' |        True |        5000 |  25.598 ms | 0.3935 ms | 0.3680 ms |
| 'Construct Mesh' |        True |       10000 |  54.057 ms | 0.6537 ms | 0.5459 ms |
| 'Construct Mesh' |        True |       30000 | 194.824 ms | 2.2398 ms | 1.9856 ms |
  
FUTURE WORK:
- This intersects method works best for meshes with more-or-less uniform triangles. If we encounter a lot of meshes with very oblong triangles, it might not be ideal. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/849)


andrewheumann:
So, before you review this guy, it turns out building the octree on construction can be a real performance drag for large meshes. I think we at least need a way to opt out.

sanchez:
Can this be a delayed function or something that can be triggered by a function later down the chain of execution? But if we have it execute within a separate function then it would also need to be serialized too which would massively blow up Model size?

andrewheumann:
@sanchez I don't think we want to serialize the Octree — I think it makes sense to build it at deserialization / mesh construction time. As it stands the `PointOctree` class from the library we're using isn't serializable in any case. I've made building the Octree optional, so you can construct a mesh quickly in one function, then read it into another function and build an octree as needed (either with `BuildOctree`= true, or by forcing it with `ForceBuildOctree()`

sanchez:
Could we instead by default not build it, then the first time a query is executed that requires the octree it is built?

andrewheumann (commented):
You're right, that makes more sense! 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/849#-:-N9-LBayBGLRT3Gtrdj1:b-q21lqy)* status: 0 of 1 approvals obtained (waiting on @cdriesler)

cdriesler (approved):
> Reviewed 4 of 6 files at r1, 6 of 9 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/849)* status: :shipit: complete! 1 of 1 approvals obtained

PR #851, opened 2022-08-07
In memory images for textures.

ikeough:
BACKGROUND:
The API for creating textures for materials in Elements currently takes a relative file path and writes an image file to disk internally. This will not work for wasm scenarios where disk i/o operations are forbidden.

DESCRIPTION:
This pull request adds two new APIs to `Material`, enabling the user to build materials from `byte[]` and from `Image<Rgba32>` objects. Additionally, it extends the current relative path support to more robust `Uri` support, including absolute file paths using `file://`, relative paths, and protocols like `http://` and `https://`. 

TESTING:
- New tests are added that get images from the web to be usedd as textures.

REQUIRED:

### DIFF
--- a/Elements/src/Analysis/AnalysisImage.cs
+++ b/Elements/src/Analysis/AnalysisImage.cs
@@ -4,8 +4,6 @@ using Elements.Geometry;
 using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.PixelFormats;
 
-using System.IO;
-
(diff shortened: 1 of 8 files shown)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/851)* status: :shipit: complete! 1 of 1 approvals obtained

PR #864, opened 2022-08-10
Update build-docs.yml

wynged:
use explicit ref since the default for releases appears to be something else

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/864)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/864)* status: :shipit: complete! 1 of 1 approvals obtained

PR #867, opened 2022-08-17
Add AllowOutsideBoundary and Intersects to Obstacle 

mwojcik-sagiton:
BACKGROUND:
I wanted to combine two behaviors of `Obstacle` subtraction. 
1. When there is a `Boundary` set in `AdaptiveGrid` only the intersection between `Boundary` and `Obstacle` perimeter is subtracted from `Grid`. 
2. When `Boundary` is `null`, whole `Obstacle` perimeter is subtracted from `Grid`

DESCRIPTION:
Together with @wynged we came up with solution of adding `AllowOutsideBoundary` property which is configurable for each `Obstacle`. When new property is set to true, the whole perimeter is subtracted from `Grid`

I also added `Intersects(Polyline polyline)` method to `Obstacle` which I found useful.  It uses `IsLineInDomain` method from `AdaptiveGrid` which I had to make public and static. 

TESTING:
Run unit tests, which are in this PR
  
FUTURE WORK:
None

REQUIRED:

COMMENTS:
Maybe `IsLineInDomain` method be part of `Obstacle` class. It is a question to main author of this class @DmytroMuravskyi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - `Ray.NearbyPoints()`
 - `PointOctree<T>`
 - `Message` class along with helper creation methods.
+- `AdaptiveGrid.Obstacle.AllowOutsideBoundary` property 
+- `AdaptiveGrid.Obstacle.Intersects(Polyline polyline, double tolerance = 1e-05)` method
 
 ### Changed
 
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -1255,7 +1255,7 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="tolerance">How far outside the range numbers are considered inside.
         /// When tolerance positive - range is increased by it, when negative - decreased.</param>
         /// <returns>Low, Inside or High.</returns>
-        private PointOrientation NumberOrientation(
+        private static PointOrientation NumberOrientation(
             double number,
             (double Min, double Max) domain,
             double tolerance)
@@ -1280,7 +1280,7 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="startInside">Is start point of line inside the domain.</param>
         /// <param name="endInside">Is end point of line inside the domain.</param>
         /// <returns>Low, Inside or High.</returns>
-        private bool IsLineInDomain(
+        public static bool IsLineInDomain(
             (Vector3 Start, Vector3 End) line,
             (Vector3 Min, Vector3 Max) domain,
             double xyTolerance, double zTolerance, 
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -282,10 +282,10 @@ namespace Elements.Spatial.AdaptiveGrid
                     var cornersAtElevation = corners.Select(
                         c => c.ProjectAlong(frame.ZAxis, plane)).ToList();
 
-                    AddEdgesOnLine(cornersAtElevation[0], cornersAtElevation[1], intersections);
-                    AddEdgesOnLine(cornersAtElevation[1], cornersAtElevation[2], intersections);
-                    AddEdgesOnLine(cornersAtElevation[2], cornersAtElevation[3], intersections);
-                    AddEdgesOnLine(cornersAtElevation[3], cornersAtElevation[0], intersections);
+                    AddEdgesOnLine(cornersAtElevation[0], cornersAtElevation[1], intersections, obstacle.AllowOutsideBoudary);
+                    AddEdgesOnLine(cornersAtElevation[1], cornersAtElevation[2], intersections, obstacle.AllowOutsideBoudary);
+                    AddEdgesOnLine(cornersAtElevation[2], cornersAtElevation[3], intersections, obstacle.AllowOutsideBoudary);
+                    AddEdgesOnLine(cornersAtElevation[3], cornersAtElevation[0], intersections, obstacle.AllowOutsideBoudary);
 
                     foreach (var item in group)
                     {
@@ -949,9 +949,9 @@ namespace Elements.Spatial.AdaptiveGrid
             return addedEdges;
         }
 
-        private void AddEdgesOnLine(Vector3 start, Vector3 end, IEnumerable<Vector3> candidates)
+        private void AddEdgesOnLine(Vector3 start, Vector3 end, IEnumerable<Vector3> candidates, bool allowEdgesOutsideBoudnary)
         {
-            if (Boundaries != null)
+            if (Boundaries != null && !allowEdgesOutsideBoudnary)
             {
                 var boundary2d = new Polygon(Boundaries.Vertices.Select(v => new Vector3(v.X, v.Y)).ToList());
                 var inside = new Line(new Vector3(start.X, start.Y), new Vector3(end.X, end.Y)).Trim(boundary2d, out var _);
--- a/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs
@@ -1,10 +1,7 @@
-﻿using Elements;
-using Elements.Geometry;
-using Elements.Spatial.AdaptiveGrid;
+﻿using Elements.Geometry;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Elements.Spatial.AdaptiveGrid
 {
@@ -22,8 +19,9 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="column">Column to avoid.</param>
         /// <param name="offset">Extra space around obstacle bounding box.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <returns>New obstacle object.</returns>
-        public static Obstacle FromColumn(Column column, double offset = 0, bool perimeter = false)
+        public static Obstacle FromColumn(Column column, double offset = 0, bool perimeter = false, bool allowOutsideBoundary = false)
         {
             var p = column.Profile.Perimeter.TransformedPolygon(
                 new Transform(column.Location));
@@ -31,7 +29,7 @@ namespace Elements.Spatial.AdaptiveGrid
             points.AddRange(p.Vertices);
             points.AddRange(p.Vertices.Select(
                 v => new Vector3(v.X, v.Y, v.Z + column.Height)));
-            return new Obstacle(points, offset, perimeter, null);
+            return new Obstacle(points, offset, perimeter, allowOutsideBoundary, null);
         }
 
         /// <summary>
@@ -40,8 +38,9 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="wall">Wall to avoid.</param>
         /// <param name="offset">Extra space around obstacle bounding box.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <returns>New obstacle object.</returns>
-        public static Obstacle FromWall(StandardWall wall, double offset = 0, bool perimeter = false)
+        public static Obstacle FromWall(StandardWall wall, double offset = 0, bool perimeter = false, bool allowOutsideBoundary = false)
         {
             var ortho = wall.CenterLine.Direction().Cross(Vector3.ZAxis);
             List<Vector3> points = new List<Vector3>();
@@ -52,7 +51,7 @@ namespace Elements.Spatial.AdaptiveGrid
             points.AddRange(points.Select(v => new Vector3(v.X, v.Y, v.Z + wall.Height)).ToArray());
             var transfrom = new Transform(Vector3.Origin,
                 wall.CenterLine.Direction(), ortho, Vector3.ZAxis);
-            return new Obstacle(points, offset, perimeter, transfrom);
+            return new Obstacle(points, offset, perimeter, allowOutsideBoundary, transfrom);
         }
 
         /// <summary>
@@ -61,10 +60,11 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="box">Bounding box to avoid.</param>
         /// <param name="offset">Extra space around obstacle bounding box.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <returns>New obstacle object.</returns>
-        public static Obstacle FromBBox(BBox3 box, double offset = 0, bool perimeter = false)
+        public static Obstacle FromBBox(BBox3 box, double offset = 0, bool perimeter = false, bool allowOutsideBoundary = false)
         {
-            return new Obstacle(box.Corners(), offset, perimeter, null);
+            return new Obstacle(box.Corners(), offset, perimeter, allowOutsideBoundary, null);
         }
 
         /// <summary>
@@ -74,13 +74,14 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="height">Height of the obstacle.</param>
         /// <param name="offset">Extra space around obstacle bounding box.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <returns>New obstacle object.</returns>
-        public static Obstacle From2DPolygon(Polygon polyon, double height, double offset = 0, bool perimeter = false)
+        public static Obstacle From2DPolygon(Polygon polyon, double height, double offset = 0, bool perimeter = false, bool allowOutsideBoundary = false)
         {
             List<Vector3> points = new List<Vector3>();
             points.AddRange(polyon.Vertices.Select(p => new Vector3(p.X, p.Y)));
             points.AddRange(polyon.Vertices.Select(p => new Vector3(p.X, p.Y, height)));
-            return new Obstacle(points, offset, perimeter, null);
+            return new Obstacle(points, offset, perimeter, allowOutsideBoundary, null);
         }
 
         /// <summary>
@@ -89,8 +90,9 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="line">Line to avoid.</param>
         /// <param name="offset">Extra space around obstacle bounding box. Should be larger than 0.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <returns>New obstacle object.</returns>
-        public static Obstacle FromLine(Line line, double offset = 0.1, bool perimeter = false)
+        public static Obstacle FromLine(Line line, double offset = 0.1, bool perimeter = false, bool allowOutsideBoundary = false)
         {
             if (offset < Vector3.EPSILON)
             {
@@ -110,7 +112,7 @@ namespace Elements.Spatial.AdaptiveGrid
                 frame = new Transform(Vector3.Origin, forward, rigth, up);
             }
 
-            return new Obstacle(points, offset, perimeter, frame);
+            return new Obstacle(points, offset, perimeter,allowOutsideBoundary, frame);
         }
 
         /// <summary>
@@ -119,12 +121,14 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <param name="points"></param>
         /// <param name="offset">Extra space around obstacle bounding box.</param>
         /// <param name="perimeter">Should edges be created around obstacle.</param>
+        /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
         /// <param name="transformation">Transformation of the obstacle.</param>
-        public Obstacle(List<Vector3> points, double offset, bool perimeter, Transform transformation)
+        public Obstacle(List<Vector3> points, double offset, bool perimeter, bool allowOutsideBoundary, Transform transformation)
         {
             Points = points;
             Offset = offset;
             Perimeter = perimeter;
+            AllowOutsideBoudary = allowOutsideBoundary;
             Transform = transformation;
         }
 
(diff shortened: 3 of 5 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton)

___
*[`Elements/test/AdaptiveGridTests.cs` line 399 at r1](https://reviewable.io/reviews/hypar-io/elements/867#-N9iRSaV9TGBEFR79JyE:-N9iRSaV9TGBEFR79JyF:bwonnva) ([raw file](https://github.com/hypar-io/elements/blob/3ccf49382cc09e5e3ea4c14e897dc99337fcf979/Elements/test/AdaptiveGridTests.cs#L399)):*
> ```Smalltalk
>             var profile = Polygon.Rectangle(1, 1);
> 
>             //Column outside of boundary and does not intersect with any edge or vertex
> ```

I think we need one more pair of tests for Column fully inside boundary

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/test/AdaptiveGridTests.cs` line 399 at r1](https://reviewable.io/reviews/hypar-io/elements/867#-N9iRSaV9TGBEFR79JyE:-N9j_Bn01p0hTbueOuYK:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/3ccf49382cc09e5e3ea4c14e897dc99337fcf979/Elements/test/AdaptiveGridTests.cs#L399)):*

I think we need one more pair of tests for Column fully inside boundary
</blockquote></details>

Done.

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton and @wynged)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 63 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k89l86_9kDWj26Obr:-N9k89l86_9kDWj26Obs:bv4674b) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L63)):*
> ```Smalltalk
>         /// <param name="offset">Extra space around obstacle bounding box.</param>
>         /// <param name="perimeter">Should edges be created around obstacle.</param>
>         /// <param name="allowOutsideBoundary">Should edges be created when obstacle is outside of <see cref="AdaptiveGrid.Boundaries"/></param>
> ```

I'm curious about all repetitive parameter stacking. And there is also Transform parameter that is not even here because some functions set it during construction automatically but there is still value in modifying it. Maybe perimeter and allowOutsideBoundary can also be just property instead of parameters?
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9k6NXL5jyzMNvoWxq5:b-7u6owh) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*
> ```Smalltalk
>         /// <param name="tolerance">Tolerance for checks</param>
>         /// <returns>Result of check</returns>
>         public bool Intersects(Polyline polyline, double tolerance = 1e-05)
> ```

IsLineInDomain does not guarantee that you intersect or inside. It guarantees that if it returned false - line is not inside and not intersects. This allows to skip expensive but precise operations like Intersects. For example, for rotated polygon 0;0 -> 5;5 -> 0; 10 -> -5; 5 and line -4; 6 -> -1; 9 the function would return true and both startInside and endInside would be true but this is only about numerical domain - line itself is not inside or intersects. So I'm hesitant about this function.



_Code quote:_
```Smalltalk
IsLineInDomain
```

wynged (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 63 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k89l86_9kDWj26Obr:-N9l3Oy_8eXHNLx143n4:b-j65iry) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L63)):*

I'm curious about all repetitive parameter stacking. And there is also Transform parameter that is not even here because some functions set it during construction automatically but there is still value in modifying it. Maybe perimeter and allowOutsideBoundary can also be just property instead of parameters?
</blockquote></details>

These parameters are all optional, so the user doesn't have to set them if they would rather set them on the object directly....  This feels like it makes it more discoverable.  I think I would move to a "configuration" object that consolidates these options rather than getting rid of them, but is a config option for only 2 params worth it?  what looks cleanest?
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9l2vrR9OmqjHVDEDAf:bl553m1) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*

IsLineInDomain does not guarantee that you intersect or inside. It guarantees that if it returned false - line is not inside and not intersects. This allows to skip expensive but precise operations like Intersects. For example, for rotated polygon 0;0 -> 5;5 -> 0; 10 -> -5; 5 and line -4; 6 -> -1; 9 the function would return true and both startInside and endInside would be true but this is only about numerical domain - line itself is not inside or intersects. So I'm hesitant about this function.
</blockquote></details>

@mwojcik-sagiton  can we improve this to use `IsLineInDomain` as a quick filter, but then do a real `Intersect` with any lines that pass the first filter.  Then I think we can be more confident in the result, right @dmytromuravskyi  ?

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 63 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k89l86_9kDWj26Obr:-N9kBpvD7EyyXSfUnLq3:b-aw9m6o) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L63)):*

These parameters are all optional, so the user doesn't have to set them if they would rather set them on the object directly....  This feels like it makes it more discoverable.  I think I would move to a "configuration" object that consolidates these options rather than getting rid of them, but is a config option for only 2 params worth it?  what looks cleanest?
</blockquote></details>

I think it depends on parameter. In my use case I always set `Perimeter` value in constructor, because when I create na `Obstacle` object I know I would need its `Perimeter` to be added to the grid.

Regarding `AllowOutsideBoudnary` case is a bit different, because I need to create `Obstacle` object first and then I use `Intersects` method to check if specific obstacle intersects with Boundary of roof section

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9l5L_-0wJ9BNuTnwiO:bsf8rss) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*

@mwojcik-sagiton  can we improve this to use `IsLineInDomain` as a quick filter, but then do a real `Intersect` with any lines that pass the first filter.  Then I think we can be more confident in the result, right @dmytromuravskyi  ?
</blockquote></details>

Probably we can add method `Intersects(Line line, out Vector result)` I imagine it would be useful for `AdaptiveGrid` to get intersections between edges and Obstacle. 

I checked how AdaptiveGrid will subtract Obstacle that is rotated and it adds orthogonal box of it 

I checked rotated polygon and orthogonal rectangle with applied Transform

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton and @wynged)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 63 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k89l86_9kDWj26Obr:-N9lSL5sEQcgGTxi3o_B:b-805eag) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L63)):*
<details><summary><i>Previously, mwojcik-sagiton wrote…</i></summary><blockquote>

I think it depends on parameter. In my use case I always set `Perimeter` value in constructor, because when I create na `Obstacle` object I know I would need its `Perimeter` to be added to the grid.

Regarding `AllowOutsideBoudnary` case is a bit different, because I need to create `Obstacle` object first and then I use `Intersects` method to check if specific obstacle intersects with Boundary of roof section 
</blockquote></details>

I'm fine with keeping them both optional
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9lSjinCMSIwh0S5mgC:b-mrt69q) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*
<details><summary><i>Previously, mwojcik-sagiton wrote…</i></summary><blockquote>

Probably we can add method `Intersects(Line line, out Vector result)` I imagine it would be useful for `AdaptiveGrid` to get intersections between edges and Obstacle. 

I checked how AdaptiveGrid will subtract Obstacle that is rotated and it adds orthogonal box of it 

I checked rotated polygon and orthogonal rectangle with applied Transform  
</blockquote></details>

Yes, it works in the grid, because it takes into account Transform property and if you set it to 45 degree rotate than obstacle is rotated on the grid. This function doesn't take transformation into account. 
Just try this case it IntersectsTest:
            yield return new object[] { new Polyline
            (
                new Vector3(5, -10),
                new Vector3(20, 5),
                new Vector3(5, 20),
                new Vector3(-10, 5)
            ), false };
it fails, although polyline is around obstacle and never intersects it

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9lxYxX-qkjUaZxgBFe:bbdpm6l) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*

Yes, it works in the grid, because it takes into account Transform property and if you set it to 45 degree rotate than obstacle is rotated on the grid. This function doesn't take transformation into account. 
Just try this case it IntersectsTest:
            yield return new object[] { new Polyline
            (
                new Vector3(5, -10),
                new Vector3(20, 5),
                new Vector3(5, 20),
                new Vector3(-10, 5)
            ), false };
it fails, although polyline is around obstacle and never intersects it

</blockquote></details>

So I think the plan is to keep this method with polyline, and add an `out Vector3[] hits` argument.
Keep the step you have below with the IsLineInDomain checks, but then we need to add another step.  something like this I think.

``` C#
var allHits = new List<Vector3>();

foreach(var line in polyline.Segments()) 
{
    if(!AdaptiveGrid.IsLineInDomain((x.Start, x.End), domain, tolerance, tolerance, out var _, out var _))) 
    {
         continue;
    }
    foreach(var edge in Edges) {
        if(AdaptiveGrid.IsLineInDomain((edge.Start, edge.End), (line.Start, line.End), tolerance, tolerance, out var _, out var_)) 
        {
             if(line.Intersects(edge, out var hit)) 
             {
                 allHits.Add(hit);
             }
        }
}
hits = allHits.ToArray();
return allHits.Any();
```

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 168 at r2](https://reviewable.io/reviews/hypar-io/elements/867#-N9k6NXL5jyzMNvoWxq4:-N9ppfPw8Ppr3coBUZF_:b-pi0jl3) ([raw file](https://github.com/hypar-io/elements/blob/7a56fe238642e27cd69daed84c1c2eef28557409/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L168)):*

So I think the plan is to keep this method with polyline, and add an `out Vector3[] hits` argument.
Keep the step you have below with the IsLineInDomain checks, but then we need to add another step.  something like this I think.

``` C#
var allHits = new List<Vector3>();

foreach(var line in polyline.Segments()) 
{
    if(!AdaptiveGrid.IsLineInDomain((x.Start, x.End), domain, tolerance, tolerance, out var _, out var _))) 
    {
         continue;
    }
    foreach(var edge in Edges) {
        if(AdaptiveGrid.IsLineInDomain((edge.Start, edge.End), (line.Start, line.End), tolerance, tolerance, out var _, out var_)) 
        {
             if(line.Intersects(edge, out var hit)) 
             {
                 allHits.Add(hit);
             }
        }
}
hits = allHits.ToArray();
return allHits.Any();
```
</blockquote></details>

@wynged it won't work because I want to run `Obstacle.Intersect` without `AdaptiveGrid`

DmytroMuravskyi (commented):
> Reviewed 1 of 1 files at r2, 1 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 315 at r4](https://reviewable.io/reviews/hypar-io/elements/867#-NA3n5Kk5i-ITeYfj3OW:-NA3n5Kk5i-ITeYfj3OX:b-kl6j0n) ([raw file](https://github.com/hypar-io/elements/blob/e261ffe79c46166c81e04ad1af671a0719f43883/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L315)):*
> ```Smalltalk
>                 line.Intersects(x, out var result);
>                 return result;
>             }).ToList();
> ```

That's the valid way to solve this issue. It reminds me of Line.Intersects(BBox). I guess it would be great to incorporate this somehow into general Elements code. We basically do the same but the box has transformation and we also check if parameter is inside instead of only intersections. 
CreatePolygonFromPoints looks suspicious. It's fine for horizontalPlanes and only when where are only 4 points on minZ plane but will fail for  more complex cases like column where bottom perimeter consist of many points. What if instead of it we just do it the same way as in AdaptiveGrid.SubtractObstacle: create box from transformed points BBox3 localBox = new BBox3(localPoints).Offset(obstacle.Offset); and then use box corner points to create planes. This will make code much smaller and more consistand with other places.



_Code quote:_
```Smalltalk
f (!IsBetweenPlanesOrOnAnyOfThem(_horizontalPlanes, point, tolerance))
```

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi and @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 80 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9cYca8Zmcfk4DPaSP:-NA9cYca8Zmcfk4DPaSQ:b-87vcah) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L80)):*
> ```Smalltalk
>             var filteredVerices = new List<Vector3>();
> 
>             foreach(var vertex in minZVertices)
> ```

I don't think we need any filtering here. 
We can make this much cleaner overall:
var polygon = newPolygon(new Vertex3(box.Min.X, box.Min.Y, box.Min.Z, new Vertex3(box.Min.X, box.Max.Y, box.Min.Z, new Vertex3(box.Max.X, box.Max.Y, box.Min.Z, new Vertex3(box.Max.X, box.Min.Y, box.Min.Z);
var height = box.Max.Z - box.Min.Z;
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 112 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9dDpF2yXyKTFnv2LN:-NA9dDpF2yXyKTFnv2LO:bbd1rcg) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L112)):*
> ```Smalltalk
>                 throw new ArgumentException("Offset should be larger then zero.");
>             }
> ```

Please do custom transform only when `if (!forward.IsParallelTo(Vector3.ZAxis))` is true. Otherwise I want grid transformation to be used so obstacle is properly oriented.
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 154 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9dnHw2C-9fL89WPz1:-NA9dnHx5l82kTasNkdw:b-fe6vpz) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L154)):*
> ```Smalltalk
>         public List<Vector3> Points => Boundary?
>             .Vertices
>             .SelectMany(x => new List<Vector3> { x, x + new Vector3(0, 0, Height) })
> ```

This wont work for obstacle from line where Height is not always in Z direction. Might want to use some extrusion axis internally.
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 275 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9fuWmAWekuhhV5tgL:-NA9fuWmAWekuhhV5tgM:b9vukpb) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L275)):*
> ```Smalltalk
>             }
> 
>             var distance = plane.SignedDistanceTo(intersection);
> ```

Since intersection is on the plane, I can't understand when distance is not 0?
___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 277 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9g8kQ-45YPDW3r0WF:-NA9g8kQ-45YPDW3r0WG:b-yexeaq) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L277)):*
> ```Smalltalk
>             var distance = plane.SignedDistanceTo(intersection);
> 
>             return plane.Normal.IsParallelTo(line.Direction()) && (distance.ApproximatelyEquals(0, tolerance) || distance < 0);
> ```

Why is it has to be parallel?

DmytroMuravskyi (commented):
> Reviewed 1 of 4 files at r4, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @mwojcik-sagiton)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 112 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9dDpF2yXyKTFnv2LN:-NAAFK8C6PD2809Fl6Wn:b9w72kz) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L112)):*

Please do custom transform only when `if (!forward.IsParallelTo(Vector3.ZAxis))` is true. Otherwise I want grid transformation to be used so obstacle is properly oriented.
</blockquote></details>

It's fine as it is as long as it's working

mwojcik-sagiton (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Spatial/AdaptiveGrid/Obstacle.cs` line 112 at r5](https://reviewable.io/reviews/hypar-io/elements/867#-NA9dDpF2yXyKTFnv2LN:-NAE2Eft9dGVDS-Kymus:b-fvpq1e) ([raw file](https://github.com/hypar-io/elements/blob/46552a1adfb7266de607deb263b841eba592f142/Elements/src/Spatial/AdaptiveGrid/Obstacle.cs#L112)):*
<details><summary><i>Previously, mwojcik-sagiton wrote…</i></summary><blockquote>

I think I need to check how it generates Points for vertical polygons/lines. I'll add adequate unit tests for that 
</blockquote></details>

I changed logic to support other polygons that horizontal, inside `UpdatePolygons` I calculate `Transformation` of `Boundary`. Then I invert this transformation to calculate offset polygon and finally I can transform offset polygon back. 

I had to fix some issues inside `FromLine` function and now Obstacles from created from lines works as expected. In unit tests I checked also angled lines.

DmytroMuravskyi (commented):
Looks good

> Reviewed 1 of 2 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867#-:-NAdbnkO99L7lsoXH5lS:bsrfkwe)* status: 1 change requests, 0 of 1 approvals obtained

wynged (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/867#-:-NAdk8EoAvv9KJjBD4XW:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #870, opened 2022-08-22
Stj

ikeough:
BACKGROUND:

DESCRIPTION:

TESTING:
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/870)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mattc)

___
*[`Elements/src/Model.cs` line 340 at r8](https://reviewable.io/reviews/hypar-io/elements/870#-NA4IQaY8fn9i7kQqX2c:-NA4IQaY8fn9i7kQqX2d:b-2yrx29) ([raw file](https://github.com/hypar-io/elements/blob/5a6726753587d84f47b46994a4f45c8f35f726bf/Elements/src/Model.cs#L340)):*
> ```Smalltalk
>                             // Qualify element as a geometric element by seeing 
>                             // whether it has a representation.
>                             if (element.Value.TryGetProperty("Representation", out var repProperty))
> ```

let's also verify that `repProperty` does not have a null value. some `GeometricElement`s (`MeshElement`, for instance) may have a null `Representation`. Here's an easy failing test:
```csharp
var model = new Model();
var meshSphere = Mesh.Sphere(5, 10);
var meshElement = new MeshElement(meshSphere);
model.AddElement(meshElement);
var json2 = model.ToJson();
var newModel2 = Model.GeometricElementModelFromJson(json2);
```
___
*[`Elements/src/Model.cs` line 355 at r8](https://reviewable.io/reviews/hypar-io/elements/870#-NA4Io_HAhbe-iPJFTW1:-NA4Io_HAhbe-iPJFTW2:b-xwkb0f) ([raw file](https://github.com/hypar-io/elements/blob/5a6726753587d84f47b46994a4f45c8f35f726bf/Elements/src/Model.cs#L355)):*
> ```Smalltalk
>                                     {
>                                         case "Elements.Geometry.Solids.Extrude":
>                                             var profile = (Profile)resolver.ResolveReference(solidOp.GetProperty("Profile").GetString());
> ```

Our current deserialization logic handles inline references as well as id-based references for other elements, like Profiles. This probably doesn't occur much in normal function execution, where the Id replacement happens at `AddElements` time, but other pathways might use it (I ran into it w/ typescript functions where I was hand-generating model json, but I can imagine it popping up in revit, content workflows, etc). Your call — perhaps we want to be strict that the "Profile" property can always be assumed to be a string — but we could make this a bit more tolerant with something like this:
```csharp
       private static T ResolveOrReturn<T>(JsonElement jsonElement, string propertyName, ElementReferenceResolver resolver) where T : Element
        {
            if (jsonElement.TryGetProperty(propertyName, out var property))
            {
                if (property.ValueKind == JsonValueKind.String)
                {
                    return resolver.ResolveReference(property.GetString()) as T;
                }
                else if (property.ValueKind == JsonValueKind.Object)
                {
                    return JsonSerializer.Deserialize<T>(property);
                }
            }
            return null;
        }
```
so the call could just be `var profile = ResolveOrReturn<Profile>(solidOp, "Profile", resolver);`.  I could also see this being folded into `ResolveReference` by making it take a `JsonElement` instead of a string.

andrewheumann (commented):
Haven't done a full review yet, just marking some stuff I noticed over the weekend. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/870#-:-NA4Knbi4l2KPou1tENB:bpvir0c)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mattc)

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/870)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mattc)

___
*[`Elements/src/Serialization/JSON/ElementConverter.cs` line 112 at r9](https://reviewable.io/reviews/hypar-io/elements/870#-NAGrbPX9UduzFEU_zCS:-NAGrbPX9UduzFEU_zCT:b-kd9on7) ([raw file](https://github.com/hypar-io/elements/blob/a141a38f664f42c4576b415bf19fcae56a5b84e1/Elements/src/Serialization/JSON/ElementConverter.cs#L112)):*
> ```Smalltalk
>                             int end = discriminator.IndexOf(">", start);
>                             string elementType = discriminator.Substring(start, end - start);
>                             if (!resolver.TypeCache.TryGetValue(elementType, out var genericType))
> ```

It's quite common to come across a type you don't have a class for. We need to be able to fall back to `Element` or `GeometricElement` if appropriate.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/870)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough and @mattc)

___
*[`Elements/src/Serialization/JSON/ElementConverter.cs` line 112 at r9](https://reviewable.io/reviews/hypar-io/elements/870#-NAGrbPX9UduzFEU_zCS:-NAGt6mbBYNcZrL9EhzP:b-r0qymo) ([raw file](https://github.com/hypar-io/elements/blob/a141a38f664f42c4576b415bf19fcae56a5b84e1/Elements/src/Serialization/JSON/ElementConverter.cs#L112)):*

It's quite common to come across a type you don't have a class for. We need to be able to fall back to `Element` or `GeometricElement` if appropriate.
</blockquote></details>

(actually, since this is just the proxy code, it's probably safe to just fallback to "Element.")

PR #872, opened 2022-08-23
Playground

ikeough:
BACKGROUND:
I was on summer vacation and wanted to update the playground to have better lighting, shadows, and better error handling and messaging about code and performance timings.

DESCRIPTION:
- Adds timers to give us insight into how long compilation, execution, and gltf creation take.
- Adds the concept of inputs to demonstrate re-execution without recompile of the code, and accessing input values from code. This work probably won't go any further because we're going to wire all of this into Hypar and use geometry inputs from there.
- Add HDRI lighting and shadows to the scene.
- Removes the auto-running of the code.
- Saves your code to local storage so that you don't have to re-write it every time you rebuild.

TESTING:
```
cd Elements.Playground
dotnet run
```
 
REQUIRED:

https://user-images.githubusercontent.com/1139788/186271716-59b4d49a-45ba-4b54-a05e-40771436af73.mov

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/872)


ikeough:
@Gytaco also noted that the CDN that served the ACE editor went kaput. This fixes that as well, but we could also fix it in a smaller PR.

ikeough:
@andrewheumann I'm opening this up for discussion. We should decide if we want to make this into a blazor component, and if so, where the boundary should be. We could just keep the compiler and the bridge API and provide the code editor and calls to the API from Hypar.

PR #874, opened 2022-08-27
Elements web assembly API

ikeough:
BACKGROUND:
We've had the Elements Playground for some time, but we always knew that we wanted to move the live code editing environment into Hypar. That is, we didn't want to maintain a separate code playground. The `Elements.Playground` project was previously the 

DESCRIPTION:
This PR combines work being done in `Elements.Wasm` and `Elements.Playground` into one minimal implementation of Elements as a web assembly API. It includes a small demonstration app and instructions for how to use the Elements web assembly API in your own application.

TESTING:
`dotnet watch run`
  
FUTURE WORK:
As noted in the `README`, we'd like to remove the requirement to create a Blazor component in the future. It adds an empty div to the DOM.

REQUIRED:


ikeough:
Requires #875.

PR #876, opened 2022-08-29
System.Text.Json fixes

andrewheumann:
BACKGROUND:
- There were some issues using the STJ version of Elements for client-side geometry generation
- There were also some issues building / code-generating hypar functions that worked correctly with STJ. 

DESCRIPTION:
- Fixes `enum` code generation — STJ's `JsonStringEnumConverter` ignores `[EnumMember]` attributes, which means JSON from hypar containing enums was frequently failing to deserialize and not serializing with the correct member names. the `JsonStringEnumConverter` is replaced with `Macross.Json.Extensions`'s `JsonStringEnumMemberConverter`.
- deserializing `ProxyElement`s of types that were not loaded was throwing an exception.
- Adds a generic test that reads some example models (which were previously failing). Over time we can add more real models here to test.

TESTING:
- I tested by using this branch in tandem with [elements/wasm-explore-fixes](https://github.com/hypar-io/Elements/tree/wasm-explore-fixes) and [explore/wasm-playground](https://github.com/hypar-io/explore/tree/wasm-playground) and testing client-side geometry generation on explore
- I also tested this branch in tandem with [hypar/stj](https://github.com/hypar-io/hypar/tree/stj). Functions aren't working all the way yet, since we're basically forced to upgrade to .net 6.0 — but most codegen and local building/running is working.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/876)


ikeough (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/876)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Serialization/JSON/ElementConverter.cs` line 115 at r1](https://reviewable.io/reviews/hypar-io/elements/876#-NAdpMW6Cz7ae4sAV7bj:-NAdpMW6Cz7ae4sAV7bk:b-j7qk9g) ([raw file](https://github.com/hypar-io/elements/blob/6256fbef984c5a151f2f489c2716c5c56d2a2fa9/Elements/src/Serialization/JSON/ElementConverter.cs#L115)):*
> ```Smalltalk
>                             {
>                                 // This may occur if the object is a proxy of a type we don't have the class for. Fallback to `Element`.
>                                 genericType = typeof(Element);
> ```

Very nice.
___
*[`Elements.CodeGeneration/src/HyparFilters.cs` line 56 at r1](https://reviewable.io/reviews/hypar-io/elements/876#-NAdpWIF0HJJxNQDxPIn:-NAdpWIF0HJJxNQDxPIo:btse4tr) ([raw file](https://github.com/hypar-io/elements/blob/6256fbef984c5a151f2f489c2716c5c56d2a2fa9/Elements.CodeGeneration/src/HyparFilters.cs#L56)):*
> ```Smalltalk
>         /// Strip the nullable annotation (?) from a type.
>         /// </summary>
>         public static string Removenull(string input)
> ```

Do we use this in this PR?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/876)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Serialization/JSON/ElementConverter.cs` line 115 at r1](https://reviewable.io/reviews/hypar-io/elements/876#-NAdpMW6Cz7ae4sAV7bj:-NAduZhH6u0wa9HSwtlo:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/6256fbef984c5a151f2f489c2716c5c56d2a2fa9/Elements/src/Serialization/JSON/ElementConverter.cs#L115)):*

Very nice.
</blockquote></details>

Done.
___
*[`Elements.CodeGeneration/src/HyparFilters.cs` line 56 at r1](https://reviewable.io/reviews/hypar-io/elements/876#-NAdpWIF0HJJxNQDxPIn:-NAduafv0HL5gwd0Hb-P:b-kfd158) ([raw file](https://github.com/hypar-io/elements/blob/6256fbef984c5a151f2f489c2716c5c56d2a2fa9/Elements.CodeGeneration/src/HyparFilters.cs#L56)):*

Do we use this in this PR?
</blockquote></details>

Ah, good catch — I was using it, but now it's no longer needed with the third-party Enum converter.

ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/876)* status: :shipit: complete! 1 of 1 approvals obtained

PR #878, opened 2022-08-29
fix splitting of clockwise profiles

### LINKED ISSUE #877: Profile.Split does not split as expected
Reported by @anthoniekramer
```csharp
var polygon = new Polygon((0,0,0), (0,10,0),(10,10,0), (10,0,0));
var polylineOffsetInside = new Polygon((3,3), (3,7), (7,7), (7,3));
var extendedLines = polylineOffsetInside.Segments().Select(s => s.ExtendTo(polygon, true).ToPolyline(1)).ToList();
var splitResults = Profile.Split(new[] {new Profile(polygon)}, extendedLines);
```
We should be getting 9 distinct regions here, instead we get a single one with a void.

andrewheumann:
BACKGROUND:
- Fixes #877 

DESCRIPTION:
- Forces split profiles to be counter-clockwise wound, since the code measures rotation angle relative to positive Z. 

TESTING:
- I added a new test from user code, and verified it produced the expected results. 
  
FUTURE WORK:
- We might want to figure out an average `Normal` for the profiles and feed that as the normal to the polygonizer, so we can respect the input winding. But in the case that some profiles are CW, and some are CCW, it's not clear what the algorithm should do. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/878)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - `AdaptiveGrid.SubtractObstacle` worked incorrectly in `AdaptiveGrid.Boundary` had elevation.
 - #805
 - `Polyline.Intersects(Polygon polygon, out List<Polyline> sharedSegments)` bug when polyline start/end is on polygon perimeter 
+- `Profile.Split` would fail if the perimeter was clockwise-wound.
 
 ## 1.1.0
 
--- a/Elements/src/Geometry/Profile.cs
+++ b/Elements/src/Geometry/Profile.cs
@@ -488,14 +488,16 @@ namespace Elements.Geometry
             List<Profile> resultProfiles = new List<Profile>();
             foreach (var inputProfile in profiles)
             {
+                var polygons = new List<Polygon>();
                 if (inputProfile.Perimeter.IsClockWise())
                 {
-                    inputProfile.Perimeter = inputProfile.Perimeter.Reversed();
+                    polygons.Add(inputProfile.Perimeter.Reversed());
                 }
-                inputProfile.OrientVoids();
-                var polygons = new List<Polygon>() {
-                    inputProfile.Perimeter
-                };
+                else
+                {
+                    polygons.Add(inputProfile.Perimeter);
+                }
+
                 var splitters = new List<Polyline>(splitLines);
                 if (inputProfile.Voids != null)
                 {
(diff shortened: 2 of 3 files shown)


katehryhorenko (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/878)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Profile.cs` line 493 at r1](https://reviewable.io/reviews/hypar-io/elements/878#-NC5kzqs4GnRTlRVXS9B:-NC5kzqs4GnRTlRVXS9C:b-ywc2a7) ([raw file](https://github.com/hypar-io/elements/blob/da4d7e0477088fddaddfb3173c5d9cef55264386/Elements/src/Geometry/Profile.cs#L493)):*
> ```Smalltalk
>                 if (inputProfile.Perimeter.IsClockWise())
>                 {
>                     inputProfile.Perimeter = inputProfile.Perimeter.Reversed();
> ```

Are we sure we want to change the original profiles? Should not we work with copy? 



_Code snippet:_
```Smalltalk
var polygons = new List<Polygon>()
 if ( inputProfile.Perimeter.IsClockWise())
{
    polygons .Add( inputProfile.Perimeter.Reversed());
}
else
{
    polygons .Add( inputProfile.Perimeter);
}
```

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/878)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @katehryhorenko)

___
*[`Elements/src/Geometry/Profile.cs` line 493 at r1](https://reviewable.io/reviews/hypar-io/elements/878#-NC5kzqs4GnRTlRVXS9B:-NCaixyb3Ffz5jtrFr3y:bk7ik76) ([raw file](https://github.com/hypar-io/elements/blob/da4d7e0477088fddaddfb3173c5d9cef55264386/Elements/src/Geometry/Profile.cs#L493)):*

Are we sure we want to change the original profiles? Should not we work with copy? 
</blockquote></details>

Good call. I've removed the `OrientVoids` call as well — it also mutates the profile (and wasn't needed here).

katehryhorenko (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/878#-:-ND4l9eZ40uGPPyWf4UY:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #879, opened 2022-08-31
always make the default material available

wynged:
BACKGROUND:
- While writing gltf for a model the default material was not available during mesh material assignment, which caused errors.

DESCRIPTION:
- Make it so that the default material is always available to the gltf mesh writers.

TESTING:
- tests all still pass, material is available.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/879)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - `Ray.NearbyPoints()`
 - `PointOctree<T>`
 - `Message` class along with helper creation methods.
-- `AdaptiveGrid.Obstacle.AllowOutsideBoundary` property 
+- `AdaptiveGrid.Obstacle.AllowOutsideBoundary` property
 - `AdaptiveGrid.Obstacle.Intersects(Polyline polyline, double tolerance = 1e-05)` method
 - `AdaptiveGrid.Obstacle.Intersects(Line line, double tolerance = 1e-05)` method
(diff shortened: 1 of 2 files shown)


andrewheumann (commented):
Approving, but not clear to me how this situation can occur — if an element has the default material, shouldn't it already be available? Fine with this change, but let's just make sure it's not papering over something else. Perhaps a case where we're keeping a "null" material all the way up to GLTF time (which is maybe fine, and maybe GLTF time is a reasonable moment to resolve that w/ a default material).

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/879#-:-NBSN2m09i3bS3kGJJ2j:bz6mvv3)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/879#-:-NBSNd5f9niisvHy2338:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

wynged:
Nope!  adn 

> Should this get merged @wynged?

Nope!  I thought I was fixing an issue but there was a different root cause.  it wouldn't hurt but we might as well leave it as is, especially merging this would paper over other bugs.

PR #897, opened 2022-10-07
Make vertex console writing optional

wynged:
BACKGROUND:
- When using elements library and in debug mode geometry generation can be very slow because we get a load of logging about vertex management.  
image

DESCRIPTION:
- Make the logging of vertex data an explicit flag that needs to be set, intended for use while debugging.

TESTING:
- When using elements in a test for example this should change the default behavior so that it doesn't show all of these logs.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/897)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@
 - Moved `BranchSide`, `RoutingVertex`, `RoutingConfiguration`, `RoutingHintLine`, `TreeOrder` from `AdaptiveGraphRouting` to their own files.
 - `RoutingVertex` - removed `Guides`.
 - `AdaptiveGraphRouting.BuildSpanningTree` functions are simplified. Also, they use only single `tailPoint` now.
-- `AdaptiveGraphRouting.BuildSpanningTree` no longer require to have at least one hint line. 
+- `AdaptiveGraphRouting.BuildSpanningTree` no longer require to have at least one hint line.
+- Don't log all vertex creation actions during Debug mode geometry generation.
 
 
 ### Fixed
--- a/Elements/src/Geometry/Tessellation/Tessellation.cs
+++ b/Elements/src/Geometry/Tessellation/Tessellation.cs
@@ -13,7 +13,10 @@ namespace Elements.Geometry.Tessellation
     /// </summary>
     internal static class Tessellation
     {
-        private static bool LOG_TESSELATION = false;
+        // TODO remove this when we have a logging system with more granular control over logging levels.
+        // Switch this to true to see the tessellation progress of all elements.
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+        public static bool LOG_TESSELATION = false;
 
         /// <summary>
         /// Triangulate a collection of CSGs and pack the triangulated data into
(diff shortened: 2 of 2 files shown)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/897)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Geometry/Tessellation/Tessellation.cs` line 16 at r1](https://reviewable.io/reviews/hypar-io/elements/897#-NE5lYbc3a8mWI0E3Hxo:-NE5lYbc3a8mWI0E3Hxp:b-7lue38) ([raw file](https://github.com/hypar-io/elements/blob/42108a2ba230e43831d28e980bc722ed585f99bc/Elements/src/Geometry/Tessellation/Tessellation.cs#L16)):*
> ```Smalltalk
>     internal static class Tessellation
>     {
>         private static bool LOG_TESSELATION = false;
> ```

The intention being that you'd just edit the code if you wanted to see the logging messages? We might consider an "unlisted" static boolean that's public, similar to `SkipCSGUnion`, so a power user who's not editing `Elements` but still wanted this logging could enable it. 

(IOW, public but with `   [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]` annotation)

But I'm not sure that case is particularly likely, since if you're debugging `Tesselate`, you probably need to edit Elements anyway. Your call. In any case, would be good to have a comment above describing what this is for, even if it's relatively obvious.

ikeough:
Just comment the lines out for now. This is something that a logging framework can take care of for us, with varied levels of logging verbosity.

wynged (commented):
added a TODO that we should remove this when we have a logging framework.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/897#-:-NE8Xl839K7QLUa_1e8J:b902kd4)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Tessellation/Tessellation.cs` line 16 at r1](https://reviewable.io/reviews/hypar-io/elements/897#-NE5lYbc3a8mWI0E3Hxo:-NE8XfbM9KbwFMq4UBO1:bno7bp2) ([raw file](https://github.com/hypar-io/elements/blob/42108a2ba230e43831d28e980bc722ed585f99bc/Elements/src/Geometry/Tessellation/Tessellation.cs#L16)):*

The intention being that you'd just edit the code if you wanted to see the logging messages? We might consider an "unlisted" static boolean that's public, similar to `SkipCSGUnion`, so a power user who's not editing `Elements` but still wanted this logging could enable it. 

(IOW, public but with `   [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]` annotation)

But I'm not sure that case is particularly likely, since if you're debugging `Tesselate`, you probably need to edit Elements anyway. Your call. In any case, would be good to have a comment above describing what this is for, even if it's relatively obvious.  
</blockquote></details>

added comment and TODO that we should remove this when we get a more robust logging system.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/897)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #898, opened 2022-10-10
Certain networks experience infinite loop when finding closed regions.


ikeough:
**Describe the bug**
For certain networks, calling `FindAllClosedRegions(...)` experiences an infinite loop related to the graph traversing back over itself. We previously handled this with a break when a loop was found, but there are "spiral" cases where the traversal is not exactly a loop.

**Expected behavior**
Loops should be cut and tails removed.

PR #899, opened 2022-10-10
Fix 898

ikeough:
BACKGROUND:
See #898.

DESCRIPTION:
This PR updates the network traversal logic for finding closed regions to have more robust loop recognition.

TESTING:
See the `NetworkTests.FindAllClosedRegionsDoesNotLoopInfinitely` test.

REQUIRED:

Prior to this fix, this network could not find all of its closed regions. Note the missing region in the lower right, which was the problem region. This region is now missing because it doesn't create a valid polygon, but it is no longer causing the region calculation to loop infinitely.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/899)


ikeough:
@andrewheumann this is related to the vertical circulation elements in Hypar being placed directly on top of the edge of the floor.

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/899)* status: :shipit: complete! 1 of 1 approvals obtained

PR #908, opened 2022-10-21
Fixes to planar trimming.

ikeough:
BACKGROUND:
While working on calculating complex allowable envelope shapes, involving clipping lots of polygons with planes at odd angles, I found that there were numerous failure modes for polygon trim operations with planes.

DESCRIPTION:
- Simplifies the `Trimmed(...)` method while handling all existing cases and new cases like planes which intersect a polygon exactly at its corner. 
- Improves `Ray` intersection. Previously we didn't disambiguate parallel and coincident rays correctly.
- Standardizes on 2D winding calculation for point inclusion in polygons and removes ray casting entirely. For 3D point inclusion, we now flatten and use winding. 

TESTING:
Several new tests have been added with specific test cases taken from the work mentioned above. In addition, several existing test cases have been split apart into "one assertion per test" tests for easier debugging of failure cases.

REQUIRED:

Here's an example of the kind of geometry which demonstrated failures to trim previously. Note that in many cases for the planes labelled "sky", that the plane often intersects some of the polygons exactly at a corner.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Search/Vector3ToleranceComparer.cs` line 9 at r1](https://reviewable.io/reviews/hypar-io/elements/908#-NEzIl2p7n8KkMN_I4qW:-NEzIl2p7n8KkMN_I4qX:b60bg5b) ([raw file](https://github.com/hypar-io/elements/blob/ec29b19a70c03827f50ee453c7f018fcbfd7dc84/Elements/src/Search/Vector3ToleranceComparer.cs#L9)):*
> ```Smalltalk
>     /// Vector3 
>     /// </summary>
>     public class Vector3ToleranceComparer : IEqualityComparer<Vector3>
> ```

This seems like nearly a duplicate of `Vector3Comparer`.  Can we consolidate the two? 
`Vector3Comparer` has a caution about `Distinct` and other methods that use `GetHashCode` — I didn't realize if it encountered the same hash code it would fall through to `Equals`, so perhaps it is safe to do this after all (and revise the caution accordingly)

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Search/Vector3ToleranceComparer.cs` line 9 at r1](https://reviewable.io/reviews/hypar-io/elements/908#-NEzIl2p7n8KkMN_I4qW:-NEzLXo76M9Uq1bIlB0v:b-tgzyj1) ([raw file](https://github.com/hypar-io/elements/blob/ec29b19a70c03827f50ee453c7f018fcbfd7dc84/Elements/src/Search/Vector3ToleranceComparer.cs#L9)):*

This seems like nearly a duplicate of `Vector3Comparer`.  Can we consolidate the two? 
`Vector3Comparer` has a caution about `Distinct` and other methods that use `GetHashCode` — I didn't realize if it encountered the same hash code it would fall through to `Equals`, so perhaps it is safe to do this after all (and revise the caution accordingly)
</blockquote></details>

Also — am I crazy or is this new Comparer not being used anywhere?

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Search/Vector3ToleranceComparer.cs` line 9 at r1](https://reviewable.io/reviews/hypar-io/elements/908#-NEzIl2p7n8KkMN_I4qW:-NFA2fMk4UPEsWCRssBw:b-j50tti) ([raw file](https://github.com/hypar-io/elements/blob/ec29b19a70c03827f50ee453c7f018fcbfd7dc84/Elements/src/Search/Vector3ToleranceComparer.cs#L9)):*

Also — am I crazy or is this new Comparer not being used anywhere?
</blockquote></details>

It's not! Apologies. It was used in an earlier iteration of this work. It has been removed.

Yes, it's how the comparers work that they first try the hash and fall back to the _actual_ comparison if they have hash collision. I'll leave revising that comparer to another PR though because it's now completely out of scope :)

andrewheumann (changes requested):
> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Polygon.cs` line 548 at r2](https://reviewable.io/reviews/hypar-io/elements/908#-NFAPUQV4d-IFSQaQgvU:-NFAPUQV4d-IFSQaQgvV:baaklum) ([raw file](https://github.com/hypar-io/elements/blob/a7d00829cdc0255c2128a6baf42154c410ae6d96/Elements/src/Geometry/Polygon.cs#L548)):*
> ```Smalltalk
>         /// the default test axis.</param>
>         /// <returns>True if the point is contained in the polygon, otherwise false.</returns>
>         internal bool Contains3D(Vector3 point, bool unique = true, bool useRandomRay = false)
> ```

`useRandomRay` makes me twitchy..... anything non-deterministic can cause issues that are impossible to reproduce and test. If this was failing with the X axis, then it will occasionally happen to fail along the random ray, and it will be impossible to figure out why it failed, because it won't fail the next time. It seems like `Contains3D` should be able to work without this.

Elements/test/ObstacleTests.cs
```diff
@@ -235,7 +235,7 @@ public void IntersectsObstacleFromLine()
             Assert.True(horizontalObstacle.Intersects(horizontalLineOnBottom));
 
             var horizontalLineOnSide = horizontalLine.TransformedLine(new Transform(offset, 0, 0));
-            Assert.False(horizontalObstacle.Intersects(horizontalLineOnSide));
```
ikeough, line 238: @wynged @DmytroMuravskyi This image shows this test geometrically. The `horizontalLine` is in the plane of the face of the obstacle. This is rightly classified as an intersection by our code now, but was previously asserted to be false. Why?

image

Elements/test/ObstacleTests.cs
```diff
@@ -113,7 +113,7 @@ public void IntersectsRotatedPolygon()
 
             var result = obstacle.Intersects(polyline);
 
-            Assert.False(result);
```
ikeough, line 116: @wynged @DmytroMuravskyi This image shows this test geometrically. The `polyline` is along the edge of the rotated obstacle. This is rightly classified as an intersection by our code now, but was previously asserted to be false. Why?

image

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Polygon.cs` line 548 at r2](https://reviewable.io/reviews/hypar-io/elements/908#-NFAPUQV4d-IFSQaQgvU:-NH1DujqDGagd5A9j2nk:b-vin4vf) ([raw file](https://github.com/hypar-io/elements/blob/a7d00829cdc0255c2128a6baf42154c410ae6d96/Elements/src/Geometry/Polygon.cs#L548)):*

I've solved this using the strategy recommended in Preparata and Shamos and the wikipedia article which is to only add a vertex intersection when the other vertex of the edge is to the "left" of the ray. This is a clever way to deduplicate ray intersections through vertices _and_ it removes a call to `Contains`.
</blockquote></details>

We already had winding number calculation for the flat case. Instead of implementing yet another way to check for containment in a polygon, I updated this to use the "flat" algorithm by transforming the polygon into the XY plane. The ray/polygon intersection now simply calls the `Contains3D` method which transforms and calls the `Contains` method into the 3D plane.

andrewheumann (approved):
One little comment, otherwise looks good to me! 

> Reviewed 4 of 4 files at r3, 11 of 14 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908#-:-NH22ciF1haVL30GS0FS:b4adm49)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/test/SolidTests.cs` line 610 at r5](https://reviewable.io/reviews/hypar-io/elements/908#-NH21U1y55OyeGutjKY0:-NH21U1z9CFvMpeyhnZi:b-qcrlyu) ([raw file](https://github.com/hypar-io/elements/blob/ac12cb9cdb0d54f42971c4902aca044124d4a071/Elements/test/SolidTests.cs#L610)):*
> ```Smalltalk
>             Assert.Equal(20, result1.Faces.Count);
>             Assert.Equal(32, result1.Vertices.Count);
>             // Assert.Equal(48, result1.Edges.Count);
> ```

This was commented out — intentional? should we remove / update the test / uncomment?

DmytroMuravskyi (commented):
It's hard to tell for sure what's the problem with Obstacle.Intersects as i'm not the author of the function but it looks like it doesn't find all the intersections when it's on the border. It feels inconsistent as it says that some lines on border are intersected but other don't. It should behave the same way for all borders or at least have "includeBorders" parameter as sometimes you want to go around obstacle but not inside. I will keep this problem in mind for the next time I'll have spare time

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908#-:-NH4brf54m2knbkvaBsa:b-rr1fbr)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

DmytroMuravskyi (commented):
Oh, so it IS consistent now. I think the changes are great and if we would get any troubles in downstream function then I'll adjust downstream function or add a parameter here

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908#-:-NH5I-Ag2-McMAL1l_OZ:b-ja5t97)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/908)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/test/SolidTests.cs` line 610 at r5](https://reviewable.io/reviews/hypar-io/elements/908#-NH21U1y55OyeGutjKY0:-NH6H3OO76CT0uuZGgTx:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/ac12cb9cdb0d54f42971c4902aca044124d4a071/Elements/test/SolidTests.cs#L610)):*

This was commented out — intentional? should we remove / update the test / uncomment? 
</blockquote></details>

Done.

PR #914, opened 2022-11-01
set current culture to en-US so numbers are correct

wynged:
BACKGROUND:
- User in another country reported that content catalogs were being generated with invalid nubmer formatting where `,` is used instead of `.` 
image

DESCRIPTION:
- Set the current culture to `en-US` so numbers are correctly formatted

TESTING:
- I tested this by setting the culture to one that would create this problem, for example "pl-PL" and seeing that this did indeed affect the outcome, so I presume that it will fix the issue in reverse.
  
FUTURE WORK:
- Other places in code taht need this same treatment?  it seems like it most places we don't have explicity values for numbers so this isn't really an issue.

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/914)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - `Line.IsCollinear` - added `tolerance` parameter.
 - `Polygon.CollinearPointsRemoved` - added `tolerance` parameter.
 - `Line.TryGetOverlap` - added `tolerance` parameter.
+- `CatalogGenerator` always uses en-US culture.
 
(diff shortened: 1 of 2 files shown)


cdriesler (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/914)* status: :shipit: complete! 1 of 1 approvals obtained

PR #915, opened 2022-11-02
Add TreeNode and use it for AdaptiveGraphRouting. Do additional step of leaf routing.

DmytroMuravskyi:
BACKGROUND:
- Were reported next issues with AdaptiveGraphRouting.
A) When leaf can connect to the tree in several points with the same cost - random path is chosen.
B) Longer as expected leaf is user for leafs that are connected to the tree first.  

DESCRIPTION:
- Added TreeNode and use it instead of just dictionary. This allows to track leaf/trunk connections in the tree.
- If several connection points on the tree have the same cost, one is chosen that is closer to the end.
- Do additional step of leaf routing. When trunk is constructed leafs can find better connection point. 
- Make decision making for choosing where to connect to the trunk more predictable:
A) Lowest travel cost
B) If same - closest point to the exit
C) From two calculated paths choose the one that has no turn with trunk direction.
- Moved common logic to RouteBranch function.

TESTING:
- Added new tests: 
A) AdaptiveGraphRoutingEqualLeafCheck
B) AdaptiveGraphRoutingClosestLeafCheck
C) AdaptiveGraphRoutingCorrectTurn

FUTURE WORK:
- AdaptiveGraphRouting.ConnectionPoint went through changes to make it more predictable but it may require more polishing after additional testing and feedback.

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/915)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - `RoutingHintLine.Is2D`
 - `Obstacle.Orientation`
 - `Elements.Spatial.AdaptiveGrid.EdgeInfo`
+- `Elements.Spatial.AdaptiveGrid.TreeNode`
 - `IEnumerable<Vector3>.UniqueWithinTolerance(double tolerance = Vector3.EPSILON)`
 - `Plane.XY`, `Plane.XZ`, and `Plane.YZ` static properties
 
@@ -23,7 +24,7 @@
 - `RoutingVertex` - removed `Guides`.
 - `AdaptiveGraphRouting.BuildSpanningTree` functions are simplified. Also, they use only single `tailPoint` now.
 - `AdaptiveGraphRouting.BuildSpanningTree` no longer require to have at least one hint line.
-- `AdaptiveGraphRouting.BuildSpanningTree` no longer require to have at least one hint line.
+- `AdaptiveGraphRouting.BuildSpanningTree` and `AdaptiveGraphRouting.BuildSimpleNetwork` now return `IDictionary<ulong, TreeNode>`.
 - Don't log all vertex creation actions during Debug mode geometry generation.
 - `Polyline.GetSubsegment` changes direction of output polyline when parameters reversed
 - `Line.IsCollinear` - added `tolerance` parameter.
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs
@@ -1038,11 +1038,15 @@ namespace Elements.Spatial.AdaptiveGrid
                         candidateBranch = BranchSide.Right;
                     }
 
+                    //Connection point with the lower cost is better. 
                     if (candidateCost > bestCost + Vector3.EPSILON)
                     {
                         continue;
                     }
 
+                    //If two ways to travel from point A to point B has the same cost,
+                    //the one is used that joins the tree without a turn. 
+                    //We both need/don't need a turn - first path is chosen.
                     if (costs.Item1.ApproximatelyEquals(costs.Item2) &&
                         tree.TryGetValue(index, out var node) && node.Trunk != null)
                     {
@@ -1074,6 +1078,8 @@ namespace Elements.Spatial.AdaptiveGrid
                         }
                     }
 
+                    //If several connection points on the tree have the same cost,
+                    //one is chosen that is closer to the end.
                     if (candidateCost.ApproximatelyEquals(bestCost))
                     {
                         if (bestCostToTrunk < 0)
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs
@@ -622,8 +622,8 @@ namespace Elements.Spatial.AdaptiveGrid
                 var vertex = _grid.GetVertex(u);
                 foreach (var e in vertex.Edges)
                 {
-                    var edgeWeight = edgeInfos[e.Id];
-                    if (edgeWeight.Factor == double.PositiveInfinity)
+                    var edgeInfo = edgeInfos[e.Id];
+                    if (edgeInfo.Factor == double.PositiveInfinity)
                     {
                         continue;
                     }
@@ -658,7 +658,7 @@ namespace Elements.Spatial.AdaptiveGrid
                     }
 
                     //Compute cost of each its neighbors as cost of vertex we came from plus cost of edge.
-                    var newWeight = EdgeCost(edgeWeight);
+                    var newWeight = EdgeCost(edgeInfo);
                     BranchSide bestBranch = BranchSide.Left;
 
                     //We need as little change of direction as possible. A penalty is added if
@@ -670,7 +670,7 @@ namespace Elements.Spatial.AdaptiveGrid
                         if (startDirection.HasValue &&
                             !Vector3.AreCollinearByAngle(_grid.GetVertex(startDirection.Value).Point, vertex.Point, v.Point))
                         {
-                            newWeight += TurnCost(edgeWeight, vertex, startDirection.Value, edgeInfos);
+                            newWeight += TurnCost(edgeInfo, vertex, startDirection.Value, edgeInfos);
                         }
                     }
                     else
@@ -682,7 +682,7 @@ namespace Elements.Spatial.AdaptiveGrid
                         var leftCost = cost.Item1 + newWeight;
                         if (!leftCollinear)
                         {
-                            leftCost += TurnCost(edgeWeight, vertex, leftBefore.Id, edgeInfos);
+                            leftCost += TurnCost(edgeInfo, vertex, leftBefore.Id, edgeInfos);
                         }
 
                         var rigthCost = Double.MaxValue;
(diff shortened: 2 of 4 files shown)


wynged (changes requested):
very close, seems like a great improvement in terms of code organization.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/915#-:-NFtIlQcDzL8YgoIB2gS:b-kc4498)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 285 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtEQen5_rntC9TDcqh:-NFtEQen5_rntC9TDcqi:b-heiim5) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L285)):*
> ```Smalltalk
> 
>         private void RouteBranch(
>             List<ulong> leafTerminals, ulong trunkTerminal,
> ```

format each argument on own line.



_Code quote:_
```Smalltalk
magnetTerm
```
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 295 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtEVAP7DAZTlZChOBw:-NFtEVAP7DAZTlZChOBx:bxvdvdi) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L295)):*
> ```Smalltalk
>             //One at a time we choose the connection point that is cheapest to travel to existing
>             //network and its path is added to the network until all are added.
>             HashSet<ulong> magnetTerminals = new HashSet<ulong> { trunkTerminal };
> ```

can you be more explicit about how the magnetTerminals are used?  I think it's something like:
"Magnet terminals are all of the vertices that are currently part of the tree.  They are used as the list of good destinations while routing the leaves so that we join the tree as soon as possible."
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 657 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtA2vR5H_iFFhl7Ihk:-NFtA2vS5wOaEfoUQYKa:b-1bk34k) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L657)):*
> ```Smalltalk
> 
>                     //Compute cost of each its neighbors as cost of vertex we came from plus cost of edge.
>                     var newWeight = EdgeCost(edgeWeight);
> ```

rename this `edgeWeight` to `edgeInfo` I think for clarity.
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 1087 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtGoiw7aLzG_2fhR_c:-NFtGoiw7aLzG_2fhR_d:b-o9sdfz) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L1087)):*
> ```Smalltalk
>                         if (bestCostToTrunk < 0)
>                         {
>                             bestCostToTrunk = CostToTrunk(tree[bestIndex], weights);
> ```

see note below about how this CostToTrunk doesn't protect us from a TreeNode that is not actaully successfully connected to the trunk TreeNode.



_Code quote:_
```Smalltalk
CostToTrunk
```
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 1121 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtGFcdDxzVtx72mR5S:-NFtGFcdDxzVtx72mR5T:bdrzbs0) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L1121)):*
> ```Smalltalk
>         }
> 
>         private double CostToTrunk(
> ```

this doesn't necessarilly get all the way to the final Trunk of the tree, it depends on the `start` TreeNode actually being connected to that trunk correctly right?  Do we want to also provide a "trunk" ulong/TreeNode that is the destination so that if we don't connect we can either error or return infinite cost?
___
*[`Elements/test/AdaptiveGraphRoutingTests.cs` line 553 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtHtP92Ene6F__kvG7:-NFtHtPA2RAHlKVeDrdZ:bfbu4b4) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/test/AdaptiveGraphRoutingTests.cs#L553)):*
> ```Smalltalk
>             foreach (var input in inputVertices.SelectMany(iv => iv))
>             {
>                 var node = tree[input.Id];
> ```

we use this exact pattern at least  twice, method that returns "TreeLines" I think.  Or really, shouldn't we have a GridVisualization method that does what we want for tests as well as for general visualization?

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/915)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 657 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtA2vR5H_iFFhl7Ihk:-NFxXyGyEdZqU-ybGykO:bfvaywg) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L657)):*

rename this `edgeWeight` to `edgeInfo` I think for clarity.
</blockquote></details>

just waiting on this one I think.

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/915)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 295 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtEVAP7DAZTlZChOBw:-NFxYRu6CPZMi35lC8Wt:b-896fjx) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L295)):*

can you be more explicit about how the magnetTerminals are used?  I think it's something like:
"Magnet terminals are all of the vertices that are currently part of the tree.  They are used as the list of good destinations while routing the leaves so that we join the tree as soon as possible."
</blockquote></details>

Done
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 657 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtA2vR5H_iFFhl7Ihk:-NFx_R9o6NGImmrH0P80:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L657)):*

just waiting on this one I think.
</blockquote></details>

Done.
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs` line 1121 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtGFcdDxzVtx72mR5S:-NFxYg1X8kHmoWgB3KOB:b-yz2xkn) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/src/Spatial/AdaptiveGrid/AdaptiveGraphRouting.cs#L1121)):*

this doesn't necessarilly get all the way to the final Trunk of the tree, it depends on the `start` TreeNode actually being connected to that trunk correctly right?  Do we want to also provide a "trunk" ulong/TreeNode that is the destination so that if we don't connect we can either error or return infinite cost?
</blockquote></details>

Yes, right now it assumes that last node is what we need. I don't except it to fail. In my next PR I plan to finally add error messages to Routing and then if cost is PositiveInfinity - end can't be reached due to filters or angles, if MaxDouble - due to not being connected. Let's leave it as is for now as I don't know how it will end up yet.  



_Code quote:_
```Smalltalk
edgeWeight
```
___
*[`Elements/test/AdaptiveGraphRoutingTests.cs` line 553 at r1](https://reviewable.io/reviews/hypar-io/elements/915#-NFtHtP92Ene6F__kvG7:-NFxOqoc4B7p_KZJ9Wjy:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/acc20925416af2b90431626785a673253c1a1572/Elements/test/AdaptiveGraphRoutingTests.cs#L553)):*

we use this exact pattern at least  twice, method that returns "TreeLines" I think.  Or really, shouldn't we have a GridVisualization method that does what we want for tests as well as for general visualization?
</blockquote></details>

Done.



_Code quote:_
```Smalltalk
foreach (var input in inputVertices.SelectMany(iv => iv))
```

PR #921, opened 2022-11-28
Switch to the metal roughness model for glTF.

ikeough:
BACKGROUND:
When integrating the three gpu path tracer in Hypar, I noticed that materials were overly reflective. After opening https://github.com/gkjohnson/three-gpu-pathtracer/issues/310, I was informed that the extension we have used previously, `KHR_materials_pbrSpecularGlossiness` is being "archived" in favor of newer extensions.

DESCRIPTION:
This PR updates our glTF serialization to support `KHR_materials_specular` and `KHR_materials_ior`, with conversions as noted in https://github.com/donmccurdy/glTF-Transform/blob/d77ca6a12c5b56efa1b6594806450dd38df19b25/packages/functions/src/metal-rough.ts

TESTING:
Run a test and export a glTF file. The results should look very similar. I've exported several glTF's and run them through Don McCurdy's viewer and they pass all validation and look correct. I've updated the Materials example to indicated roughness and specularity.

Here it is in Don McCurdy's viewer (latest three):
image

Here it is in Hypar (note: different HDRI environments are used)
image

FUTURE WORK:
- The `Material` class should be updated with properties that align with the metallic roughness model.

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/921)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 ### Fixed
 
 - Fix `Obstacle.FromLine` if line is vertical and start point is positioned higher than end.
+- Materials exported to glTF now have their `RoughnessFactor` set correctly.
+- Materials exported to glTF no longer use the `KHR_materials_pbrSpecularGlossiness` extension, as this extension is being sunset in favor of `KHR_materials_specular` and `KHR_materials_ior`.
 
 ## 1.3.0
 
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -197,11 +197,16 @@ namespace Elements.Serialization.glTF
                 var gltfMaterial = new glTFLoader.Schema.Material();
                 newMaterials.Add(gltfMaterial);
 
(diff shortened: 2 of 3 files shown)


andrewheumann:
This looks good to me! Have we tested on Hypar's 3d viewer / do we know if our existing GLTF loader already supports these extensions out of the box? We shouldn't merge this until that's verified + tested, so we don't break our ability to rely on alphas for hypar function development.

ikeough:
Good question. Threejs shows support for these extensions already: https://threejs.org/docs/#examples/en/loaders/GLTFLoader. So as long as we stay current with three, it should work. I have tried the drag and drop gltf workflow on Hypar and things look fine which is expected even without these extensions because it defaults to the metallness-roughness workflow. What I need to do is build a test with lots of shiny and matte stuff.

andrewheumann (approved):
It looks like this support is already in the version of three we are using, so I'm OK to proceed w/ this one, but more tests would be valuable regardless! 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/921#-:-NHzmancCo-x7zky5Pxe:b-yr1f6p)* status: :shipit: complete! 1 of 1 approvals obtained

PR #934, opened 2022-12-28
Clean profiles

andrewheumann:
BACKGROUND:
- When editing profiles via overrides in the Hypar web UI, certain configurations result in "illegal" topologies, like infinitely thin triangles, which break the editor. Some examples of these configurations are:
   - adjacent profiles that are nearly-but-not-quite aligned
   - profiles where one adjacent edge has an intermediate vertex but the other does not
- This is avoided by providing a "clean" set of profiles.

DESCRIPTION:
- Adds a static `Profile.Cleaned` extension method which cleans up a set of profiles into a clean graph which can be edited.

TESTING:
- I added several tests for the method which passed
- I have also been utilizing this method in production in a real hypar function, with good success. I am moving it to elements because I have another function that needs it. 
  
FUTURE WORK:
- the Hypar web UI should become more resilient to bad geometry input. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/934)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - `AdaptiveGraphRouting.AddPlaneModifier(string name, Plane plane, double factor)`
 - `SvgSection.SaveAsSvg`, `SvgSection.SaveAsPdf`
 - `Network.TraverseLeftWithoutLeaves()`
+- `Profile.Cleaned()`
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cdriesler (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/934)* status: :shipit: complete! 1 of 1 approvals obtained

PR #936, opened 2022-12-29
fix mesh merge exception

andrewheumann:
BACKGROUND:
- Adding the first vertex to a mesh with "merge" on would throw an exception.

DESCRIPTION:
- Fixes the exception.

TESTING:
-Added a new test which failed and now passes.

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/936)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Gltfs that are merged that require additional extensions will also merge their extensions.
 - Don't try to save test models that have no name, they can interfere with each other because they want to save to the same file.
 - Fixed an issue where `Grid2d.GetCells()` multiple times could fail to return the correct results on subsequent calls, because changes to the axis grids were not invalidating the grid's computed cells.
+- Adding the first vertex to a mesh with `merge: true` would throw an exception, this is fixed.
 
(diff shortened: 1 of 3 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/936)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Mesh.cs` line 351 at r1](https://reviewable.io/reviews/hypar-io/elements/936#-NKU-8k6ChRhO19fmAlF:-NKU-8k6ChRhO19fmAlG:b-duzb24) ([raw file](https://github.com/hypar-io/elements/blob/438c5bb911923db37d0fe095548f4d27180140b3/Elements/src/Geometry/Mesh.cs#L351)):*
> ```Smalltalk
>             {
>                 // if we've never added any vertices, we may not have a valid bbox, so we have to guess at a default size.
>                 var bbox = _bbox.IsValid() ? _bbox : new BBox3((-10, -10, -10), (10, 10, 10));
> ```

that's a  big default box, but approved!

PR #937, opened 2023-01-09
store ID and selectability in GLTF without relying on `name`

andrewheumann:
BACKGROUND:
- In early days of Elements, the `name` of the object in the GLTF wasn't utilized for anything. Over time, the Hypar front end has evolved to rely on parsing out critical information from the name — specifically, the element Id, and whether or not it is meant to be selectable. 

DESCRIPTION:
- Establishes an ad-hoc gltf extension called `HYPAR_info` which contains information necessary for the hypar front end to connect element data to GLTF mesh objects. 

TESTING:
- I added a test which verifies that the data is written correctly to the GLTF. 
- I have an in-progress `explore` branch which can read this info if it's present. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/elements/937)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Removed rule exception from `AdaptiveGraphRouting` that prevented vertical edges turn cost being discounter.
 - `Message.FromLine` is set obsolete.
 - In `AdaptiveGridRouting`, if there are several connection points with the same cost - choose one that is closer to the trunk.
+- GLTF writing now includes an ad-hoc `HYPAR_info` extension which aids in mapping between GLTF content and element ids in the model.
 
 ### Fixed
 
--- a/Elements/src/GeometricElement.cs
+++ b/Elements/src/GeometricElement.cs
@@ -25,6 +25,11 @@ namespace Elements
         internal BBox3 _bounds;
         internal Csg.Solid _csg;
 
+        // Used to attach a "selectable: false" flag to the element in the
+        // generated GLB. NOTE: currently only considered in the "ModelLines /
+        // ModelPoints" pathways — this is not yet supported for mesh elements.
+        internal bool _isSelectable = true;
+
         /// <summary>
         /// The element's bounds.
         /// The bounds are only available when the geometry has been
--- a/Elements/src/ModelCurve.cs
+++ b/Elements/src/ModelCurve.cs
@@ -18,8 +18,6 @@ namespace Elements
         /// </summary>
         public Curve Curve { get; set; }
 
-        private bool _isSelectable = true;
-
         /// <summary>
         /// Create a model curve.
         /// </summary>
--- a/Elements/src/ModelLines.cs
+++ b/Elements/src/ModelLines.cs
@@ -19,8 +19,6 @@ namespace Elements
         /// </summary>
         public IList<Line> Lines { get; set; }
 
-        private bool _isSelectable = false;
-
         /// <summary>
         /// Create a collection of lines. They share Material, Transformation and other parameters.
         /// </summary>
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -482,7 +482,7 @@ namespace Elements.Serialization.glTF
                     }}
                 }
                 };
-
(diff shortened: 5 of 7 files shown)


wynged (changes requested):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/937)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Serialization/glTF/NodeUtilities.cs` line 171 at r2](https://reviewable.io/reviews/hypar-io/elements/937#-NLSG6WX3f5zMXjriyjN:-NLSG6WX3f5zMXjriyjO:b-ydawtv) ([raw file](https://github.com/hypar-io/elements/blob/d98fc6a6e89d77c2a787aa5a081216be57caed5d/Elements/src/Serialization/glTF/NodeUtilities.cs#L171)):*
> ```Smalltalk
>         }
> 
>         internal static int CreateNodeForMesh(int meshId, List<glTFLoader.Schema.Node> nodes, Guid elementId, Transform transform = null)
> ```

seems like elementId this should also be optional with default=null since it's optional in all other methods that take an elementId?  Is there some reason we are sure to have a value here but not in the other methods?
___
*[`Elements/src/Serialization/glTF/NodeUtilities.cs` line 195 at r2](https://reviewable.io/reviews/hypar-io/elements/937#-NLSGRLT6MIUIFdyjoMj:-NLSGRLT6MIUIFdyjoMk:b-cz1zmf) ([raw file](https://github.com/hypar-io/elements/blob/d98fc6a6e89d77c2a787aa5a081216be57caed5d/Elements/src/Serialization/glTF/NodeUtilities.cs#L195)):*
> ```Smalltalk
>             if (selectable.HasValue)
>             {
>                 node.Extensions["HYPAR_info"] = new Dictionary<string, object>
> ```

nit.  I think my code preference would be to make the dictionary with the "id", optionally add the "selectable" property and then always assign the dictionary to the dictionary to `Extensions["HYPAR_info"]` once the correct dictionary is built.  feels like a more extensible pattern.

wynged (commented):
mostly just nits.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/937#-:-NLSKDqj4QY9f4wLQgoL:bmzv4tn)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/937)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Serialization/glTF/NodeUtilities.cs` line 171 at r2](https://reviewable.io/reviews/hypar-io/elements/937#-NLSG6WX3f5zMXjriyjN:-NLSKjbo0gcvQjjiByN7:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/d98fc6a6e89d77c2a787aa5a081216be57caed5d/Elements/src/Serialization/glTF/NodeUtilities.cs#L171)):*

seems like elementId this should also be optional with default=null since it's optional in all other methods that take an elementId?  Is there some reason we are sure to have a value here but not in the other methods?
</blockquote></details>

Done.
___
*[`Elements/src/Serialization/glTF/NodeUtilities.cs` line 195 at r2](https://reviewable.io/reviews/hypar-io/elements/937#-NLSGRLT6MIUIFdyjoMj:-NLSLPle53xShWgxwRQ4:b-896fix) ([raw file](https://github.com/hypar-io/elements/blob/d98fc6a6e89d77c2a787aa5a081216be57caed5d/Elements/src/Serialization/glTF/NodeUtilities.cs#L195)):*

nit.  I think my code preference would be to make the dictionary with the "id", optionally add the "selectable" property and then always assign the dictionary to the dictionary to `Extensions["HYPAR_info"]` once the correct dictionary is built.  feels like a more extensible pattern.
</blockquote></details>

Done.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/elements/937)* status: :shipit: complete! 1 of 1 approvals obtained

PR #940, opened 2023-02-07
add support for flipped extrude

andrewheumann:
BACKGROUND:
- For Hypar Space, we rely on a visual trick for our spaces — we turn them inside out, with their normals inwards, so that you only see the faces on the opposite side of the room, and can look through the top to see the furniture, and easily select it. 
- This is presently achieved by manually building a `ConstructedSolid` direct from faces, since `Extrude` has a built-in check to ensure that normals point out. This works fine, but means that we serialize much more data than is needed in the model: we have to store every vertex, edge, and face, which makes the model json heavy. 

DESCRIPTION:
- Adds an optional `Flipped` property, which inverts the normal check when building the extrude. if `Flipped` is true, the profile is flipped w/r/t the extrusion direction, resulting in an inside-out shape. 

TESTING:
- I added a new test which verifies technically and visually that the `Flipped` parameter has the desired effect.
image

FUTURE WORK:
- It is conceivable, if a bit unlikely, that other primitives might want similar behavior. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/940)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.5.0
+
+### Added
+
+- `Extrude` Solid Operation supports an optional `Flipped` parameter to purposely turn its normals inside out.
+
 ## 1.4.0
 
 ### Added
--- a/Elements/src/Geometry/Kernel.cs
+++ b/Elements/src/Geometry/Kernel.cs
@@ -38,9 +38,9 @@ namespace Elements.Geometry
         /// Create an extrude.
         /// </summary>
         /// <returns>A solid.</returns>
-        public Solid CreateExtrude(Profile profile, double depth, Vector3 direction)
+        public Solid CreateExtrude(Profile profile, double depth, Vector3 direction, bool flipped = false)
         {
-            if (profile.Perimeter.Normal().Dot(direction) < 0)
+            if (profile.Perimeter.Normal().Dot(direction) < 0 != flipped)
             {
                 profile = profile.Reversed();
             }
--- a/Elements/src/Geometry/Solids/Extrude.cs
+++ b/Elements/src/Geometry/Solids/Extrude.cs
@@ -12,6 +12,7 @@ namespace Elements.Geometry.Solids
         private Profile _profile;
         private double _height;
         private Vector3 _direction;
+        private bool _flipped;
 
         /// <summary>The id of the profile to extrude.</summary>
         [JsonProperty("Profile", Required = Required.AllowNull)]
@@ -59,6 +60,21 @@ namespace Elements.Geometry.Solids
             }
         }
 
+        /// <summary>Is the extrusion flipped inside out?</summary>
+        [JsonProperty("Flipped")]
+        public bool Flipped
+        {
+            get { return _flipped; }
+            set
+            {
+                if (_flipped != value)
+                {
+                    _flipped = value;
+                    RaisePropertyChanged();
+                }
+            }
+        }
+
         /// <summary>
         /// Construct an extrusion.
         /// </summary>
@@ -67,7 +83,7 @@ namespace Elements.Geometry.Solids
         /// <param name="direction"></param>
         /// <param name="isVoid"></param>
         [JsonConstructor]
-        public Extrude(Profile @profile, double @height, Vector3 @direction, bool @isVoid)
+        public Extrude(Profile @profile, double @height, Vector3 @direction, bool @isVoid = false, bool flipped = false)
             : base(isVoid)
         {
             if (!Validator.DisableValidationOnConstruction)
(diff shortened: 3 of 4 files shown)


anthonie-kramer (changes requested):
Looks good! I was just curious about what the expected behavior should be if there are voids defined. I tested it out just to see and this is what it looks like with the current code:


I think this is preferred over the inverse for voids, but just wanted to point it out!

Code if you want to copy/paste for yourself:
```
var profileNormal = new Profile(Polygon.Rectangle(1, 1), Polygon.Rectangle(0.5, 0.5));
var normalExtrude = new Extrude(profileNormal, 1, Vector3.ZAxis);

var profileFlipped = new Profile(Polygon.Rectangle(1, 1).TransformedPolygon(new Transform(2, 0, 0)), Polygon.Rectangle(0.5, 0.5).TransformedPolygon(new Transform(2, 0, 0)));
var flippedExtrude = new Extrude(profileFlipped, 1, Vector3.ZAxis, false, true);

var geoNormal = new GeometricElement
{
Representation = new Representation(normalExtrude)
};

var geoFlipped = new GeometricElement
{
Representation = new Representation(flippedExtrude)
};
```

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/940#-:-NOCF-GX-Asp6ORzUKZi:bxb0yy1)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Solids/Extrude.cs` line 84 at r1](https://reviewable.io/reviews/hypar-io/Elements/940#-NOCHN9GAOXuD0xEOaIB:-NOCHN9GAOXuD0xEOaIC:b-grpa7o) ([raw file](https://github.com/hypar-io/Elements/blob/cdb788d6c6c269881dc1bbfd2b775589873321ac/Elements/src/Geometry/Solids/Extrude.cs#L84)):*
> ```Smalltalk
>         /// <param name="height"></param>
>         /// <param name="direction"></param>
>         /// <param name="isVoid"></param>
> ```

Add `/// <param name="flipped"></param>` for documentation... Also do we not need to call it `@flipped` ? I don't know the exact reasoning for the `@` symbol in all these cases across the code.

andrewheumann (commented):
This is what I would expect for an "inside out extrude with voids." Looks a little funny, but I think the alternative would too! 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/940#-:-NOE_dcGETtMHtvrRfGB:b-yd2y4b)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer)

___
*[`Elements/src/Geometry/Solids/Extrude.cs` line 84 at r1](https://reviewable.io/reviews/hypar-io/Elements/940#-NOCHN9GAOXuD0xEOaIB:-NOEYxfV0f4nXKk-MxNg:bz2g6p1) ([raw file](https://github.com/hypar-io/Elements/blob/cdb788d6c6c269881dc1bbfd2b775589873321ac/Elements/src/Geometry/Solids/Extrude.cs#L84)):*

Add `/// <param name="flipped"></param>` for documentation... Also do we not need to call it `@flipped` ? I don't know the exact reasoning for the `@` symbol in all these cases across the code.
</blockquote></details>

The @sign is a legacy from when we codegenned ALL our classes, even the ones built in to elements. The @ symbol MOSTLY does nothing, but it helps protect us if a variable name from the JSON is a reserved word in C#, like `class`. If you did `Polygon class`, you'd get a compiler error, but `Polygon @class` builds. Still, it really doesn't serve us anymore — I've gotten rid of the others here, and added xml comments.

anthonie-kramer (changes requested):
Agreed, :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/940#-:-NOFvuEx2fBYMHNxEYo6:b-6k8pc1)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Solids/Extrude.cs` line 84 at r1](https://reviewable.io/reviews/hypar-io/Elements/940#-NOCHN9GAOXuD0xEOaIB:-NOFw1eB58dAA8MeajlN:b-6w3e8n) ([raw file](https://github.com/hypar-io/Elements/blob/cdb788d6c6c269881dc1bbfd2b775589873321ac/Elements/src/Geometry/Solids/Extrude.cs#L84)):*

The @sign is a legacy from when we codegenned ALL our classes, even the ones built in to elements. The @ symbol MOSTLY does nothing, but it helps protect us if a variable name from the JSON is a reserved word in C#, like `class`. If you did `Polygon class`, you'd get a compiler error, but `Polygon @class` builds. Still, it really doesn't serve us anymore — I've gotten rid of the others here, and added xml comments.
</blockquote></details>

Sounds good! Thanks for the context.

anthonie-kramer (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/940)* status: :shipit: complete! 1 of 1 approvals obtained

PR #941, opened 2023-02-09
AdaptiveGrid. Fix tolerances and vertices loopkup.

DmytroMuravskyi:
BACKGROUND:
- During my work with AdaptiveGrid I encountered weird errors time to time: two vertices were inserted into the same spot, or I was not able to retrieve the vertex that is there by it approximate coordinates within the tolerance.  

DESCRIPTION:
- AdaptiveGrid.AddCutEdge: provided missing tolerance parameter PointOnLine. It's necessary since non default tolerance is used in AdaptiveGrid. 
- Define AdaptiveGrid.Tolerance as distance tolerance. Set coordinate tolerance as half of tolerance, so sum of deviations is still less than twice the tolerance.
- Rework vertices lookup of AdaptiveGrid. Don't just check the first suitable X/Y coordinate. Go over each suitable coordinate branch until good vertex is found. The complexity stays the same but function is no longer error prone.

TESTING:
- Added ClipperTest file that shows data modification tolerances.
- Added AdaptiveGridFindsCorrectVertex test that shows weak spot of previous lookup implementation.
- All existing tests are passing.
  
FUTURE WORK:
- I'm considering not using Grid2D for splitting Polygon by a set points. The problem is simple enough and not really require boolean operation. This way I'll eliminate input data modifications made by Clipper.

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - `Message.FromLine` is set obsolete.
 - In `AdaptiveGridRouting`, if there are several connection points with the same cost - choose one that is closer to the trunk.
 - GLTF writing now includes an ad-hoc `HYPAR_info` extension which aids in mapping between GLTF content and element ids in the model.
+- `AdaptiveGrid.Tolerance` is not distance tolerance. Half the tolerance is used for individual coordinates snapping inside the grid.
 
 ### Fixed
 
@@ -42,6 +43,7 @@
 - Fixed an issue where `Grid2d.GetCells()` multiple times could fail to return the correct results on subsequent calls, because changes to the axis grids were not invalidating the grid's computed cells.
 - Adding the first vertex to a mesh with `merge: true` would throw an exception, this is fixed.
 - Handle quotes in string literals for content catalog code generation by doubling them up.
+- Fix `AdaptiveGrid.TryGetVertexIndex` returning `false` for existing vertex if other vertex has similar X or Y coordinate.
 
 ## 1.3.0
 
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -61,9 +61,8 @@ namespace Elements.Spatial.AdaptiveGrid
         #region Properties
 
         /// <summary>
-        /// Tolerance for points being considered the same.
-        /// Applies individually to X, Y, and Z coordinates, not the cumulative difference!
-        /// Tolerance is twice the epsilon to make sure graph has no cracks when new sections are added.
+        /// Distance tolerance for points being considered the same.
+        /// Tolerance is twice the epsilon because gird uses single tolerance for individual coordinates snapping.
         /// </summary>
         public double Tolerance { get; } = Vector3.EPSILON * 2;
 
@@ -357,13 +356,13 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <returns>True if any Vertex is close enough.</returns>
         public bool TryGetVertexIndex(Vector3 point, out ulong id)
         {
-            var zDict = GetAddressParent(_verticesLookup, point, tolerance: Tolerance);
+            var zDict = GetAddressParent(_verticesLookup, point, tolerance: Tolerance / 2);
             if (zDict == null)
             {
                 id = 0;
                 return false;
             }
-            return TryGetValue(zDict, point.Z, out id, Tolerance);
+            return TryGetValue(zDict, point.Z, out id, Tolerance / 2);
         }
 
         /// <summary>
@@ -395,7 +394,7 @@ namespace Elements.Spatial.AdaptiveGrid
         {
             if (!TryGetVertexIndex(point, out var id))
             {
-                var zDict = GetAddressParent(_verticesLookup, point, true, Tolerance);
+                var zDict = GetAddressParent(_verticesLookup, point, true, Tolerance / 2);
                 id = this._vertexId;
                 var vertex = new Vertex(id, point);
                 zDict[point.Z] = id;
@@ -1295,14 +1294,14 @@ namespace Elements.Spatial.AdaptiveGrid
         {
             var vertex = _vertices[id];
             _vertices.Remove(id);
-            var zDict = GetAddressParent(_verticesLookup, vertex.Point, tolerance: Tolerance);
+            var zDict = GetAddressParent(_verticesLookup, vertex.Point, tolerance: Tolerance / 2);
             if (zDict == null)
             {
                 return;
             }
             zDict.Remove(vertex.Point.Z);
 
-            TryGetValue(_verticesLookup, vertex.Point.X, out var yzDict, Tolerance);
+            TryGetValue(_verticesLookup, vertex.Point.X, out var yzDict, Tolerance / 2);
             if (zDict.Count == 0)
             {
                 yzDict.Remove(vertex.Point.Y);
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -908,10 +908,10 @@ namespace Elements.Spatial.AdaptiveGrid
                 }
                 else if (oldEdgeLine.Direction().IsParallelTo(newEdgeLine.Direction()))
                 {
-                    var isNewEdgeStartOnOldEdge = oldEdgeLine.PointOnLine(newEdgeLine.Start);
-                    var isNewEdgeEndOnOldEdge = oldEdgeLine.PointOnLine(newEdgeLine.End);
-                    var isOldEdgeStartOnNewEdge = newEdgeLine.PointOnLine(oldEdgeLine.Start, true);
-                    var isOldEdgeEndOnNewEdge = newEdgeLine.PointOnLine(oldEdgeLine.End, true);
+                    var isNewEdgeStartOnOldEdge = oldEdgeLine.PointOnLine(newEdgeLine.Start, false, Tolerance);
+                    var isNewEdgeEndOnOldEdge = oldEdgeLine.PointOnLine(newEdgeLine.End, false, Tolerance);
+                    var isOldEdgeStartOnNewEdge = newEdgeLine.PointOnLine(oldEdgeLine.Start, true, Tolerance);
+                    var isOldEdgeEndOnNewEdge = newEdgeLine.PointOnLine(oldEdgeLine.End, true, Tolerance);
                     // new edge is inside old edge
                     if (isNewEdgeStartOnOldEdge && isNewEdgeEndOnOldEdge &&
                         AddEdgeInsideExisting(edgeV0, edgeV1, startVertex, endVertex))
--- a/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
+++ b/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs
@@ -356,13 +356,8 @@ namespace Elements.Spatial.AdaptiveGrid
         /// <returns>True if any Vertex is close enough.</returns>
         public bool TryGetVertexIndex(Vector3 point, out ulong id)
         {
-            var zDict = GetAddressParent(_verticesLookup, point, tolerance: Tolerance / 2);
-            if (zDict == null)
-            {
-                id = 0;
-                return false;
-            }
-            return TryGetValue(zDict, point.Z, out id, Tolerance / 2);
+            id = GetVertexFromDictionary(point, out _, out _, tolerance: Tolerance / 2);
+            return id != 0;
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 65 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrkmawCHz7nlysjXxA:-NNrkmawCHz7nlysjXxB:bjkvygr) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L65)):*
> ```Smalltalk
>         /// <summary>
>         /// Distance tolerance for points being considered the same.
>         /// Tolerance is twice the epsilon because gird uses single tolerance for individual coordinates snapping.
> ```

spelling "grid"
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1444 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrlIF-4XPfwX3b0mEY:-NNrlIF-4XPfwX3b0mEZ:bcg6hsr) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1444)):*
> ```Smalltalk
>         }
> 
>         private static IEnumerable<T> GetValue<T>(Dictionary<double, T> dict, double key, double? tolerance = null)
> ```

if this returns an IEnumerable it should be called GetValues with an 's'
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1463 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrmIHABEzANzhnftQS:-NNrmIHABEzANzhnftQT:b-725jie) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1463)):*
> ```Smalltalk
>         }
> 
>         private ulong GetVertexFromDictionary(Vector3 point,
> ```

here and the next methods that are To/FromDictionary I don't think we need to specify "FromDictionary"  There is an internal dictionary that is used to store these things, but that's not important to the method is it?
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1471 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrmglU4CM3pErRQ65t:-NNrmglU4CM3pErRQ65u:b-c3lzsu) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1471)):*
> ```Smalltalk
>             zDict = null;
> 
>             foreach (var yz in GetValue(_verticesLookup, point.X, tolerance))
> ```

can we rename some of these dictionary variables?  _verticesLookup is kind of a xyzDict, but so we could rename it to that, or maybe go with xyzLookup, yzLookup, and zLookup rather than calling them dict?  I think that would be more consistent.
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1511 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrlz5A2U1UD2hLnYEt:-NNrlz5A2U1UD2hLnYEu:bb0tz9k) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1511)):*
> ```Smalltalk
>             if (id == 0 || id != vertex.Id || zDict == null || yzDict == null)
>             {
>                 throw new Exception("Vertex can't be removed. Coordinate dictionary is broken.");
> ```

This warning seems not quite right, if id==0 it could also be that there's simply no vertex at that point right?
___
*[`Elements/test/ClipperTest.cs` line 11 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrnGcUAzv2AmgVGcnq:-NNrnGcUAzv2AmgVGcnr:b7rh1t3) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L11)):*
> ```Smalltalk
> {
>     public class ClipperTest
>     {
> ```

can we get a comment description of why we have this test?
___
*[`Elements/test/ClipperTest.cs` line 13 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrnAlE0ma1T-a7cFJc:-NNrnAlE0ma1T-a7cFJd:b-8zgd8l) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L13)):*
> ```Smalltalk
>     {
>         [Fact]
>         public void CliperChangesInputNumbers()
> ```

spelling, two 'p' in Clipper
___
*[`Elements/test/ClipperTest.cs` line 19 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrnQHV8bis4tWwe4LX:-NNrnQHV8bis4tWwe4LY:b-bu72gc) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L19)):*
> ```Smalltalk
>             var random = new Random();
>             var z = 1.23456789;
>             random.NextDouble();
> ```

I don't think we should use random in our tests unless it's required.  Could we use a fixed list of values?
___
*[`Elements/test/ClipperTest.cs` line 32 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrnj1wBKqyWQGNw-aB:-NNrnj1wBKqyWQGNw-aC:b-vzsggs) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L32)):*
> ```Smalltalk
>             polygon = polygon.TransformedPolygon(new Transform().Rotated(Vector3.ZAxis, 25));
> 
>             var tolerance = Vector3.EPSILON * 2;
> ```

describe the significance of the *2 for the epsilon here.
___
*[`Elements/test/ClipperTest.cs` line 47 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNroJvP2pPou-5nUYSg:-NNroJvP2pPou-5nUYSh:b9jbnzg) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L47)):*
> ```Smalltalk
>                 Assert.True(changedPolygon.Vertices[i].Z == 0d);
> 
>                 var v2d = new Vector3(polygon.Vertices[i].X, polygon.Vertices[i].Y);
> ```

why make a new Vector3 from the Vertex?  we compare it to a polygon vertex below, are they different types somehow?

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 65 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrkmawCHz7nlysjXxA:-NNuPm7p7a2YfrpPHfqu:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L65)):*

spelling "grid"
</blockquote></details>

Done.
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1444 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrlIF-4XPfwX3b0mEY:-NNuPmWR2og8kqM0goFK:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1444)):*

if this returns an IEnumerable it should be called GetValues with an 's'
</blockquote></details>

Done.
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1463 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrmIHABEzANzhnftQS:-NNuPyuV12rg-Jgdt9Re:b1piusx) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1463)):*

here and the next methods that are To/FromDictionary I don't think we need to specify "FromDictionary"  There is an internal dictionary that is used to store these things, but that's not important to the method is it?
</blockquote></details>

I don't want it to be just another AddVertex/GetVertex/RemoveVertex as it's doing only part of the job. Maybe instead of abstract "dictionary"  Use AddToXYZLookup / GetFromXYZLookup / RemoveFromXYZLookup?
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1471 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrmglU4CM3pErRQ65t:-NNuYK2U9VplSJkaczPz:bilq45q) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1471)):*

can we rename some of these dictionary variables?  _verticesLookup is kind of a xyzDict, but so we could rename it to that, or maybe go with xyzLookup, yzLookup, and zLookup rather than calling them dict?  I think that would be more consistent.
</blockquote></details>

Since I can use existing variable in foreach, I need two names per variable. I ended up with:

_verticesLookup  -> _xyzLookup
yzLookup and zLookup for parameter names
yzToId and zToId for internal names
___
*[`Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs` line 1511 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNrlz5A2U1UD2hLnYEt:-NNuRRp685Xlx2Qo_W-I:b-rg91w4) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/src/Spatial/AdaptiveGrid/AdaptiveGrid.cs#L1511)):*

This warning seems not quite right, if id==0 it could also be that there's simply no vertex at that point right?
</blockquote></details>

At this point we want to remove one of existing vertices. If the vertex is not part of the grid, we will fail much earlier. Removing from lookup is like cleanup, so if we can't something is really wrong happened.



_Code quote:_
```Smalltalk
zDict == null || yzDic
```
___
*[`Elements/test/ClipperTest.cs` line 47 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNroJvP2pPou-5nUYSg:-NNu_v0N0WomEgVJgg3G:brcw78b) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L47)):*

why make a new Vector3 from the Vertex?  we compare it to a polygon vertex below, are they different types somehow?
</blockquote></details>

Clipper works only with 2D points so it's Z is always 0.  I created a 2D point to calculate distance.

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/test/ClipperTest.cs` line 47 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNroJvP2pPou-5nUYSg:-NNvd-J34AsOZKOO8uzy:b-8z92mg) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L47)):*

but if the point is already 2D do you need to make a new 2D point?
</blockquote></details>

Original points are with 1.23456789 elevation to show that it's lost after we go back



_Code quote:_
```Smalltalk
1.23456789;
```

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/test/ClipperTest.cs` line 47 at r1](https://reviewable.io/reviews/hypar-io/Elements/941#-NNroJvP2pPou-5nUYSg:-NNvoBJV9FqGwxrhH9tf:b-frifqh) ([raw file](https://github.com/hypar-io/Elements/blob/b8b783063bcd1d6703e1494f5cff6ba2e1fc11d2/Elements/test/ClipperTest.cs#L47)):*

Original points are with 1.23456789 elevation to show that it's lost after we go back
</blockquote></details>

oh derp....  thanks!

wynged (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/941#-:-NNwUYGB1JEu7z6Is3yy:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @DmytroMuravskyi)

PR #948, opened 2023-02-27
add Mappings to Elements and new RevitFamilyPointInstance mapping

wynged:
BACKGROUND:
- We would like to start storing a variety of optional "MappingData" in models, and store references to that mapping data on Elements.  Only Revit mappings are currently supported, but this is being design to support other mappings in the future.
- designed this so that the mapping dictionary is a bit protected, preventing people from adding arbitrary mapping data that isn't considered supported.  This may not always be the right call but feels right given that users can always add their own mapping data to AdditionalProperties if they aren't using features that are built into elements

DESCRIPTION:
- The MappingBase class which inherits from Elements and the first supported subclass "RevitFamilyMapping" 
- The internal mapping dictionary needed to be explicitly serialized as Elements so they would be serialized by reference.
- Change the validation logic for when we are serializing a "Top Level" element, which needs to be added directly to the JSON vs an element that is nested inside.  Previously we would not serialize dictionaries correctly because their paths would be multi-part, just like the top level dictionary.  We have not improved, but would still have duplicate serialization issues if an Element had a property called "Elements" that was also a dictionary with GUID keys.
- Ensure that Mapping data is serialized only once, and that it de-serializes correctly.

TESTING:
- A test demonstrates the desired serial/deserial behavior.
  
FUTURE WORK:
- More mapping types as we expand this generic support.

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)


andrewheumann (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/MappingBase.cs` line 18 at r1](https://reviewable.io/reviews/hypar-io/Elements/948#-NPNINmd-v4cCxlJ-cy4:-NPNINmd-v4cCxlJ-cy5:b5rv1yn) ([raw file](https://github.com/hypar-io/Elements/blob/3cff5f08138fa608c06dbb382c204727afd73625/Elements/src/MappingBase.cs#L18)):*
> ```Smalltalk
>         /// </summary>
>         [System.Runtime.Serialization.EnumMember(Value = @"Revit")]
>         Revit
> ```

per previous comment — could we keep Revit-specificity out here, by making an `IMappingContext` interface, which a separate library could implement, instead of a hard-coded enum? If I want to build mappings to my application "WonderModeler," I should be able to do that without submitting a pull request to `Elements` — I should be able to just define a `MappingContext` which implements `IMappingContext`, and a `WonderModelerElementMapping` which inherits from `MappingBase`. If there is some reason why this is extra hard or annoying, happy to revisit, but my preference would be to have anything revit specific live in `Hypar.Revit`. 
___
*[`Elements/src/RevitFamilyPointInstanceMapping.cs` line 6 at r1](https://reviewable.io/reviews/hypar-io/Elements/948#-NPNI2Pw9d7nX3_9Trub:-NPNI2Pw9d7nX3_9Truc:b-ghi9x1) ([raw file](https://github.com/hypar-io/Elements/blob/3cff5f08138fa608c06dbb382c204727afd73625/Elements/src/RevitFamilyPointInstanceMapping.cs#L6)):*
> ```Smalltalk
>     /// The mapping class for Revit
>     /// </summary>
>     public class RevitFamilyPointInstanceMapping : MappingBase
> ```

Does this have to live in Elements? I feel like there shouldn't be anything `Revit`-specific in elements — this would be the first such instance.

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/MappingBase.cs` line 18 at r1](https://reviewable.io/reviews/hypar-io/Elements/948#-NPNINmd-v4cCxlJ-cy4:-NPNWYjA6njVlO4Oe9sV:byx2vpb) ([raw file](https://github.com/hypar-io/Elements/blob/3cff5f08138fa608c06dbb382c204727afd73625/Elements/src/MappingBase.cs#L18)):*

per previous comment — could we keep Revit-specificity out here, by making an `IMappingContext` interface, which a separate library could implement, instead of a hard-coded enum? If I want to build mappings to my application "WonderModeler," I should be able to do that without submitting a pull request to `Elements` — I should be able to just define a `MappingContext` which implements `IMappingContext`, and a `WonderModelerElementMapping` which inherits from `MappingBase`. If there is some reason why this is extra hard or annoying, happy to revisit, but my preference would be to have anything revit specific live in `Hypar.Revit`. 
</blockquote></details>

I think that should totally be possible.  any reason not to just use a string for the dictionary?  Does the IMappingcontext give us antying?

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/MappingBase.cs` line 18 at r1](https://reviewable.io/reviews/hypar-io/Elements/948#-NPNINmd-v4cCxlJ-cy4:-NPO_hCd5lRIUGzCKqSF:b-6zb3sq) ([raw file](https://github.com/hypar-io/Elements/blob/3cff5f08138fa608c06dbb382c204727afd73625/Elements/src/MappingBase.cs#L18)):*

I think that should totally be possible.  any reason not to just use a string for the dictionary?  Does the IMappingcontext give us antying?
</blockquote></details>

using elements as keys proved to be tricky, just doing strings for now.
___
*[`Elements/src/RevitFamilyPointInstanceMapping.cs` line 6 at r1](https://reviewable.io/reviews/hypar-io/Elements/948#-NPNI2Pw9d7nX3_9Trub:-NPO_oBo9-97tFwPsdpM:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/3cff5f08138fa608c06dbb382c204727afd73625/Elements/src/RevitFamilyPointInstanceMapping.cs#L6)):*

Does this have to live in Elements? I feel like there shouldn't be anything `Revit`-specific in elements — this would be the first such instance. 
</blockquote></details>

Done.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Element.cs` line 95 at r2](https://reviewable.io/reviews/hypar-io/Elements/948#-NPPNc-E2epsPf5HHBXY:-NPPNc-E2epsPf5HHBXZ:bbu7ylf) ([raw file](https://github.com/hypar-io/Elements/blob/843bb7f5ccae9e5af03bcd6d8bb6cfb612e3a7e6/Elements/src/Element.cs#L95)):*
> ```Smalltalk
>         /// An optional dictionary of mappings.
>         /// </summary>
>         [JsonProperty("Mappings", Required = Required.Default)]
> ```

Let's make sure this doesn't show up in the JSON if it's null, and add a test to verify that. 

I don't love that this is `_mappings` but annotated to serialize to `Mappings`. My linter says this is a naming violation — can it just be `Mappings`, even if it's internal? 
___
*[`Elements/src/MappingBase.cs` line 4 at r2](https://reviewable.io/reviews/hypar-io/Elements/948#-NPPLLfaAkL8p9NAYWW3:-NPPLLfaAkL8p9NAYWW4:b7sh3j6) ([raw file](https://github.com/hypar-io/Elements/blob/843bb7f5ccae9e5af03bcd6d8bb6cfb612e3a7e6/Elements/src/MappingBase.cs#L4)):*
> ```Smalltalk
> {
>     /// <summary>
>     /// The base class for all mapping classes.
> ```

Provide a little more detail here. "A mapping specifies additional data utilized to translate an element to a different application context or platform." (or something)

wynged (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Element.cs` line 95 at r2](https://reviewable.io/reviews/hypar-io/Elements/948#-NPPNc-E2epsPf5HHBXY:-NPSULjy-Pj2nzcRlhwq:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/843bb7f5ccae9e5af03bcd6d8bb6cfb612e3a7e6/Elements/src/Element.cs#L95)):*

Let's make sure this doesn't show up in the JSON if it's null, and add a test to verify that. 

I don't love that this is `_mappings` but annotated to serialize to `Mappings`. My linter says this is a naming violation — can it just be `Mappings`, even if it's internal? 
</blockquote></details>

Done.
___
*[`Elements/src/MappingBase.cs` line 4 at r2](https://reviewable.io/reviews/hypar-io/Elements/948#-NPPLLfaAkL8p9NAYWW3:-NPST63k14mt21euh0rm:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/843bb7f5ccae9e5af03bcd6d8bb6cfb612e3a7e6/Elements/src/MappingBase.cs#L4)):*

Provide a little more detail here. "A mapping specifies additional data utilized to translate an element to a different application context or platform." (or something)
</blockquote></details>

Done.

andrewheumann (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/948)* status: :shipit: complete! 1 of 1 approvals obtained

PR #949, opened 2023-03-08
Add "Level" property to floor constructor

andrewheumann:
BACKGROUND:
- I need to add a `Level` property to the floor schema, so that floors can have relationships to levels.
- This broke codegen code for types which inherit from floor, like `CirculationSegment`, because the constructor positional arguments were wrong. 

DESCRIPTION:
- Adds an additional constructor with arguments that match the ones in the updated schema.

TESTING:
- Haven't tested. If we merge and it's still broken, I will own cleanup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/949)

### DIFF
--- a/Elements/src/Floor.cs
+++ b/Elements/src/Floor.cs
@@ -56,7 +56,7 @@ namespace Elements
         /// <param name="name">The floor's name.</param>
         public Floor(Profile profile,
                      double thickness,
-                     Guid? level = null,
+                     Guid? level,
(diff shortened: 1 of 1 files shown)


serenayl (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/949)* status: :shipit: complete! 1 of 1 approvals obtained

PR #956, opened 2023-03-31
Updated Polygon.Rectangle function to enforce flat polygon geometry.

jamesbradleym:
BACKGROUND:
Running Polygon.Rectangle() with vectors that don't share the same Z coordinate leads to the creation of non-flat geometry.

DESCRIPTION:
Enforces flat geometry creation by projecting all points to min.Z when floor = true 
- or -
Creates pitched rectangle with point b sharing Z with max and point d sharing Z with min

TESTING:
Resultant Rectangles created with vectors not sharing the same Z should be flat
  
FUTURE WORK:
Future work may include solving for other points more intelligently to what user may anticipate when z units differ, not essential as other modeling software tends to project resulting rectangle onto XY plane similarly to default behavior.

REQUIRED:

COMMENTS:
- Any other notes.

`var rectangle = Polygon.Rectangle(new Vector3(0, 0, 1), new Vector3(1, 1, 1));
            output.Model.AddElement(rectangle);`
Screenshot 2023-03-30 at 4 38 05 PM

`var rectangle = Polygon.Rectangle(new Vector3(0, 0, 0), new Vector3(1, 1, 1));
            output.Model.AddElement(rectangle);`

Screenshot 2023-03-30 at 4 38 51 PM

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)


andrewheumann (changes requested):
My sense is that there isn't a foolproof default behavior here. I am not sure forcing to min z is especially logical — if a function author is supplying two points at different elevations, thinking they were creating a vertical rectangle, they might be surprised / confused by this behavior. 

Happy for pushback, but my intuition is:
- If the two Zs are within `Vector3.EPSILON` (our global tolerance) of each other, it's probably OK to use just min.Z.
- Otherwise, we throw an `ArgumentException` which informs the user that their inputs are invalid, and suggests a remediation. 
- Possible remediations we could implement would be:
  1. Include an optional `Plane plane` parameter, and if present, project the points to the Plane.
  2. Include an optional `Vector3? normal` parameter, and if present, project the points to the same parameter along the normal (an implicit plane, basically.)
  3. Include an optional third point such that the third point establishes a plane (another version of implicit plane)

I am inclined towards option 1 — it seems natural to want to be able to create a rectangle from two points in a plane. This is the Rhino API for similar: https://developer.rhino3d.com/docify/api/rhinocommon/rhino.geometry.rectangle3d 
Their classes all explicitly differentiate between `2d` and `3d` in their naming convention, which is something we've talked about for Elements as well.

It would be good to know what case caused you to pursue this — what was in your mind when you accidentally tried to create a rectangle w/ two different z coordinates? what did you expect should happen? Is this mostly a mitigation for "two zs that are ever so slightly off," or were you actually trying to build a rectangle with a different orientation? 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-NSVUwwN61g7Kz9mm8L6:b-k5slss)* status: 1 change requests, 0 of 1 approvals obtained

jamesbradleym:
Considering this more I agree having floor as the default behavior would be surprising when creating an intentionally Z-differentiated rectangle. I've gone ahead and run over a few more options that from my perspective behave more in line with expectations.

Polygon Rectangle(double width, double height, Plane plane = null)
When a plane is provided the resulting rectangle is a projection of the original points with a provided width and height.
When a plane is not provided the resulting rectangle is drawn on the xy plane, centered at (0,0,0) with a provided width and height.

Polygon Rectangle(Vector3 min, Vector3 max, Plane plane = null)
When a plane is provided the resulting rectangle is a projection of the original points with naively found edges perpendicular to the vector between min and max. This will always result in a rectangle of equal width and height (square).
When a plane is not provided the resulting rectangle is drawn oriented to a naive reference, currently set as one of the Vector.Axis values. This results in a square rectangle as well but is more in line with my expectations of a rectangle drawn in 3D space with only 2 points provided.

I also looked into another method that takes into consideration a prevailing axis (defined as a guide for an edge to follow when constructing the rectangle). This results in a similar rectangle as other modeling software, namely Rhino's Rectangle3D. It seems when creating a Plane object rhino records the Axes of the plane based on the originally(or updated) used points. If we had a similar method to track down the plane's axes we could also implement this method.

andrewheumann (commented):
I buy the first one (`Polygon.Rectangle(double width, double height, Plane plane = null)`), and I see the challenge with the second one — a plane and two points are insufficient information. I don't think we should try to force this — an always-square projection seems useless to me. 

Our `Plane`s don't have "axes" the way rhino's do — they are just a position and a normal. Instead, when we want to represent a "coordinate frame", we use a `Transform`. So maybe that's a way forward.

I think your suggestion about an alignment axis makes sense, but I also think we can infer a sensible one in most cases. Here's what I'd propose:
1. a version that utilizes a transform, and always projects to the transform's plane
```cs
Polygon Rectangle(Vector3 min, Vector3 max, Transform transform = null) {
    transform ??= new Transform();
    var inverse = transform.Inverted();
    var a = inverse.OfPoint(min);
    a.Z = 0;
    var c = inverse.OfPoint(max);
    c.Z = 0;
    var b = ((c.X, a.Y));
    var d = ((a.X, c.Y));
    return new Polygon(new []{a, b, c, d}).TransformedPolygon(transform);
}
```
2. One that uses a plane and an optional alignment vector. In most architectural cases, if you're drawing a rectangle that's not *in* the XY plane, one of its sides wants to be parallel to the XY plane. 
```cs

Polygon Rectangle(Vector3 min, Vector3 max, Plane plane, Vector3? alignment = null) {
    alignment ??= plane.Normal.IsParallelTo(Vector3.ZAxis) ? Vector3.XAxis : Vector3.ZAxis;
    var transform = new Transform(plane.Origin, alignment.Value, plane.Normal);
    return Rectangle(min, max, transform);
}
```

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-NSgmu_KEKuL97pCtwLs:b-442cx8)* status: 1 change requests, 0 of 1 approvals obtained

jamesbradleym:
It doesn't feel intuitive to me that if a min/max with varied Z values are provided that the output is flat in the Z axis, but ultimately I agree it shouldn't be something that is forced and this projection is a better outcome than bizarre non-Z plane weirdness. That being said it feels even less intuitive that inputs with Z values that are not 0 would be projected to 0 outright?

Looking at the results of `Rectangle(Vector3 min, Vector3 max, Transform transform = null)` it feels like there's a detachment between expectation and result. Maybe I'm reading this wrong but it looks like it projects to the XY plane, not the transform's, and then afterwards is transformed a la `transform`. If someone is resulting to providing a plane to construct a rectangle it seems reasonable to assume the plane takes precedence in consideration for constructing that rectangle?

andrewheumann (commented):
I think you're misreading the method which uses a transform. It is projecting the provided points to the plane described by the transform. Audio got cut off at the beginning, but this video hopefully helps clarify how it behaves: 
[Screen Recording 2023-05-03 at 9.05.21 AM.mp4](https://files.reviewableusercontent.io/images/reviewable/31935763/YtMgrgalGeiti_sakdPpq5qx/Screen%20Recording%202023-05-03%20at%209.05.21%20AM.mp4)

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-NUWiOxm2yJJr5HlVoKj:b-w2q0wh)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @jamesbradleym)

___
*[`Elements/src/Geometry/Polygons.cs` line 21 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFHGp1ytroOuwtI6E:-NUgF2Dg5KazOlqZNqqV:bhzmu7v) ([raw file](https://github.com/hypar-io/Elements/blob/1ae61fa5d23e4bd8ec640a116ed43ec30170af38/Elements/src/Geometry/Polygons.cs#L21)):*
> ```Smalltalk
>         {
>             Vector3 a, b, c, d;
>             if (plane != null && plane.Normal.Dot(Vector3.ZAxis) != 1)
> ```

If the normal is anti-parallel to the Z Axis, this method will fail. Check the absolute value of the dot product, and use a small amount of tolerance to account for floating point error. `plane.Normal.Dot(Vector3.ZAxis) != 1` => `Math.Abs(plane.Normal.Dot(Vector3.ZAxis)) > 1- Vector3.EPSILON` is probably what I would do 
___
*[`Elements/src/Geometry/Polygons.cs` line 23 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgF8jZ2FcMR4OQAhPq:-NUgF8jZ2FcMR4OQAhPr:b-anw7z0) ([raw file](https://github.com/hypar-io/Elements/blob/1ae61fa5d23e4bd8ec640a116ed43ec30170af38/Elements/src/Geometry/Polygons.cs#L23)):*
> ```Smalltalk
>             if (plane != null && plane.Normal.Dot(Vector3.ZAxis) != 1)
>             {
>                 Vector3 origin = plane.ClosestPoint(Vector3.Origin);
> ```

Why the closest point — why not the plane's own origin?
___
*[`Elements/src/Geometry/Polygons.cs` line 25 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFnIo9Qtvj0ZOvyJJ:-NUgFnIo9Qtvj0ZOvyJK:byhu82k) ([raw file](https://github.com/hypar-io/Elements/blob/1ae61fa5d23e4bd8ec640a116ed43ec30170af38/Elements/src/Geometry/Polygons.cs#L25)):*
> ```Smalltalk
>                 Vector3 origin = plane.ClosestPoint(Vector3.Origin);
>                 // calculate Vector3 for each component based on plane normal
>                 Vector3 XZ = plane.Normal.Cross(Vector3.ZAxis).Unitized() * width / 2;
> ```

I would call these something else — XZ/YZ makes it sound like we're talking about the world XZ/YZ planes, when we're talking about a vector along the width / height of the rectangle
___
*[`Elements/src/Geometry/Polygons.cs` line 62 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgEUxwDvb8y2U7MQAf:-NUgEUxwDvb8y2U7MQAg:b-jb0f9p) ([raw file](https://github.com/hypar-io/Elements/blob/1ae61fa5d23e4bd8ec640a116ed43ec30170af38/Elements/src/Geometry/Polygons.cs#L62)):*
> ```Smalltalk
>             var c = inverse.OfPoint(max);
>             c.Z = 0;
>             var b = ((c.X, a.Y));
> ```

are these double parens necessary? (Sorry, I realize those are from my code snippet, but I don't think they're needed)

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Polygons.cs` line 21 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFHGp1ytroOuwtI6E:-NVgrYmCCblGsekRtuaU:b-960zli) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L21)):*

If the normal is anti-parallel to the Z Axis, this method will fail. Check the absolute value of the dot product, and use a small amount of tolerance to account for floating point error. `plane.Normal.Dot(Vector3.ZAxis) != 1` => `Math.Abs(plane.Normal.Dot(Vector3.ZAxis)) > 1- Vector3.EPSILON` is probably what I would do 
</blockquote></details>

100%, didn't catch this in the first go around. Had trouble finding info on lamba in if statements so I re-wrote this as 



_Code snippet:_
```Smalltalk
// Confirm the provided plane is valid and is not either Z-Normal or would produce a valid rect from U or V
// any abs less than 1-Epsilon would be a valid normal
if (plane != null && Math.Abs(plane.Normal.Dot(Vector3.ZAxis)) > (1 - Vector3.EPSILON))
{ ... }
```
___
*[`Elements/src/Geometry/Polygons.cs` line 23 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgF8jZ2FcMR4OQAhPq:-NVgrTPN7RhDxzPQU4Qu:blqxzmq) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L23)):*

Why the closest point — why not the plane's own origin?
</blockquote></details>

No reason, makes sense to use the plane.origin as it exists - implemented.
___
*[`Elements/src/Geometry/Polygons.cs` line 25 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFnIo9Qtvj0ZOvyJJ:-NVgrJzA9n25OAhgtjky:b-j1y72u) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L25)):*

I would call these something else — XZ/YZ makes it sound like we're talking about the world XZ/YZ planes, when we're talking about a vector along the width / height of the rectangle
</blockquote></details>

Switched over to a more conventional U/V syntax which I think is more appropriate here.
___
*[`Elements/src/Geometry/Polygons.cs` line 62 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgEUxwDvb8y2U7MQAf:-NVgrEzV4SpaIWFoCLs6:bzi1hn4) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L62)):*

are these double parens necessary? (Sorry, I realize those are from my code snippet, but I don't think they're needed)
</blockquote></details>

Nope, removed!

jamesbradleym:
Additionally added some clarification to the returns xml for 

`/// <returns>A rectangular Polygon with its lower left corner projected from min and its upper right corner projected from max onto a plane with a sided alignment.</returns>
Polygon Rectangle(Vector3 min, Vector3 max, Transform transform = null)` 

and 

`/// <returns>A rectangular Polygon with its lower left corner projected from min and its upper right corner projected from max onto a plane with a sided alignment.</returns>
Polygon Rectangle(Vector3 min, Vector3 max, Plane plane, Vector3? alignment = null)`:

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @jamesbradleym)

___
*[`Elements/src/Geometry/Polygons.cs` line 21 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFHGp1ytroOuwtI6E:-NVpWpbm7H3WusmpU73w:b-ful4rs) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L21)):*
<details><summary><i>Previously, jamesbradleym wrote…</i></summary><blockquote>

100%, didn't catch this in the first go around. Had trouble finding info on lamba in if statements so I re-wrote this as 
</blockquote></details>

Oh apologies for the confusion — I wasn't suggesting a lambda, just using an "arrow" symbol for the before => after
___
*[`Elements/src/Geometry/Polygons.cs` line 25 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFnIo9Qtvj0ZOvyJJ:-NVpWzI0A3bESTWavepy:b-mihccw) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L25)):*
<details><summary><i>Previously, jamesbradleym wrote…</i></summary><blockquote>

Switched over to a more conventional U/V syntax which I think is more appropriate here.
</blockquote></details>

nit — parameters are not capitalized in c# (https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN). use `u` and `v`.

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Polygons.cs` line 21 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFHGp1ytroOuwtI6E:-NVpaJYLDZDpx2hr9SfA:b-bp2ojq) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L21)):*

Oh apologies for the confusion — I wasn't suggesting a lambda, just using an "arrow" symbol for the before => after
</blockquote></details>

Gotcha 👍🏻
___
*[`Elements/src/Geometry/Polygons.cs` line 25 at r4](https://reviewable.io/reviews/hypar-io/Elements/956#-NUgFnIo9Qtvj0ZOvyJJ:-NVpaGY27v-yYj0tBs4s:b-896fja) ([raw file](https://github.com/hypar-io/Elements/blob/894a1e4e09d9c595ce73d847ea32f106e1531a4e/Elements/src/Geometry/Polygons.cs#L25)):*

nit — parameters are not capitalized in c# (https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN). use `u` and `v`.
</blockquote></details>

Done!

ikeough:
Any updates here?

andrewheumann (changes requested):
This code does not compile. 

> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-N_ZR8ao6FC7LZ3nRxRt:bdizmcn)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann (changes requested):
Specifically, it is using newer language features than `Elements` is set up to use, and introduces an ambiguity which means calls to `Polygon.Rectangle(Vector3 a, Vector b)` get mistakenly treated as calls to `Polygon.Rectangle(double a, double b)`. I think we can resolve this ambiguity by making `Polygon Rectangle(Vector3 min, Vector3 max, Transform transform = null)` `public static`.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-N_ZRq1AEjn1hZ-WkW_m:b-egx4n1)* status: 1 change requests, 0 of 1 approvals obtained

jamesbradleym (commented):
Corrected 👍🏻

> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956#-:-N_ZZj-d9r0j-qvJZF4B:b-i9qfed)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/956)* status: :shipit: complete! 1 of 1 approvals obtained

PR #957, opened 2023-04-03
Unbound, bound, and trimmed curves.

ikeough:
BACKGROUND:
- We would like to support `Ellipse` and `EllipticalArc`.
- We would like to support curved surfaces.
- We would like to support a `Polycurve` type that corresponds to `IfcIndexedPolycurve`. 
- We would like to align with certain industry standards, like STEP, to ease compatibility.

DESCRIPTION:

- Introduces `BoundedCurve`
  - Introduces the `IBoundedCurve` interface for bound curves that have a start and end.
  - Makes `Polyline` inherit from `BoundedCurve`
- Introduces `TrimmedCurve`
  - Introduces the `ITrimmedCurve<T>` interface for trimmed curves which have a basis curve and a domain.
  - Makes `Circle` inherit from `Curve`, and makes `Arc` a `TrimmedCurve<Circle>`.
  - Makes the existing `Line` a `TrimmedCurve<InfiniteLine>` (a line segment) and introduces the `InfiniteLine` class which represents an unbound line from a start point along a direction. Optimally, `Line`, should become `LineSegment` and `InfiniteLine` should be `Line`, but the renaming would break all serialized instances of lines. This should be considered in the future.
- Introduces `IConic` for curves which are planar sections of cones like `Circle` and (soon) `Ellipse`.
- Introduces "natural" parameterization.
  - Changes the parameterization of `Circle` to be [0,2PI], and adjusts `Arc` domain accordingly to use a start and end angle.
  - Changes the parameterization of `Line` to be [0,Length]
  - Changes the parameterization of `Polyline` to be [0,length]

TESTING:
- All tests currently pass, but we should visually check results to ensure that behavior is consistent. 
  
FUTURE WORK:
- Bezier should have its domains adjusted for natural parameterization when we have better length support.
- Convenience methods to get parameterization 0->1 should be added like `NormalizedParameterAt(double u)`.

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)


Elements/src/Math/Domain1d.cs
```diff
@@ -124,6 +132,16 @@ public Domain1d[] DivideByCount(int n)
             return results;
         }
 
-
+        /// <summary>
+        /// Calculate the middle of the domain.
+        /// </summary>
+        public double Mid()
+        {
```
andrewheumann, line 139: I think this code is wrong in the non-increasing case. Can't this just be `(this.Min + this.Max) / 2` in all cases?

ikeough: Done.

andrewheumann (changes requested):
This is really nice work! Feels "right." 

Couple minor comments, and a question: since this is a breaking change, does this mean we'd need to go to 2.0 for this change b/c semver? 

> Reviewed 37 of 48 files at r1, 1 of 2 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957#-:-NSVkt4f8mCc64VYMevo:b-sw19y)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer and @ikeough)

___
*[`Elements/src/Units.cs` line 360 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVaDT_8bVPvkA4JGqB:-NSVaDT_8bVPvkA4JGqC:b-nss6jn) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Units.cs#L360)):*
> ```Smalltalk
>         }
>     
>         internal static bool IsParameterBetweenOrAlmostEqualTo(double a, double min, double max)
> ```

Not sure why this method belongs in `Units`. Also, appears to be unused by the code  — remove?
___
*[`Elements/src/Geometry/Arc.cs` line 14 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVbo5T56nrfzuEMJOq:-NSVbo5T56nrfzuEMJOr:bb79qxb) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L14)):*
> ```Smalltalk
>     /// [!code-csharp[Main](../../Elements/test/ArcTests.cs?name=example)]
>     /// </example>
>     public partial class Arc : TrimmedCurve<Circle>, IEquatable<Arc>
> ```

The Arc validator throws if you use a startAngle / endAngle > 360, but doesn't care anything about < -360. Since the code actually seems to tolerate start/end angles outside of this range, should we remove that validation? alternately, we can set it to clamp at the negative end too (but I would hate to lose the ability to do something like `new Arc(5, -45, 45)`)! 
___
*[`Elements/src/Geometry/Arc.cs` line 49 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVaV-AAQha5r7itnAG:-NSVaV-AAQha5r7itnAH:b1jyqh6) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L49)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// Create an circular arc.
> ```

an circular => a circular
___
*[`Elements/src/Geometry/Polyline.cs` line 1416 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVi-HE61CQ5JWLIEP4:-NSVhlBS0AC9E3cGANbU:bbtrb8n) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Polyline.cs#L1416)):*
The fact that we can get away with this — not implement this, since the only callers are `Frames` and `RenderVertices`, which we also override in this class, feels like *maybe* a design flaw? If nothing else, it might be worth a comment in the base class which says something like:
> You don't need to implement this so long as you override `Frames` and `RenderVertices`.

But that makes me nervous too... the minute we add another capability to `BoundedCurve` which relies on `GetSampleParameters`, now all of a sudden we have an issue. IDK, I don't have super strong feelings about this, but as-is it feels vaguely risky. 



_Code quote:_
```Smalltalk
 throw new NotImplementedException();
```
___
*[`Elements/src/Geometry/Interfaces/IBoundedCurve.cs` line 5 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVebC3-TSobYwuQqQe:-NSVebC3-TSobYwuQqQf:b-970bk7) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Interfaces/IBoundedCurve.cs#L5)):*
> ```Smalltalk
>     /// <summary>
>     /// A curve with a start and an end.
>     /// Examples of bounded curces include polylines and bezier curves.
> ```

curces => curves
___
*[`Elements/src/Math/Domain1d.cs` line 37 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVf3Lh7FoGFN8nvdta:-NSVf3Lh7FoGFN8nvdtb:b-kl8g4c) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Math/Domain1d.cs#L37)):*
> ```Smalltalk
>             else
>             {
>                 return value < Min && value > Max;
> ```

good catch!
___
*[`Elements/test/LineTests.cs` line 646 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVkC9tBr2xpirfPt56:-NSVkC9tBr2xpirfPt57:bfcvgnt) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/test/LineTests.cs#L646)):*
> ```Smalltalk
> 
>             var middle = new Vector3(2.5, 2.5, 2.5);
>             Assert.Equal(line.Length()/2, line.GetParameterAt(middle));
> ```

Seems like we missed an automatic lint fix here, this should be 
`line.Length() / 2` with spaces
___
*[`Elements.Serialization.IFC/src/Serialization/IFC/IFCExtensions.cs` line 37 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVkejK1rTcbAJZM8M-:-NSVkejK1rTcbAJZM8M0:bo42mtw) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements.Serialization.IFC/src/Serialization/IFC/IFCExtensions.cs#L37)):*
> ```Smalltalk
>                     var cl = new Line(Vector3.Origin,
>                         solid.ExtrudedDirection.ToVector3(), (IfcLengthMeasure)solid.Depth);
>                     var result = new Beam((Line)cl.Transformed(solidTransform),
> ```

I think this can be `.TransformedLine` to avoid the cast.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @anthonie-kramer)

___
*[`Elements/src/Geometry/Arc.cs` line 14 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVbo5T56nrfzuEMJOq:-NSWd7-nBVcq44ZTuQ0m:bfoct17) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L14)):*

The Arc validator throws if you use a startAngle / endAngle > 360, but doesn't care anything about < -360. Since the code actually seems to tolerate start/end angles outside of this range, should we remove that validation? alternately, we can set it to clamp at the negative end too (but I would hate to lose the ability to do something like `new Arc(5, -45, 45)`)! 
</blockquote></details>

I've changed the validation for arc span to check whether the proposed span is > 360.0, regardless of start /end. LMK what you think.
___
*[`Elements/src/Geometry/Arc.cs` line 49 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVaV-AAQha5r7itnAG:-NSVn6mF8PI1Qiv1is-6:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L49)):*

an circular => a circular
</blockquote></details>

Done.
___
*[`Elements/src/Geometry/Polyline.cs` line 1416 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVi-HE61CQ5JWLIEP4:-NSWC3ig2I8viAGSeJwF:b-vb8kje) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Polyline.cs#L1416)):*

The fact that we can get away with this — not implement this, since the only callers are `Frames` and `RenderVertices`, which we also override in this class, feels like *maybe* a design flaw? If nothing else, it might be worth a comment in the base class which says something like:
> You don't need to implement this so long as you override `Frames` and `RenderVertices`.

But that makes me nervous too... the minute we add another capability to `BoundedCurve` which relies on `GetSampleParameters`, now all of a sudden we have an issue. IDK, I don't have super strong feelings about this, but as-is it feels vaguely risky. 

</blockquote></details>

This is a good catch. I've made `GetSampleParameters` abstract so every curve has to implement it. I've also simplified `RenderVertices()` to have one implementation, which required allowing curve types to specify which GL line mode they should use to render. See the addition of the internal property `IsClosedForRendering` which is set to `false` by default and `true` for polygons so that the line mode switches to `LOOP`. All `RenderVertices` was doing differently for polygons was adding one more vertex to the collection so it would draw as a loop.
___
*[`Elements/src/Geometry/Interfaces/IBoundedCurve.cs` line 5 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVebC3-TSobYwuQqQe:-NSWWkEN5RTveOeJt_f2:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Interfaces/IBoundedCurve.cs#L5)):*

curces => curves
</blockquote></details>

Done.
___
*[`Elements/src/Math/Domain1d.cs` line 37 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVf3Lh7FoGFN8nvdta:-NSWD0QpE3BbsV6D4hR6:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Math/Domain1d.cs#L37)):*

good catch!
</blockquote></details>

Done.
___
*[`Elements/test/LineTests.cs` line 646 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVkC9tBr2xpirfPt56:-NSWWhLu7v3XKhIMPtcA:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/test/LineTests.cs#L646)):*

Seems like we missed an automatic lint fix here, this should be 
`line.Length() / 2` with spaces
</blockquote></details>

Done.
___
*[`Elements.Serialization.IFC/src/Serialization/IFC/IFCExtensions.cs` line 37 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVkejK1rTcbAJZM8M-:-NSWWffC2cNrenkML0B6:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements.Serialization.IFC/src/Serialization/IFC/IFCExtensions.cs#L37)):*

I think this can be `.TransformedLine` to avoid the cast. 
</blockquote></details>

Done.
___
*[`Elements/src/Units.cs` line 360 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVaDT_8bVPvkA4JGqB:-NSVmj0h9_5y_nLGyEMt:b-a7zhew) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Units.cs#L360)):*

Not sure why this method belongs in `Units`. Also, appears to be unused by the code  — remove?
</blockquote></details>

Ah yes, created this and used it before realizing that `Domain.Includes` existed. Removed.

andrewheumann (changes requested):
> Reviewed 10 of 17 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer and @ikeough)

___
*[`Elements/src/Geometry/Arc.cs` line 14 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVbo5T56nrfzuEMJOq:-NSa2-2C3yMzt2qCQTxX:bb6ywmx) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L14)):*

I've changed the validation for arc span to check whether the proposed span is > 360.0, regardless of start /end. LMK what you think.
</blockquote></details>

LGTM
___
*[`Elements/src/Geometry/Polygon.cs` line 28 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NSa2KMO09VGpIRDgGoT:-NSa2KMO09VGpIRDgGoU:bcqmjpv) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Geometry/Polygon.cs#L28)):*
> ```Smalltalk
>         internal Plane _plane;
> 
>         internal override bool IsClosedForRendering => false;
> ```

shouldn't a polygon be `IsClosedForRendering` true?
___
*[`Elements/src/Geometry/Polyline.cs` line 1416 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVi-HE61CQ5JWLIEP4:-NSa1imx-uLYzDnmHndJ:b-alji5c) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Polyline.cs#L1416)):*

This is a good catch. I've made `GetSampleParameters` abstract so every curve has to implement it. I've also simplified `RenderVertices()` to have one implementation, which required allowing curve types to specify which GL line mode they should use to render. See the addition of the internal property `IsClosedForRendering` which is set to `false` by default and `true` for polygons so that the line mode switches to `LOOP`. All `RenderVertices` was doing differently for polygons was adding one more vertex to the collection so it would draw as a loop.
</blockquote></details>

That feels a lot cleaner!

anthonie-kramer (approved):
> Reviewed 10 of 48 files at r1, 1 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Arc.cs` line 98 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NST2WQkEkkgMi1qVmyS:-NST2WQkEkkgMi1qVmyT:b-qk7rtl) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L98)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="radius">The radius of the arc.</param>
>         /// <param name="startAngle">The angle from 0.0, in degrees, at which the arc will start with respect to the positive X axis.</param>
> ```

assuming counter-clockwise, but good to add hints here...

anthonie-kramer (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Line.cs` line 578 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NShKeZ3CYvTMLx_0hlZ:-NShKeZ3CYvTMLx_0hl_:b-k67v8i) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Geometry/Line.cs#L578)):*
> ```Smalltalk
>         public override Vector3 Mid()
>         {
>             return Start.Average(End);
> ```

or `PointAt(0.5)` ? Fewer operations = more precise (though probably negligible anyway)
___
*[`Elements/src/Search/LeftMostPointComparer.cs` line 35 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NShObCM6Ghw9dn7RA6E:-NShObCM6Ghw9dn7RA6F:b3cnq4u) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Search/LeftMostPointComparer.cs#L35)):*
> ```Smalltalk
>             var b = _getSegment(t2);
> 
>             const double small = 0.001;
> ```

I don't know if we use constants like these in other places, but there should probably be global constants for tolerances, etc.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @anthonie-kramer)

___
*[`Elements/src/Geometry/Arc.cs` line 98 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NST2WQkEkkgMi1qVmyS:-NSlTviaDOP4tT96lQ-c:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Arc.cs#L98)):*

assuming counter-clockwise, but good to add hints here...
</blockquote></details>

Done.
___
*[`Elements/src/Geometry/Line.cs` line 578 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NShKeZ3CYvTMLx_0hlZ:-NSlU7Ls2gKsm_6j8_uc:b-f113z8) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Geometry/Line.cs#L578)):*

or `PointAt(0.5)` ? Fewer operations = more precise (though probably negligible anyway)
</blockquote></details>

Parameterization of lines is now 0->length, so `PointAt` would have to know the length of the line. `Average` is actually the shortest path.
___
*[`Elements/src/Geometry/Polygon.cs` line 28 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NSa2KMO09VGpIRDgGoT:-NSlUGvU-itbL1HYFS2R:b-896fjx) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Geometry/Polygon.cs#L28)):*

shouldn't a polygon be `IsClosedForRendering` true?
</blockquote></details>

Done
___
*[`Elements/src/Geometry/Polyline.cs` line 1416 at r3](https://reviewable.io/reviews/hypar-io/Elements/957#-NSVi-HE61CQ5JWLIEP4:-NSlUkVEDJ_eeXShpe1f:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/77adecf95edb979ffe119fa1933cc16a41cfb4b7/Elements/src/Geometry/Polyline.cs#L1416)):*

That feels a lot cleaner! 
</blockquote></details>

Done.
___
*[`Elements/src/Search/LeftMostPointComparer.cs` line 35 at r5](https://reviewable.io/reviews/hypar-io/Elements/957#-NShObCM6Ghw9dn7RA6E:-NSlUrnGDkNxX4kVbRdD:b-roo687) ([raw file](https://github.com/hypar-io/Elements/blob/1fcad68f61e7b039549f61f95eec069f29ef52e1/Elements/src/Search/LeftMostPointComparer.cs#L35)):*

I don't know if we use constants like these in other places, but there should probably be global constants for tolerances, etc.
</blockquote></details>

We can address in a separate PR. The idea here was to keep the functionality as close to the original functionality, while accommodating the reparameterization.

andrewheumann (approved):
As discussed, when we're ready to merge this, let's try to get an elements release out the door with any in-flight bug fixes before we make the 2.0 bump. but LGTM!

> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957#-:-NSpyWtD6AzafYxTxDsV:b3rgskm)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @anthonie-kramer)

anthonie-kramer (approved):
> Reviewed 19 of 48 files at r1, 1 of 2 files at r2, 2 of 10 files at r3, 10 of 17 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/957)* status: :shipit: complete! 2 of 1 approvals obtained

PR #958, opened 2023-04-04
Improve ForceAngleCompliance in case of plane change.

DmytroMuravskyi:
BACKGROUND:
- While working with the function I have found that although it produces correct angles, it does unnecessary deviations sometimes.

DESCRIPTION:
- If there are two segments that need to be aligned but previous segment is perpendicular to both of them, it's better to use triangle that include those two segment instead, since perpendicular segment will have correct angle with any configuration of that triangle.

TESTING:
- Added new test PolylineForceAngleCompliancePathChangesPlane that shows a case new changes does better.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- I also included new cast fixes in Grid2d that were in my repo for a while. They don't break any build but help avoid unnecessary up casts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/958)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 1.6.0
+
+### Changed
+
+- `Grid2D.GetCellGeometry` return type changed from `Curve` to `Polygon`.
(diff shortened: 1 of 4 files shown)


DmytroMuravskyi:
Refinement is required

PR #960, opened 2023-04-09
Increment tag to support multiple similar CSGs.

ikeough:
BACKGROUND:
See #965.

DESCRIPTION:
This PR adds an additional field to the tuple vertex `tag` which makes vertices unique.

REQUIRED:

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/960)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 ### Changed
 - Element deserialization no longer requires `Name` to be present — it can be omitted.
 
+### Fixed
+- #965
+
 ## 1.4.0
 
 ### Added
--- a/Elements/src/Geometry/CsgExtensions.cs
+++ b/Elements/src/Geometry/CsgExtensions.cs
@@ -39,7 +39,14 @@ namespace Elements.Geometry
                                                    bool mergeVertices = false,
                                                    Func<(Vector3, Vector3, UV, Color?), (Vector3, Vector3, UV, Color?)> modifyVertexAttributes = null)
         {
-            var buffers = Tessellation.Tessellation.Tessellate<GraphicsBuffers>(csgs.Select(csg => new CsgTessellationTargetProvider(csg)),
+            var providers = new List<CsgTessellationTargetProvider>();
+            uint solidId = 0;
+            foreach (var csg in csgs)
+            {
+                providers.Add(new CsgTessellationTargetProvider(csg, solidId));
+                solidId++;
+            }
+            var buffers = Tessellation.Tessellation.Tessellate<GraphicsBuffers>(providers,
                                                                                 mergeVertices,
                                                                                 modifyVertexAttributes);
             return buffers;
@@ -226,7 +233,8 @@ namespace Elements.Geometry
         /// </summary>
         /// <param name="vertices"></param>
         /// <param name="faceId"></param>
-        internal static ContourVertex[] ToContourVertexArray(this IList<Csg.Vertex> vertices, int faceId)
+        /// <param name="solidId"></param>
+        internal static ContourVertex[] ToContourVertexArray(this IList<Csg.Vertex> vertices, uint faceId, uint solidId = 0)
         {
             var contour = new ContourVertex[vertices.Count];
             for (var i = 0; i < vertices.Count; i++)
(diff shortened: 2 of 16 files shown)


andrewheumann (changes requested):
It took me a really long time to figure out how this was working — I had to read the code 4-5 times before it clicked. I think some comments would be useful, especially at moments related to this optimization / the new `offset`. 

Let me make sure I understand:
- in `ProcessSolidsAsCSG`, we were creating a collection of `SolidTesselationTargetProvider`s, one for each solid operation. 
- Each `SolidTesselationTargetProvider` has a `GetTessellationTargets` method which returns a new `SolidFaceTessAdapter`.
- each `SolidFaceTessAdapter` is responsible for the tesselation of a single face. 
- the `SolidFaceTessAdapter`'s `GetTess` method calls `face.ToContourVertexArray`.
- `ToContourVertexArray` attaches a `faceId` to the `Data` object we hang on the contour vertex.
- most accessors of `Data` don't care about this `faceId` at all, and discard it. But `PackTessellationsIntoBuffers` constructs a map of vertex indices by tag / faceId. 
- This is the code that was failing: because faceIds could be the same across solids, we would sometimes accidentally re-use the wrong vertex. 
- We solve this in this PR by attempting to guarantee unique faceIds between solids.  

So I think all that makes sense now, but I think this whole flow could stand some better commenting.

I also worry about the fragility of modifying the `faceId` on the contour vertex with an offset — this assumes faceIDs are necessarily sequential/consecutive within a given solid, and that offsetting by the face count guarantees uniqueness. I wonder if it wouldn't be better to be explicit about a `solidIndex`/`solidOperationId` or something, as an optional second piece of information in the vertex map's tuple key.

It also bothers me that we're calling something `faceId` which now is no longer the actual `face.Id` of the face it came from — If we want to use the offset/sum strategy, we should at least call it something different so future us doesn't assume it will be the faceId. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/960#-:-NS_sDtkDKM9nw_AJwMD:bh7cmwv)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Tessellation/SolidTessellationTargetProvider.cs` line 19 at r1](https://reviewable.io/reviews/hypar-io/Elements/960#-NS_v1cP922t8m6VJwn1:-NS_v1cP922t8m6VJwn2:b-zevztn) ([raw file](https://github.com/hypar-io/Elements/blob/fcf7b9aeeb307bb9e7213056355d48d76f1ca638/Elements/src/Geometry/Tessellation/SolidTessellationTargetProvider.cs#L19)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="solid"></param>
>         /// <param name="transform"></param>
> ```

Please add xml comment for offset param, and descriptions

ikeough (commented):
Good points all around. I've used your breakdown of this flow above as the comments, and I've moved to using an increasing `solidId` as a separate field in the tuple and the vertex map. Now vertices are disambiguated using their vertex id (tag), their face id, and their solid id.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/960#-:-NSmjQ6g-aBjo0Qw7VV9:b-dba4ut)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Tessellation/SolidTessellationTargetProvider.cs` line 19 at r1](https://reviewable.io/reviews/hypar-io/Elements/960#-NS_v1cP922t8m6VJwn1:-NSmjMfU5M2g0Hkt868Z:bb6un1j) ([raw file](https://github.com/hypar-io/Elements/blob/fcf7b9aeeb307bb9e7213056355d48d76f1ca638/Elements/src/Geometry/Tessellation/SolidTessellationTargetProvider.cs#L19)):*

Please add xml comment for offset param, and descriptions
</blockquote></details>

Done

andrewheumann (approved):
> Reviewed 3 of 6 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/960)* status: :shipit: complete! 1 of 1 approvals obtained

PR #959, opened 2023-04-08
Ellipse

ikeough:
This PR requires #957.

BACKGROUND:
We'd like to support more curve types. With recent work done in #957, we now have a better foundation for supporting curves and trimmed versions of those curves. 

DESCRIPTION:
- Adds a new unbound curve type `Ellipse`.
- Adds a new trimmed curve type `EllipticalArc`.
  
FUTURE WORK:
- Length calculations for ellipses are imprecise. More accurate approximations should be provided in the future.

REQUIRED:
- [x] All changes are up to date in `CHANGELOG.md`.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/959)

PR #962, opened 2023-04-10
Arc by three points and better fillets.

ikeough:
This PR requires #957.

BACKGROUND:
Building on the work of #957 and #959, I'd like to be able to build arcs by three points. `IFCIndexedPolyCurve` requires this and the ability to build an arc by three points is generally useful.

DESCRIPTION:
- Implements `Arc.ByThreePoints`.
- Creates internal versions of angle calculation methods that return radians, so that we don't have to convert internally.
- Refactors `Fillet` into a static method `Arc.Fillet` and makes the fillet code capable of creating fillets between any two lines, not just those that lie in the XY plane.

TESTING:
Is there a scenario where we shouldn't be able to create an arc by three points? Here's the ones I've captured with tests.
- Three colinear points.
- Three coincident points.
  
FUTURE WORK:
- This is a naive implementation using only ootb methods. It's quite possible there's a more efficient way to do this. 

REQUIRED:

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/962)


serenayl (approved):
> Reviewed 3 of 6 files at r2, 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/962)* status: :shipit: complete! 1 of 1 approvals obtained

PR #970, opened 2023-04-17
Fix: wall intersections where bounded areas are not being created

katehryhorenko:
BACKGROUND:
The wall intersections were not created inside some visibly bound areas:

image

image

It happens on some E-shaped intersections:
image

The problem occurs when building a tree. The order of vertexes looks like this:
image

And the tree traversal with LeftMostPointComparer:
image
 On step 2 the line on the right site is not added, because it's not the leftMost point. On step 3 the bottom line is removed as successor to  upper line. So point 4 is never added to the tree:
image

The solution is two change events order and change the LeftMost point location for lines || to Y axis:
image

DESCRIPTION:
- This PR contains changes above and tests with different E-shaped networks

TESTING:
- A added tests

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/970)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 ### Fixed
 - Using Multiple `ModelText`s would sometimes result in a corrupted texture atlas, with cutoff text. This is fixed.
 - #865
+- `Network`: intersections are not created for some E-shape cases
 
 ## 1.6.0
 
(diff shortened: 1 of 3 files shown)


ikeough (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/970)* status: :shipit: complete! 1 of 1 approvals obtained

PR #972, opened 2023-04-20
Normalized

ikeough:
BACKGROUND:
During discussion of of the impact of the changes to curve parameterization in 2.0, @wynged said that it would ease the transition for existing code that uses `PointAt`, if we could have replacement methods like `PointAtNormalized`.

DESCRIPTION:
- Adds `PointAtNormalized` and `TransformAtNormalized`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/972)

### DIFF
--- a/Elements/src/Geometry/BoundedCurve.cs
+++ b/Elements/src/Geometry/BoundedCurve.cs
@@ -124,7 +124,7 @@ namespace Elements.Geometry
         /// Get a transform whose XY plane is perpendicular to the curve, and whose
         /// positive Z axis points along the curve.
         /// </summary>
-        /// <param name="u">The transform at a parameter along the curve between 0.0 and 1.0.</param>
+        /// <param name="u">The parameter along the curve between 0.0 and 1.0.</param>
         /// <returns>A transform.</returns>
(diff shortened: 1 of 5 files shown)


wynged (approved):
> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/972)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/test/ArcTests.cs` line 322 at r2](https://reviewable.io/reviews/hypar-io/Elements/972#-NTVRXZS06smnVAwvJSs:-NTVRXZS06smnVAwvJSt:b-mpa8yw) ([raw file](https://github.com/hypar-io/Elements/blob/2a9abd6bda5318cdbaaeb8b20f6bd415bae667bf/Elements/test/ArcTests.cs#L322)):*
> ```Smalltalk
>             // The start, end, and mid points are the same, but we also want to ensure
>             // that the curve didn't flip directions as well. We do this by testing
>             // another random parameter. 
> ```

if start and end are correct how could it have become flipped?

PR #973, opened 2023-04-20
Fix coplanar check

andrewheumann:
BACKGROUND:
- Our Vector3 coplanarity check failed for certain input that it should have accepted, which was causing immediate failures on deserialization, breaking several functions. 

DESCRIPTION:
- Rewrites the coplanarity check to use a dot product method instead of a triple-product method. The triple-product computes a _volume_, which means there's no sensible way to think about it in terms of `Vector3.EPSILON` — the dot product lends itself nicely to the interpretation of "check that the points are within EPSILON distance of a plane passing through the first three non-coplanar points" 

TESTING:
- I added some tests, including one that would definitely have failed with the old method. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/973)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Using Multiple `ModelText`s would sometimes result in a corrupted texture atlas, with cutoff text. This is fixed.
 - #865
 - `Network`: intersections are not created for some E-shape cases
+- `Vector3.AreCoplanar` would sometimes return false negatives.
 
 ## 1.6.0
 
--- a/Elements/src/Geometry/Vector3Extensions.cs
+++ b/Elements/src/Geometry/Vector3Extensions.cs
@@ -10,13 +10,13 @@ namespace Elements.Geometry
     public static class Vector3Extensions
     {
         /// <summary>
-        /// Are the provided points on the same plane?
+        /// Returns the indices of the first three non-collinear points in a list.
         /// </summary>
-        /// <param name="points"></param>
-        public static bool AreCoplanar(this IList<Vector3> points)
+        /// <param name="points">The list of points to search.</param>
+        /// <returns>A tuple of three integers representing the indices of the first three non-collinear points in the list.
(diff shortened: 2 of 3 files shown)


wynged (changes requested):
one refactor suggestion

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/973#-:-NTVapfM3wNF8iLQQl5N:b-xvezst)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Vector3Extensions.cs` line 24 at r1](https://reviewable.io/reviews/hypar-io/Elements/973#-NTV_cYx72GB9mnrESk9:-NTV_cYx72GB9mnrESkA:b-e583iu) ([raw file](https://github.com/hypar-io/Elements/blob/07f0ac2a75f1acf11e200f731dc77bd429d9e27b/Elements/src/Geometry/Vector3Extensions.cs#L24)):*
> ```Smalltalk
>             int p1Index = -1;
>             int p2Index = -1;
>             for (int i = 1; i < points.Count; i++)
> ```

might be nice to pull this first step out into it's own method.  `FirstThreeNonCollinearIndices`
Maybe even `TryGetThreeNonCollinearIndices` which would let you retrun true below without additional checks, but that might be messy with out variables etc...

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/973)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[`Elements/src/Geometry/Vector3Extensions.cs` line 24 at r1](https://reviewable.io/reviews/hypar-io/Elements/973#-NTV_cYx72GB9mnrESk9:-NT_6tJe-o-qz-uCO2hr:bb7r0mg) ([raw file](https://github.com/hypar-io/Elements/blob/07f0ac2a75f1acf11e200f731dc77bd429d9e27b/Elements/src/Geometry/Vector3Extensions.cs#L24)):*

OK, did my best... I think it's still a little ugly, but it works. 
</blockquote></details>

wfm!

PR #976, opened 2023-04-21
Indexed polycurve

ikeough:
BACKGROUND:
We previously built a class called `Contour` that was `IEnumerable<Curve>`. This class was meant to support planar compositions of any curve type. It was not thoroughly designed. After completing work on trimmed curves in #957 #961 #959, we decided to rebuild a composite curve type capable of representing networks of lines and arcs.

DESCRIPTION:
This PR implements a new class `IndexedPolycurve` which is a composite curve of line and arc segments. The external representation of the curve is a set of vertices and a set of sets of of indices which implicitly describe the constructing curves. In this way it is similar to [IFCIndexedPolycurve](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD1/HTML/schema/ifcgeometryresource/lexical/ifcindexedpolycurve.htm). In addition to being `IEnumerable<BoundedCurve>`, allowing the user to do `foreach(var curve in polycurve)`, it is also a `BoundedCurve`, allowing the user to find the start, end, point at at parameter, and transforms along the curve.
The `Contour` class is deprecated, and locations that used the `Contour` class, such as the creation of a filleted polycurve from a polygon, now return an `IndexedPolycurve`.

TESTING:
  
FUTURE WORK:
- We need to take a position on whether this curve is strictly planar. We do not check and we do not enforce planarity. As such, you can make a polycurve that represents, for example, a duct run, with arc sections and lines running in 3D. This seems useful.
- This implementation sacrifices memory for inheritance. The indexed polycurve has a set of curves that are derived from its vertices. We don't want to have to recreate these whenever someone uses the iterator, so we cache them. This is an improvement over what we used to do with polygons, allowing people to call `Segments()`, but it creates curves even for the polygon case where they might not ever be used.
- This is a mutable class, which is problematic because of the curves that we cache. If the caller adds or removes vertices to the polygon, the index sets and the internal curve sets need to be rebuilt. We can handle this by making the vertex collection read only and adding modifier methods, or we can make the iterator construct the curves. 

REQUIRED:

An indexed polycurve created by filleting a polygon.
image

An indexed polycurve created by filleting a polyline.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - `Arc.Fillet`
 - `Ellipse`
 - `EllipticalArc`
+- `IndexedPolycurve`
 
 ### Changed
 
--- a/Elements/src/GeometricElement.cs
+++ b/Elements/src/GeometricElement.cs
@@ -361,8 +361,8 @@ namespace Elements
             var solidItems = solids.ToArray();
             var voidItems = voids.ToArray();
 
-            // Don't try CSG booleans if we only have one one solid and no voids.
-            if (solids.Count() == 1 && voids.Count() == 0)
+            // Don't try CSG booleans if we only have one one solid.
+            if (solids.Count() == 1)
             {
                 csg = solids.First();
             }
@@ -370,10 +370,7 @@ namespace Elements
             {
                 csg = csg.Union(solidItems);
             }
-            else
-            {
-                return csg;
-            }
+
 
             if (voids.Count() > 0)
             {
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -13,6 +13,12 @@ namespace Elements.Geometry
     /// </example>
     public partial class Arc : TrimmedCurve<Circle>, IEquatable<Arc>
     {
+        /// <summary>
+        /// The domain of the curve.
+        /// </summary>
+        [JsonIgnore]
+        public override Domain1d Domain => new Domain1d(Units.DegreesToRadians(this.StartAngle), Units.DegreesToRadians(this.EndAngle));
+
         /// <summary>The angle from 0.0, in degrees, at which the arc will start with respect to the positive X axis.</summary>
         [JsonProperty("StartAngle", Required = Required.Always)]
         [System.ComponentModel.DataAnnotations.Range(0.0D, 360.0D)]
@@ -71,7 +77,6 @@ namespace Elements.Geometry
             this.BasisCurve = new Circle();
             this.StartAngle = 0;
             this.EndAngle = 360;
-            this.Domain = new Domain1d(Units.DegreesToRadians(this.StartAngle), Units.DegreesToRadians(this.EndAngle));
         }
 
         /// <summary>
@@ -88,7 +93,6 @@ namespace Elements.Geometry
             this.BasisCurve = new Circle(@center, @radius);
             this.StartAngle = @startAngle;
             this.EndAngle = @endAngle;
-            this.Domain = new Domain1d(Units.DegreesToRadians(@startAngle), Units.DegreesToRadians(@endAngle));
         }
 
         private void Validate(double startAngle, double endAngle, double radius)
@@ -127,7 +131,6 @@ namespace Elements.Geometry
             this.BasisCurve = new Circle(radius);
             this.StartAngle = startAngle;
             this.EndAngle = endAngle;
-            this.Domain = new Domain1d(Units.DegreesToRadians(@startAngle), Units.DegreesToRadians(@endAngle));
         }
 
         /// <summary>
@@ -140,7 +143,6 @@ namespace Elements.Geometry
         {
             Validate(Units.RadiansToDegrees(startParameter), Units.RadiansToDegrees(endParameter), circle.Radius);
             this.BasisCurve = circle;
-            this.Domain = new Domain1d(startParameter, endParameter);
             this.StartAngle = Units.RadiansToDegrees(this.Domain.Min);
             this.EndAngle = Units.RadiansToDegrees(this.Domain.Max);
         }
@@ -160,7 +162,6 @@ namespace Elements.Geometry
         {
             Validate(Units.RadiansToDegrees(startParameter), Units.RadiansToDegrees(endParameter), radius);
             this.BasisCurve = new Circle(transform, radius);
-            this.Domain = new Domain1d(startParameter, endParameter);
             this.StartAngle = Units.RadiansToDegrees(this.Domain.Min);
             this.EndAngle = Units.RadiansToDegrees(this.Domain.Max);
         }
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -143,8 +143,8 @@ namespace Elements.Geometry
         {
             Validate(Units.RadiansToDegrees(startParameter), Units.RadiansToDegrees(endParameter), circle.Radius);
             this.BasisCurve = circle;
-            this.StartAngle = Units.RadiansToDegrees(this.Domain.Min);
-            this.EndAngle = Units.RadiansToDegrees(this.Domain.Max);
+            this.StartAngle = Units.RadiansToDegrees(startParameter);
+            this.EndAngle = Units.RadiansToDegrees(endParameter);
         }
 
         /// <summary>
@@ -162,8 +162,8 @@ namespace Elements.Geometry
         {
             Validate(Units.RadiansToDegrees(startParameter), Units.RadiansToDegrees(endParameter), radius);
             this.BasisCurve = new Circle(transform, radius);
-            this.StartAngle = Units.RadiansToDegrees(this.Domain.Min);
-            this.EndAngle = Units.RadiansToDegrees(this.Domain.Max);
+            this.StartAngle = Units.RadiansToDegrees(startParameter);
+            this.EndAngle = Units.RadiansToDegrees(endParameter);
         }
 
         /// <summary>
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -433,7 +433,7 @@ namespace Elements.Geometry
         /// <param name="transform">The transform to apply.</param>
         public Arc TransformedArc(Transform transform)
         {
-            return new Arc(transform.OfPoint(this.BasisCurve.Transform.Origin), this.BasisCurve.Radius, StartAngle, EndAngle);
+            return new Arc(BasisCurve.Transform.Concatenated(transform), this.BasisCurve.Radius, Units.DegreesToRadians(StartAngle), Units.DegreesToRadians(EndAngle));
         }
 
         /// <summary>
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -419,10 +419,7 @@ namespace Elements.Geometry
             return new Arc(this.BasisCurve.Transform.Origin, this.BasisCurve.Radius, newStart, newEnd);
         }
 
-        /// <summary>
-        /// Construct a transformed copy of this Curve.
-        /// </summary>
-        /// <param name="transform">The transform to apply.</param>
+        /// <inheritdoc/>
         public override Curve Transformed(Transform transform)
         {
             return TransformedArc(transform);
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -350,8 +350,7 @@ namespace Elements.Geometry
 
         /// <inheritdoc/>
         public override double[] GetSubdivisionParameters(double startSetbackDistance = 0.0,
-                                                          double endSetbackDistance = 0.0,
-                                                          double minimumChordLength = 0.01)
+                                                          double endSetbackDistance = 0.0)
         {
             var min = this.Domain.Min;
             var max = this.Domain.Max;
@@ -374,7 +373,7 @@ namespace Elements.Geometry
             // d = 2 * r * sin(t/2)
             var r = this.BasisCurve.Radius;
             var two_r = 2 * r;
-            var d = Math.Min(minimumChordLength, two_r);
+            var d = Math.Min(DefaultMinimumChordLength, two_r);
             var t = 2 * Math.Asin(d / two_r);
             var div = (int)Math.Ceiling(angleSpan / t);
 
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -348,13 +348,10 @@ namespace Elements.Geometry
             return BasisCurve.Transform.XY();
         }
 
-        /// <summary>
-        /// Get parameters to be used to find points along the curve for visualization.
-        /// </summary>
-        /// <param name="startSetbackDistance">An optional setback from the start of the curve.</param>
-        /// <param name="endSetbackDistance">An optional setback from the end of the curve.</param>
+        /// <inheritdoc/>
         public override double[] GetSubdivisionParameters(double startSetbackDistance = 0.0,
-                                                       double endSetbackDistance = 0.0)
+                                                          double endSetbackDistance = 0.0,
+                                                          double minimumChordLength = 0.01)
         {
             var min = this.Domain.Min;
             var max = this.Domain.Max;
@@ -377,7 +374,7 @@ namespace Elements.Geometry
             // d = 2 * r * sin(t/2)
             var r = this.BasisCurve.Radius;
             var two_r = 2 * r;
-            var d = Math.Min(MinimumChordLength, two_r);
+            var d = Math.Min(minimumChordLength, two_r);
             var t = 2 * Math.Asin(d / two_r);
             var div = (int)Math.Ceiling(angleSpan / t);
 
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -295,7 +295,7 @@ namespace Elements.Geometry
         /// <summary>
         /// Calculate the length of the arc between start and end parameters.
         /// </summary>
-        public override double Length(double start, double end)
+        public override double ArcLength(double start, double end)
         {
             if (!Domain.Includes(start, true))
             {
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -450,20 +450,6 @@ namespace Elements.Geometry
         }
 
         /// <summary>
-        /// Get a point at parameter u on the arc.
-        /// </summary>
-        /// <param name="u">A parameter on the curve between 0.0 and 1.0.</param>
-        /// <returns>A point on the curve at parameter u.</returns>
-        public override Vector3 PointAtNormalized(double u)
-        {
-            if (u < 0 || u > 1)
-            {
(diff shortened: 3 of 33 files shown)


ikeough:
Set this back to draft to fix all test failures and to unify parameters on polygons as well.

andrewheumann (changes requested):
Some notes on this below, but I'm a little concerned about the inconsistency between the way some curves have an internal `Transform` and others do not. Yields surprising results like:


It may be that this is just something we need to account for in any metric methods like length, but I wonder if it's not a more fundamental issue. Another example — according to elements these two curves have the same length:






Also, seems like indexed polycurves don't serialize/deserialize correctly. 


> Reviewed 9 of 20 files at r1, 17 of 19 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NYwrlrS1N5R1UgvZzAC:b2n2rpq)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/StructuralFraming.cs` line 124 at r2](https://reviewable.io/reviews/hypar-io/Elements/976#-NW4turs7lp5fLgdYUtT:-NW4turs7lp5fLgdYUtU:b-solezo) ([raw file](https://github.com/hypar-io/Elements/blob/1ecaab62e2d3de0e74f8cf0754b8931f4d8c523c/Elements/src/StructuralFraming.cs#L124)):*
> ```Smalltalk
>             if (this.Representation.SolidOperations.Count == 0)
>             {
>                 if (this.Curve.GetType() == typeof(IndexedPolycurve))
> ```

Can this just be `if(this.Curve is IndexPolycurve pc)`? (maybe I'm missing some nuance that requires the `GetType()`/`typeof` comparison)
___
*[`Elements/src/Geometry/Arc.cs` line 435 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwpqHQBjP1qB35F__l:-NYwpqHQBjP1qB35F__m:b33ljz9) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Arc.cs#L435)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="transform">The transform to apply.</param>
>         public Arc TransformedArc(Transform transform)
> ```

There's a worrying inconsistency here — a transformed `Arc` may not be an `Arc` (and same goes for a `Circle`.). Transforms applied to a circle or arc can result in any sort of conic section. Some possible solutions:
- Just leave a note/warning in the XML doc comments. (This appears to be what RhinoCommon does — transformed `Circle`s stay `Circle`s, and they just give you a warning. For most operations though, they get turned into Nurbs Curves which can handle this transform logic more appropriately.)
- throw an exception for transformations that don't preserve arc-ness (everything except euclidean transforms and uniform scales)
- Throw an exception for transformations that are non-affine, and return an `EllipticalArc`. 
___
*[`Elements/src/Geometry/Bezier.cs` line 36 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwwAtE-IA8VtywLSMK:-NYwwAtE-IA8VtywLSML:bp78cw3) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Bezier.cs#L36)):*
> ```Smalltalk
>     public class Bezier : BoundedCurve
>     {
>         private int _lengthSamples = 500;
> ```

nit: this can be `readonly`
___
*[`Elements/src/Geometry/Curve.cs` line 16 at r2](https://reviewable.io/reviews/hypar-io/Elements/976#-NW4vOtG6T43NN3BZD8d:-NW4vOtG6T43NN3BZD8e:bcgjvre) ([raw file](https://github.com/hypar-io/Elements/blob/1ecaab62e2d3de0e74f8cf0754b8931f4d8c523c/Elements/src/Geometry/Curve.cs#L16)):*
> ```Smalltalk
>         /// A lower MinimumChordLength results in smoother curves.
>         /// </summary>
>         public static double MinimumChordLength = 0.1;
> ```

Why this change?
___
*[`Elements/src/Geometry/IndexedPolycurve.cs` line 27 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYx-rPy16MyijOLhPkl:-NYx-rPy16MyijOLhPkm:b-f6zv6p) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/IndexedPolycurve.cs#L27)):*
> ```Smalltalk
>         /// <summary>
>         /// A bounding box created once during the polyline's construction.
>         /// This will not be updated when a polyline's vertices are changed.
> ```

polyline => polycurve
___
*[`Elements/src/Geometry/Line.cs` line 76 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwoaPr30BHs1tIzUhe:-NYwoaPr30BHs1tIzUhf:bfwkqeb) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Line.cs#L76)):*
> ```Smalltalk
>         /// <param name="startParameter">The start parameter of the line segment.</param>
>         /// <param name="endParameter">The end parameter of the line segment.</param>
>         public Line(InfiniteLine line, double startParameter, double endParameter)
> ```

`startParameter` and `endParameter` are entirely unused — this seems like a mistake (uncaught from an earlier PR)
___
*[`Elements/src/Geometry/Line.cs` line 94 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwnGmm5rOKVY7YQm6N:-NYwnGmm5rOKVY7YQm6O:b-ouyofk) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Line.cs#L94)):*
> ```Smalltalk
>         /// Calculate the length of the line between two parameters.
>         /// </summary>
>         public override double ArcLength(double start, double end)
> ```

It was initially super strange to me that a `Line` would have an `ArcLength`. I recognize that this is technically correct, mathematically speaking, but it might risk confusing users. I wonder whether it would be clearer to call this "SubSegmentLength" or "SubCurveLength" or something?  I could be convinced that we should keep `ArcLength`. 

Also, this can result in a negative `ArcLength` — is that desirable?
___
*[`Elements/src/Geometry/Interfaces/ICurve.cs` line 11 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwxmZkCElKRXuhRjmY:-NYwxmZkCElKRXuhRjmZ:btyq0vs) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Interfaces/ICurve.cs#L11)):*
> ```Smalltalk
>         /// Get a point along the curve at parameter u.
>         /// </summary>
>         /// <param name="u">A parameter on the curve between domain.min and domain.max.</param>
> ```

if the curve is infinite, does it have domain.min / domain.max? you can get `PointAt` on an infinite line.
___
*[`Elements/test/PolygonTests.cs` line 159 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYx-HuIE7Tw-pKL15jB:-NYx-HuIE7Tw-pKL15jC:b-79c3ca) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/test/PolygonTests.cs#L159)):*
> ```Smalltalk
>             (
>                 new[]
>                 {
> ```

These `new[] {}` aren't necessary, did you add them for a reason? 
___
*[`Elements/test/PolygonTests.cs` line 1062 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwz6kUDj8UGaGNf3zJ:-NYwz6kUDj8UGaGNf3zK:b-7jwq6f) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/test/PolygonTests.cs#L1062)):*
> ```Smalltalk
>             {
>                 var pt = contour3.TransformAt(u);
>             }
> ```

is there an Assert that goes along with this, or was it something temporary?

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/IndexedPolycurve.cs` line 27 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYx-rPy16MyijOLhPkl:-NYx3-ZI8iNeUSNsDkfY:b-xefd2) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/IndexedPolycurve.cs#L27)):*

polyline => polycurve
</blockquote></details>

(we seem to say "polyline" in a lot of places internally, might be good to update throughout.)

andrewheumann (commented):
> Reviewed 1 of 19 files at r2.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/StructuralFraming.cs` line 124 at r2](https://reviewable.io/reviews/hypar-io/Elements/976#-NW4turs7lp5fLgdYUtT:-NYyRB6i52tS3O-MLCTu:b-qz80qz) ([raw file](https://github.com/hypar-io/Elements/blob/1ecaab62e2d3de0e74f8cf0754b8931f4d8c523c/Elements/src/StructuralFraming.cs#L124)):*

Can this just be `if(this.Curve is IndexPolycurve pc)`? (maybe I'm missing some nuance that requires the `GetType()`/`typeof` comparison)
</blockquote></details>

Nice catch! I don't yet default to pattern matching in C# and the linter doesn't tell me to, so I'm sure there's lot of places that can be improved. This is done.
___
*[`Elements/src/Geometry/IndexedPolycurve.cs` line 27 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYx-rPy16MyijOLhPkl:-NYyRyfKA6-G9zn482p0:b-m4cs99) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/IndexedPolycurve.cs#L27)):*

(we seem to say "polyline" in a lot of places internally, might be good to update throughout.)
</blockquote></details>

Yes! This is because this class used to be the polyline. Updated all.
___
*[`Elements/src/Geometry/Line.cs` line 76 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwoaPr30BHs1tIzUhe:-NYySTmA5zh427GIw26c:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Line.cs#L76)):*

`startParameter` and `endParameter` are entirely unused — this seems like a mistake (uncaught from an earlier PR)
</blockquote></details>

Done.
___
*[`Elements/src/Geometry/Line.cs` line 94 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwnGmm5rOKVY7YQm6N:-NYyT9luCwaDpPio-cCe:b8wn09y) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Line.cs#L94)):*

It was initially super strange to me that a `Line` would have an `ArcLength`. I recognize that this is technically correct, mathematically speaking, but it might risk confusing users. I wonder whether it would be clearer to call this "SubSegmentLength" or "SubCurveLength" or something?  I could be convinced that we should keep `ArcLength`. 

Also, this can result in a negative `ArcLength` — is that desirable?
</blockquote></details>

Fixed the negative arc length. 

It's a good question about ArcLength. I'm of a mind to leave it alone even though it's a bit weird. It's an override that all bounded curves share. I thought about what it should be called when I named it there and for arcs, circles, beziers, and ellipses, arc length, is the terminology. This has to be implemented because the base class has this marked as abstract. So we can't change the name unless we want to change it in the base. And I'd argue that if users are looking for "length" they'll probably naturally try `myLine.Length()` first.
___
*[`Elements/src/Geometry/Interfaces/ICurve.cs` line 11 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwxmZkCElKRXuhRjmY:-NYyckH8B9QP0C3c7Koa:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Interfaces/ICurve.cs#L11)):*

if the curve is infinite, does it have domain.min / domain.max? you can get `PointAt` on an infinite line.
</blockquote></details>

Done.
___
*[`Elements/test/PolygonTests.cs` line 159 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYx-HuIE7Tw-pKL15jB:-NYyetNE5ZO6uJ70v2yP:b93qany) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/test/PolygonTests.cs#L159)):*

These `new[] {}` aren't necessary, did you add them for a reason? 
</blockquote></details>

While taking a cue from the surrounding code, I forgot that we had an overload that took `params`. I've removed the `new[]` from this method and the one mentioned above.
___
*[`Elements/test/PolygonTests.cs` line 1062 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwz6kUDj8UGaGNf3zJ:-NYyd4cUE-gPxBKAxU9k:b-3frt06) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/test/PolygonTests.cs#L1062)):*

is there an Assert that goes along with this, or was it something temporary?
</blockquote></details>

Nope. Was for debuggging. Removed.

ikeough (commented):
This is such a good question about scale transforms. This is probably pervasive in the library. We don't really have a story here. I think you could make the argument that conics only need a plane instead of a transform to represent the slice through a cone (a cone which we also don't describe), which would avoid being able to scale that transform and expect something interesting to happen.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NYyhiVq4DjpbTBxdqt_:b75hmwa)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

ikeough (commented):
On the deserialization. <sigh> There wasn't a `JsonConstructor` with the exact right signature. There is one now, and it's hit during deserialization.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NYz53qqA24iaJNDx8Ww:bngld53)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

andrewheumann (commented):
These changes look good, and I don't think we have to totally resolve the Transform problem right now. Still curious about the MinimumArcLength change, and would love at the very least a warning in the xml doc comments for `TransformedArc`, `Transformed` etc about the kinds of transforms that may cause `length` or `area` to be computed incorrectly (or whatever other unexpected consequences users should be aware of.) an example from rhino docs — I don't think it has to be much more than this:



> Reviewed 5 of 6 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NZ0cesm1A7V5kcoWZeD:bs99bho)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @ikeough)

___
*[`Elements/src/Geometry/Line.cs` line 94 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwnGmm5rOKVY7YQm6N:-NZ0cFFEElhfWEPptZ8w:b5caylp) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Line.cs#L94)):*

Fixed the negative arc length. 

It's a good question about ArcLength. I'm of a mind to leave it alone even though it's a bit weird. It's an override that all bounded curves share. I thought about what it should be called when I named it there and for arcs, circles, beziers, and ellipses, arc length, is the terminology. This has to be implemented because the base class has this marked as abstract. So we can't change the name unless we want to change it in the base. And I'd argue that if users are looking for "length" they'll probably naturally try `myLine.Length()` first.
</blockquote></details>

Yeah, I buy that. The term technically applies to all curves, so I think it's reasonable.

ikeough (commented):
I was wondering how other libraries handled this. Thanks for the Rhino reference. I've updated the docs and, for extra credit, I've used `<inheritdoc/>`, which I didn't know of until just now, to write docs for this method on the base class and have those docs inherited everywhere. 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NYz6A6u2be2DuULSW0H:bvmu4pt)* status: 1 change requests, 0 of 1 approvals obtained

___
*[`Elements/src/Geometry/Curve.cs` line 16 at r2](https://reviewable.io/reviews/hypar-io/Elements/976#-NW4vOtG6T43NN3BZD8d:-NZ1CXtz9fooUlgieKnv:b-uq59lh) ([raw file](https://github.com/hypar-io/Elements/blob/1ecaab62e2d3de0e74f8cf0754b8931f4d8c523c/Elements/src/Geometry/Curve.cs#L16)):*

Why this change?
</blockquote></details>

I have to think on a solution here. This got bumped up because in places where we were previously converting circles into polygons for the purpose of making solid geometry, we're now using actual circles and the faceting became far too small (increasing the duration of our test suite to ~15min. due to some CSG tests). We need to be able to set this on an as-needed basis.

ikeough (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Arc.cs` line 435 at r3](https://reviewable.io/reviews/hypar-io/Elements/976#-NYwpqHQBjP1qB35F__l:-NZ1DB1Q7UIt_Z77vur5:bzf3bmu) ([raw file](https://github.com/hypar-io/Elements/blob/79984a8bc6e3d66b0828c040381061ab4612e622/Elements/src/Geometry/Arc.cs#L435)):*

There's a worrying inconsistency here — a transformed `Arc` may not be an `Arc` (and same goes for a `Circle`.). Transforms applied to a circle or arc can result in any sort of conic section. Some possible solutions:
- Just leave a note/warning in the XML doc comments. (This appears to be what RhinoCommon does — transformed `Circle`s stay `Circle`s, and they just give you a warning. For most operations though, they get turned into Nurbs Curves which can handle this transform logic more appropriately.)
- throw an exception for transformations that don't preserve arc-ness (everything except euclidean transforms and uniform scales)
- Throw an exception for transformations that are non-affine, and return an `EllipticalArc`. 
</blockquote></details>

Comments are now in the docs.

andrewheumann (approved):
:lgtm:

> Reviewed 11 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/976#-:-NZ1MMjj3Ue8ttEuvhYV:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

PR #978, opened 2023-05-01
allow custom prefix exclusion

andrewheumann:
BACKGROUND:
- Revit conversion was failing weirdly because we were searching through `Autodesk.*` types in the app domain for element types. 

DESCRIPTION:
- Allows custom exclusions for the `JsonInheritanceConverter`. 

TESTING:
- I called `JsonInheritanceConverter.ExcludeTypePrefixesFromTypeCache("Autodesk")` and conversion worked again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/978)

### DIFF
--- a/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs
+++ b/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs
@@ -97,7 +97,7 @@ namespace Elements.Serialization.JSON
                 {
                     types = assembly.GetTypes();
                 }
-                catch (Exception)
+                catch (ReflectionTypeLoadException)
                 {
                     failedAssemblyErrors.Add($"Failed to load assembly: {assembly.FullName}");
                     continue;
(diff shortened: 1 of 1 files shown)


andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/978)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/Serialization/JSON/JsonInheritanceConverter.cs` line 100 at r1](https://reviewable.io/reviews/hypar-io/Elements/978#-NUT30dx6U_j2GjRV58k:-NUT3Tvm3v6WTvvnyKTC:b-owq7wn) ([raw file](https://github.com/hypar-io/Elements/blob/d4a9568bf7c7327c3dcf6e76008074c003c0393d/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L100)):*

What's the reason for changing this exception type?
</blockquote></details>

ah that was left over from debugging, fixed

PR #979, opened 2023-05-11
Emissive

ikeough:
BACKGROUND:
While debugging materials for Hypar's Unreal integration, we noticed that emissive factor was only being applied if the emissive texture was used as well. You should be able to use one without the other.

DESCRIPTION:
This PR applies the emissive factor whether there is an emissive texture or not, and it sets the emissive factor to 0 by default, as opposed to 1.0. An emissive factor of 1 would cause all materials to appear to emit light.

REQUIRED:

Materials with increasing emissive factors.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/979)


ikeough:
Emissive in the glTF spec is RGB. We currently flatten this into one value. I might also take this opportunity to make it RGB, so we can support colors of emitted light that aren't the material color.

katehryhorenko (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/979)* status: :shipit: complete! 1 of 1 approvals obtained

PR #986, opened 2023-06-14
Bug fix Polygon.Contains3D

pederlh:
BACKGROUND:
Fixed a bug causing issues in own code.

DESCRIPTION:
The `out Containment containment` parameter in Polygon.Contains3D isn't updated if one of the other polygon's vertices is outside this polygon.  In that case `Containment.Inside` would be passed instead of `Containment.Outside`.

TESTING:
None.
  
FUTURE WORK:
None.

REQUIRED:

COMMENTS:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/986)


ikeough (approved):
Thanks very much for the fix @pederlh.

ikeough:
Seems like there's one failing test. Do you have any interest in attempting to fix?

pederlh:
I’m off the grid for almost a month, so it would be great if you guys could take a look 😊 Otherwise I’ll revisit this later on

pederlh:
The `dotnet test -c release` workflow process that was causing an error earlier passes without issue locally. Could you try to rerun the workflow @ikeough?

andrewheumann:
Thanks for the nudge @pederlh, all tests are passing now, so I'm going to go ahead and merge. Sorry feedback on this PR took so long, and thanks for your contribution!!

PR #988, opened 2023-06-21
Do not break loop if one of the tess objects is empty

katehryhorenko:
BACKGROUND:
One of the walls from Revit doesn't appear in viewer 

DESCRIPTION:
Solid cannot be tessellated when it's an extrusion created from profile where not all points are on the same plane. Even when they are on the same plane but the offset is really small:
image

This offset can produce tess objects without elements.

By replacing "return" with "continue" we are allowing to tesselate this type of extrusions 

TESTING:
I published this change to the JSON to Model function (TEST). I tried it on Basic Revit model where I mostly saw this issue.

 
image

image

REQUIRED:
_**Do I need to add this change to the backlog?**_

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/988)


wynged (approved):
I don' think so, I got feedback that some low level fixes like this don't warrant entry in the change log.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/988#-:-NYYuBsjB_6c9ykjB_WR:b-4vqhcc)* status: :shipit: complete! 1 of 1 approvals obtained

katehryhorenko (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/988#-:-NZkbw_iEibqL8UPMPRG:bb6un1j)* status: 1 change requests, 0 of 1 approvals obtained

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/988)* status: :shipit: complete! 1 of 1 approvals obtained

PR #990, opened 2023-07-05
Topography.Trimmed

andrewheumann:
BACKGROUND:
- There was a request from some customers to support trimming a topography with a profile. The existing Trim method wasn't exactly what was needed, because it:
  - left behind a solid rather than just the top surface
  - didn't support voids within the projected shape
  - mutated the existing topography instead of creating a new mesh
DESCRIPTION:
- adds `Topography.Trimmed`, which returns a new Topopgraphy trimmed by a Profile
- adds `Topography.TopMesh` which returns just the upwards facing portion of the topo without the sides
- Updates `Topography.Trim` to support Profile instead of Polygon 
- Fixes an issue where cloning a mesh would result in incorrect vertex-triangle relationships, due to an oversight in one Triangle constructor

TESTING:
- Added a test which tests all new methods:
image

  
FUTURE WORK:
- calling `GetNakedBoundaries` results in some really funky shapes with these meshes... there's still something weird about the mesh topology after a CSG operation.

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/990)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - `Message.Warning`
 - `Topography.Trimmed`
 - `new Topography(Topography other)`
+- `Topography.TopMesh()`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 - `Message.Info`
 - `Message.Error`
 - `Message.Warning`
+- `Topography.Trimmed`
+- `new Topography(Topography other)`
 
 ### Changed
 
@@ -49,6 +51,7 @@
 - Fix the polygon centroid calculation to remove collinear vertices.
 - Fix the tests for 3dCentroid testing.
 - `Message` created from `Message.FromPoint` now has `Transform.Origin` set exactly on original point.
+- Certain `Triangle` constructors would not correctly update `Vertex.Triangles`, this is fixed.
 
 ## 1.6.0
 
--- a/Elements/src/Geometry/Mesh.cs
+++ b/Elements/src/Geometry/Mesh.cs
@@ -253,6 +253,18 @@ Triangles:{Triangles.Count}";
         }
 
         /// <summary>
+        /// Removes the specified triangle from the mesh and updates the vertex-triangle relationships.
+        /// </summary>
+        /// <param name="face">The triangle to remove.</param>
+        public void RemoveTriangle(Triangle face)
+        {
+            this.Triangles.Remove(face);
+            foreach(var vert in face.Vertices) {
+                vert.Triangles.Remove(face);
+            }
+        }
+
+        /// <summary>
         /// Add a vertex to the mesh.
         /// </summary>
         /// <param name="position">The position of the vertex.</param>
--- a/Elements/src/Geometry/Triangle.cs
+++ b/Elements/src/Geometry/Triangle.cs
@@ -46,20 +46,9 @@ namespace Elements.Geometry
         {
             this.Vertices = new[] { a, b, c };
 
-            if (!a.Triangles.Contains(this))
-            {
-                a.Triangles.Add(this);
-            }
-
-            if (!b.Triangles.Contains(this))
-            {
-                b.Triangles.Add(this);
-            }
-
-            if (!c.Triangles.Contains(this))
-            {
-                c.Triangles.Add(this);
-            }
+            a.Triangles.Add(this);
+            b.Triangles.Add(this);
+            c.Triangles.Add(this);
 
             var ab = (b.Position - a.Position).Unitized();
             var bc = (c.Position - a.Position).Unitized();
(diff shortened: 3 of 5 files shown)


anthonie-kramer (changes requested):
Just a comment about deep copies, and a question about the updated constructor

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/990#-:-NZbq3CH8hnJ8cGfsTwH:bfghvcq)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Topography.cs` line 226 at r1](https://reviewable.io/reviews/hypar-io/Elements/990#-NZbjyF0B2H3HOgnWP_j:-NZbjyF0B2H3HOgnWP_k:bkvfy0t) ([raw file](https://github.com/hypar-io/Elements/blob/a42860adbed106594b9cd15751d41a774f5c6db6/Elements/src/Topography.cs#L226)):*
> ```Smalltalk
>         /// Construct a new Topography by duplicating another one.
>         /// </summary>
>         public Topography(Topography topography)
> ```

My preference when making a copy of an object is just to create a deep copy which is safer and doesn't require maintenance... Not as performant as your constructor code below, but thought I would mention it.



_Code snippet:_
```Smalltalk
        public Topography CreateDeepCopy(Topography topography)
        {
            return JsonConvert.DeserializeObject<Topography>(JsonConvert.SerializeObject(topography));
        }
```
___
*[`Elements/src/Geometry/Triangle.cs` line 34 at r1](https://reviewable.io/reviews/hypar-io/Elements/990#-NZbnN0C5dCPFfkixP3h:-NZbnN0C5dCPFfkixP3i:bhjqt2d) ([raw file](https://github.com/hypar-io/Elements/blob/a42860adbed106594b9cd15751d41a774f5c6db6/Elements/src/Geometry/Triangle.cs#L34)):*
> ```Smalltalk
>             foreach (var vert in vertices)
>             {
>                 if (!vert.Triangles.Contains(this))
> ```

I'm trying to understand how this works... Since `this` is currently being created, how could `vert.Triangles.Contains(this)`?

Unless we are overriding the `Equals` method... but then I think we would be creating duplicate objects that have different references?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/990)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer)

___
*[`Elements/src/Topography.cs` line 226 at r1](https://reviewable.io/reviews/hypar-io/Elements/990#-NZbjyF0B2H3HOgnWP_j:-NZbuDlGADPTsESqy2d9:b9655lg) ([raw file](https://github.com/hypar-io/Elements/blob/a42860adbed106594b9cd15751d41a774f5c6db6/Elements/src/Topography.cs#L226)):*

My preference when making a copy of an object is just to create a deep copy which is safer and doesn't require maintenance... Not as performant as your constructor code below, but thought I would mention it.
</blockquote></details>

Everywhere else we do Copy constructors in `Elements` we do it expressly, like this. (`Grid1d`, `Box`, `Mesh`, etc). Serialize/Deserialize feels like a hack to me — I can see how it would be safe, but especially since serialized `Topography` objects can be quite large, I feel like it's maybe not appropriate here. 
___
*[`Elements/src/Geometry/Triangle.cs` line 34 at r1](https://reviewable.io/reviews/hypar-io/Elements/990#-NZbnN0C5dCPFfkixP3h:-NZbuoXZ-axi7Ym-oAtW:b-xjaeyi) ([raw file](https://github.com/hypar-io/Elements/blob/a42860adbed106594b9cd15751d41a774f5c6db6/Elements/src/Geometry/Triangle.cs#L34)):*

I'm trying to understand how this works... Since `this` is currently being created, how could `vert.Triangles.Contains(this)`?

Unless we are overriding the `Equals` method... but then I think we would be creating duplicate objects that have different references?
</blockquote></details>

This is a good catch — I guess it's not possible for the vertex to already have this triangle when accessed via this pathway. removed!

anthonie-kramer (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/990#-:-NZc2UbA-gtmow63Xf3l:b-uwkowd)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[`Elements/src/Topography.cs` line 226 at r1](https://reviewable.io/reviews/hypar-io/Elements/990#-NZbjyF0B2H3HOgnWP_j:-NZc2-3LBXIIP81tmM5Z:bl8uelx) ([raw file](https://github.com/hypar-io/Elements/blob/a42860adbed106594b9cd15751d41a774f5c6db6/Elements/src/Topography.cs#L226)):*

Everywhere else we do Copy constructors in `Elements` we do it expressly, like this. (`Grid1d`, `Box`, `Mesh`, etc). Serialize/Deserialize feels like a hack to me — I can see how it would be safe, but especially since serialized `Topography` objects can be quite large, I feel like it's maybe not appropriate here. 
</blockquote></details>

For sure, there really aren't any efficient ways to create a deep copy. We are probably safe if this object is not likely to expand its properties in the future.

PR #992, opened 2023-07-20
Closed regions search refactoring

srudenkoamc:
BACKGROUND:
 - Network.FindAllClosedRegions cannot find all closed regions correctly in some cases.

DESCRIPTION:
 - Moved the logic that searches for closed regions from Network.FindAllClosedRegions into a separate class NetworkCycleCoverage.
 - Added NetworkNode class that contains both index and position of vertex. It allows to simplify the code significantly.
 - Added NetworkEdge class that is similar to LocalEdge class, but is based on NetworkNode and has several additional methods.
 - Replaced _adjacencyList with an adjacency matrix based on NetworkNode and NetworkEdge.
 - Algorithm stays the same. The tests are not affected by the changes.

TESTING:
- The tests are not affected by the changes.
- Additionally tested on https://hypar.io/workflows/820ea7b7-1bf1-4779-9a6b-20da7dd6c969.
  
FUTURE WORK:
- Remove leaf branches before search.
- Should be a good idea to traverse edges instead of nodes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/992)

### DIFF
--- a/Elements/src/Search/Network.cs
+++ b/Elements/src/Search/Network.cs
@@ -441,129 +441,8 @@ namespace Elements.Search
         /// forming closed regions in the network.</returns>
         public List<List<int>> FindAllClosedRegions(List<Vector3> allNodeLocations)
         {
-            var regions = new List<List<int>>();
-
-            // TODO: This code is a mess. We use several methods for tracking
-            // traversal data: LocalEdge instances, nodeVisits, and structures
-            // internal to the traversal methods. These can be combined so that
-            // we are only using LocalEdges.
-
-            var leafNodes = new List<int>();
-            var allEdges = new List<LocalEdge>();
-
-            for (var i = 0; i < this.NodeCount(); i++)
-            {
-                var localEdges = EdgesAt(i);
-                var edgeCount = localEdges.Count();
-                // Leaf nodes
-                if (edgeCount == 1)
-                {
-                    leafNodes.Add(i);
-                }
-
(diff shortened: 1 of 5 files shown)


srudenkoamc:
> One little comment.
> 
> > Reviewed 5 of 5 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/992#-:-N_nMrIZ-LyfuzebJye6:bvucrkq)_ status: 1 change requests, 0 of 1 approvals obtained (waiting on @srudenkoamc)
> 
> _[`Elements/src/Search/NetworkCycleCoverage.cs` line 192 at r1](https://reviewable.io/reviews/hypar-io/Elements/992#-N_nML1F1rJxzXuyzh1V:-N_nML1F1rJxzXuyzh1W:b-vllg8u) ([raw file](https://github.com/hypar-io/Elements/blob/6b72b7d53fa3b334f3491b7723bfc84f0c10063c/Elements/src/Search/NetworkCycleCoverage.cs#L192)):_
> 
> > ```smalltalk
> >                 var edge = GetEdge(path[j], path[j + 1]);
> > 
> >                 // if (edge == null)
> > ```
> 
> Let's remove this commented code. Even though it was there previously, I don't think it should be.

Fixed in https://github.com/hypar-io/Elements/pull/992/commits/399ec29cbb7bded43592a0e552f7bc6c9818f9eb.

PR #993, opened 2023-07-21
Real mid and more

jamesbradleym:
BACKGROUND:
- In our transition to 2.0 much of our curve behavior changes in lieu of their parameterization, here we introduce a new interface IHasArcLength which enforces Midpoint(), PointAtLength(), PointAtNormalizedLength(), and DivideByLength() which applies to Circles, Lines, Polylines, and Beziers.
- Additionally Bezier felt unloved given the push to supporting curves, to perk it up we introduce Split functionality and normalized parameterization + parameter at functionality, and Bezier.ConstructPiecewiseCubicBezier()

DESCRIPTION:
- Gives Circles, Lines, Polylines, and Beziers more flexibility in the information they can return.

TESTING:
- Play around with Circles, Lines, Polylines, and Beziers looking at Midpoint(), PointAtLength(), PointAtNormalizedLength(), and DivideByLength() and for Bezier poke around ParameterAt(), PointAtNormalized() Building multi point Bezier.ConstructPiecewiseCubicBezier(), dividing Beziers, spliting Beziers.
  
FUTURE WORK:

REQUIRED:
- [ x] All changes are up to date in `CHANGELOG.md`.

COMMENTS:
- Any other notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - `Bezier` now inherits from `BoundedCurve`.
 - `Polyline` is now parameterized 0->length.
 - `Circle` is now parameterized 0->2Pi.
-- `Line` is now parameterized 0->length.
+- `Circle` now implements the `IHasArcLength` interface
 - `Vector3.DistanceTo(Ray ray)` now returns positive infinity instead of throwing.
 - `Message`: removed obsolete `FromLine` method.
 - `AdaptiveGrid`: removed obsolete `TryGetVertexIndex` with `tolerance` parameter.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,21 +23,31 @@
 - `Topography.Trimmed`
 - `new Topography(Topography other)`
 - `Topography.TopMesh()`
+- `Bezier.PointAtLength()`
+- `Bezier.PointAtNormalizedLength()`
+- `Bezier.ParameterAt()`
+- `Bezier.DivideByLength()`
+- `Bezier.Split()`
+- `Bezier.SplitAt()`
+- `Bezier.SplitByLength()`
+- `Bezier.ConstructPiecewiseCubicBezier()`
 
 ### Changed
 
 - `Polyline` now inherits from `BoundedCurve`.
 - `Polyline` is now parameterized 0->length.
-- `Polyline` now implements the `IHasArcLength` interface
+- `Polyline` now implements the `IHasArcLength` interface.
 - `Arc` now inherits from `TrimmedCurve<Circle>`.
-- `Arc` is now parameterized 0->2Pi
+- `Arc` is now parameterized 0->2Pi.
 - `Line` now inherits from `TrimmedCurve<InfiniteLine>`.
 - `Line` is now parameterized 0->length.
-- `Line` now implements the `IHasArcLength` interface
+- `Line` now implements the `IHasArcLength` interface.
 - `Bezier` now inherits from `BoundedCurve`.
+- `Bezier` now implements the `IHasArcLength` interface.
+- `Bezier.ArcLength()` now uses Gauss quadrature approximation vs linear sampling.
 - `Polyline` is now parameterized 0->length.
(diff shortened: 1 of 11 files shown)


andrewheumann (changes requested):
> Reviewed 1 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/993)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer, @ikeough, @jamesbradleym, and @wynged)

___
*[`Elements/src/Geometry/Interfaces/IHasArcLength.cs` line 7 at r2](https://reviewable.io/reviews/hypar-io/Elements/993#-Na7mjn8-uzuoeJp89Uv:-Na8Imgh4Z8LYUBhpjO_:b-c684ef) ([raw file](https://github.com/hypar-io/Elements/blob/60a92d22bb91a8e10aac94293e880f65887f91fe/Elements/src/Geometry/Interfaces/IHasArcLength.cs#L7)):*

In general, I feel like this should just be bundled with `IBoundedCurve`. I don't see the distinction here for Arc Lengths, these seem to just be general curve methods.

It seems like maybe Circle is the exception since it is not a `BoundedCurve`...

Either way, it seems like we can abstract these into `ICurve` or `IBoundedCurve` unless there is an `Arc` specific method in here, which I would then recommend it gets named appropriately. Although, I don't see anything specific to arcs, such as arguments that are an `Angle` type.
</blockquote></details>

I agree I would like to see this get bundled up in some parent classes / potentially other interfaces. Are there any `BoundedCurve`s that don't / cant support this? (re: naming though, "ArcLength" is the term for measured length along any curve, not just arcs )

ikeough:
A couple of high level comments. I'm moving this out to 2.1 for further design. Much of what is in here should be, if we choose to bring it in, part of the `IBoundedCurve` interface, which is the base interface for all bounded curves. `Circle` is not a bounded curve, so it should not have an arc length. However, `Arc` which is a trimmed curve with a `Circle` for its basis curve _should_ have an arc length. As should `EllipticalArc`. We can discuss more when 2.0 is out the door.

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/993)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer, @ikeough, and @wynged)

___
*[`Elements/src/Geometry/Interfaces/IHasArcLength.cs` line 22 at r2](https://reviewable.io/reviews/hypar-io/Elements/993#-Na7mb8yBL_t5CcECvNl:-NaCoSD16sxMvdNLWjBU:bwl797o) ([raw file](https://github.com/hypar-io/Elements/blob/60a92d22bb91a8e10aac94293e880f65887f91fe/Elements/src/Geometry/Interfaces/IHasArcLength.cs#L22)):*

Is this different than `PointAtNormalized` in `IBoundedCurve`?
</blockquote></details>

Yes, This does not normalize the parameter value to the length of the curve, it simply remaps the parameter space from 0->1. As Polyline and Bezier are not equally parameterized you would not get the midpoint from something like PointAtNormalized(0.5).



___
*[`Elements/src/Geometry/Interfaces/IHasArcLength.cs` line 28 at r2](https://reviewable.io/reviews/hypar-io/Elements/993#-Na7m5Ur3cQA3UrfCucQ:-NaCmaNn-x_UmN1E2GLn:b8dbo33) ([raw file](https://github.com/hypar-io/Elements/blob/60a92d22bb91a8e10aac94293e880f65887f91fe/Elements/src/Geometry/Interfaces/IHasArcLength.cs#L28)):*

How is this different than `Mid()`?
</blockquote></details>

Mid() finds the mid parameter, for a Line or Circle this is fine as the middle of the parameter space is the same as the length based midpoint. For Polyline and Bezier, however, parameter space is not length based. For bezier the parameter is weighted by control points and for polyline it is vertex based.

Left side = MidPoint()
Right side = Mid()

Issue #994, opened 2023-07-24
Arc.Polyline doesn't work if its domain has negative length.


DmytroMuravskyi:
It Arc is created with start angle bigger than end angle - it's domain is negative. Start is bigger than End.
Arc.ToPolyline tried to iterates from Min to Max and immediately leaves the loop with result always consist of only 1 point.

**To Reproduce**
Run following code:
```
var arc1 = new Arc(Vector3.Origin, 2.0, 0.0, -90.0);
var p = arc1.ToPolyline();
```

**Expected behavior**
Polyline with 11 points created.

**Desktop (please complete the following information):**
 - OS: Windows 10.

**Additional context**
Negative domain is considered valid in Elements. 
There are to options here:
- Do check for bigger domain in ToPolyline, similar to how it's done in GetSubdivisionParameters. There can be other places with similar issue what will not be covered by local fix.
- Force StartAngle to be smaller angle in Arc constructor. This removes certain flexibility from Arc class.

wynged:
There is a third option.  
When a decreasing angle is provided, we should flip the basis circle (e.g. -Z normal) and make the angle increasing (but preserve the user's thought about "start" "end" in the parameterization — IOW, second parameter angle corresponds to "end", always.)

So:  
in Arc constructors, if the user supplies a decreasing angle range, flip the order to be increasing and flip the basis curve direction to preserve the user's intended direction

Resulting behavior should follow this diagram.
![image](https://github.com/hypar-io/Elements/assets/5872187/37f909c8-1ca1-487e-824f-0674fc15b083)

@andrewheumann  is the real expert though, and should review the solution. :)

DmytroMuravskyi:
After some tries, it's not possible to implement what is shown because Arc parameter space is not  0 => Length but tied to the circle: first one is 0 to pi/2, second is 0 to -pi/2 or indeed 0 to pi/2 if normal is reversed. But third one is pi/2 to 0 in original direction or 3pi/2 to 2pi in reverse. Flipping Z or Transform is also have side effect of flipping X or Y axis as well to preserve correct hand rule I think this is too confusing.
It's either consistent domain and simple maintenance or mora flexibility.
I did a PR with support of negative domain and fixes for Arc functions that were not consistent: https://github.com/hypar-io/Elements/pull/998

PR #998, opened 2023-07-26
Arc: fix interactions negative domain in GetSubdivisionParameters, Complement and ToPolyline.

DmytroMuravskyi:
BACKGROUND:
- While working on Arc I realized that I can't call ToPolyon for the arc that has StartAngle > EndAngle. Result had only one point.

DESCRIPTION:
- Arc.GetSubdivisionParameters: returns parameters starting from StartAngle even if StartAngle > EndAngle.EndAngle.
- Arc,ToPolyline: now works if StartAngle > EndAngle.EndAngle.
- Arc.Complement: now works if StartAngle > EndAngle.EndAngle or if both angles are negative.

TESTING:
- Tests are added for each change
  
FUTURE WORK:
- There is discussion about the solution in https://github.com/hypar-io/Elements/issues/994 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/998)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@
 - Fix the tests for 3dCentroid testing.
 - `Message` created from `Message.FromPoint` now has `Transform.Origin` set exactly on original point.
 - Certain `Triangle` constructors would not correctly update `Vertex.Triangles`, this is fixed.
+- `Arc.GetSubdivisionParameters` returns parameters starting from StartAngle even if StartAngle > EndAngle.EndAngle.
+- `Arc.ToPolyline` now works if StartAngle > EndAngle.EndAngle.
+- `Arc.Complement` now works if StartAngle > EndAngle.EndAngle or if both angles are negative.
 
 ## 1.6.0
 
--- a/Elements/src/Geometry/Arc.cs
+++ b/Elements/src/Geometry/Arc.cs
@@ -354,17 +354,12 @@ namespace Elements.Geometry
         {
             var min = this.Domain.Min;
             var max = this.Domain.Max;
+            var sign = Math.Sign(max - min);
+            var startSetback = startSetbackDistance * sign;
+            var endSetback = (Length() - endSetbackDistance) * sign;
 
-            var flip = max < min;
-
-            if (flip)
-            {
-                max = this.Domain.Min;
-                min = this.Domain.Max;
-            }
-
-            var startParam = ParameterAtDistanceFromParameter(startSetbackDistance, min);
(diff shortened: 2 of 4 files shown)


andrewheumann:
I think this solution is the wrong approach — it leave's a curve's tangent vector in disagreement with its parameter space. 
```cs
var arc1 = new Arc(transform, 2, 0, Math.PI / 2.0);
var arc2 = new Arc(transform, 2, Math.PI / 2.0, 0);
Assert.Equal(arc1.PointAtNormalized(0.2), arc2.PointAtNormalized(0.8));
Assert.True(arc1.TransformAtNormalized(0.2).ZAxis.Dot(arc2.TransformAtNormalized(0.8).ZAxis) < -0.9999);
```
This will fail with this change. 

I have implemented the earlier proposed approach here, which satisfies this expectation.
https://github.com/hypar-io/Elements/pull/999

DmytroMuravskyi:
Few details I noticed about your implementation that might be by design or already known:
- Direction in TransformAt is opposite from Arc direction (may be by design): 
```
            var arc = new Arc(Vector3.Origin, 5.0, 0.0, 180.0);
            var t = arc.TransformAt(Math.PI / 2);
            Assert.Equal(new Vector3(0, 1, 0), t.XAxis);
            Assert.Equal(new Vector3(0, 0, 1), t.YAxis);
            Assert.Equal(new Vector3(1, 0, 0), t.ZAxis);
```
- Literals with this approach will produce values affected by transformation (users should be well aware of this behavior and understand that numbers they see in Domain are not the final one they are getting) :
```
            var arc = new Arc(Vector3.Origin, 5.0, 180, 45);
            var t = arc.TransformAt(Math.PI / 2); //X = -0.707106781186, Y = 0.70710678118654, Z = 1.22464679914735E-16
            var p = arc,PointAt(Math.PI / 2); //X = -3.5355339059327373, Y = 3.5355339059327386, Z = 6.1232339957367663E-16
```
Few changes that can be copied from this PR into new simplified one:
- GetSubdivisionParameters doesn't follow parametrization. 
```
            var arc = new Arc(Vector3.Origin, 2.0, 0.0, -90.0);
            var parameters = arc.GetSubdivisionParameters();
            Assert.Equal(0, parameters.First()); //Fails and returns -PI/2
```
- Complement for negative arc returns more negative arc.
            arc = new Arc(Vector3.Origin, 1, -40, -80);
            comp = arc.Complement();
            Assert.Equal(-40, comp.StartAngle); // returns -400
            Assert.Equal(280, comp.EndAngle); //returns -80

PR #999, opened 2023-07-26
don't permit arcs to have decreasing angle domains

andrewheumann:
BACKGROUND:
- https://github.com/hypar-io/Elements/issues/994
- We want to allow users to construct Arcs with apparently decreasing angle domains, in order to have intuitive control over arc direction.

DESCRIPTION:
- Updates Arc constructors to ensure that `EndAngle` is always greater than `StartAngle`.

TESTING:
- I added a test for every modified constructor, to verify that for any pair of angles A,B, arc(A,B) looked just like arc(B,A), but parameterized in the opposite direction. 
- I added a test to construct an `IndexedPolycurve` containing an arc which runs "clockwise" with respect to its circle / the Z axis, to ensure consistency in direction of the overall polycurve:
image
- I added a test for issue #994's original complaint, and visually verified that it worked. 
- All tests pass. 
  
FUTURE WORK:
- We may need to introduce a notion of `SenseAgreement` as STEP does for more complex curves, such that a subcurve could act like it was running in the opposite direction from its basis curve. Or maybe not `¯\_(ツ)_/¯`

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/999)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - `Polyline` is now parameterized 0->length.
 - `Arc` now inherits from `TrimmedCurve<Circle>`.
 - `Arc` is now parameterized 0->2Pi
+- `Arc` now automatically corrects decreasing angle domains to be increasing, while preserving direction.
 - `Line` now inherits from `TrimmedCurve<InfiniteLine>`.
 - `Line` is now parameterized 0->length.
 - `Bezier` now inherits from `BoundedCurve`.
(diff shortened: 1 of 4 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/999)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1011, opened 2023-08-23
Support `Color?` and `Vector3?` in elements code generation

andrewheumann:
BACKGROUND:
- In the Hypar CLI, we have supported nullable `Color` and `Vector3` values for a while, as of https://github.com/hypar-io/Hypar/pull/403 and https://github.com/hypar-io/Hypar/pull/492. However, we never brought these improvements over to Elements, meaning that it was supported in `input_schema` / `overrides` but not in element schemas. 

DESCRIPTION:
- Moves the Type Generator from `hypar` to `Elements`

TESTING:
- I tested this in tandem w/ `hypar init`, and I also wrote a test to ensure that the correct types are generated. 
  
FUTURE WORK:
- Once this is in, we should update `Hypar` to use it instead of having duplicated code. 

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - `Polygon.Contains3D` passed wrong `out Containment containment` parameter in some cases.
+- Code generation supports `Vector3?` and `Color?` types.
 
 ## 2.0.0
 
--- a/Elements.CodeGeneration/src/Templates/Class.Constructor.Record.liquid
+++ b/Elements.CodeGeneration/src/Templates/Class.Constructor.Record.liquid
@@ -8,7 +8,7 @@
 
 {% assign sortedProperties = AllProperties | sort: "Name" -%}
 {% assign sortedParentProperties = parentProperties | sort: "Name" -%}
-
+{% if AllProperties[0] %}
 [JsonConstructor]
 {% if IsAbstract %}protected{% else %}public{% endif %} {{ClassName}}({% for property in AllProperties -%}{% if skipComma -%}{% assign skipComma = false %}{% else %}, {% endif -%} {{ property.Type }} @{{ property.Name | safeidentifierlower | defaultforelementargument: baseClassName }}{% endfor -%})
 {% assign skipComma = true -%}
@@ -20,6 +20,7 @@
 this.{{property.PropertyName}} = @{{property.Name | safeidentifierlower}};
     {% endfor -%}
 }
+{% endif %}
 
 // Empty constructor
 {% if IsAbstract %}protected{% else %}public{% endif %} {{ClassName}}()
--- a/Elements.CodeGeneration/src/TypeGenerator.cs
+++ b/Elements.CodeGeneration/src/TypeGenerator.cs
@@ -413,8 +413,8 @@ namespace Elements.Generate
                 TypeNameGenerator = new ElementsTypeNameGenerator(),
                 PropertyNameGenerator = new ElementsPropertyNameGenerator(),
             };
-
-            var generator = new CSharpGenerator(schema, settings);
+            var typeResolver = new TypeResolver(settings);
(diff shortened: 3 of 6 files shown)


anthonie-kramer (changes requested):
What is the strategy for handling a `struct` definition vs an `object` definition? Are we expecting users to convert their `Nullable` types into the `Elements` `struct` types to use in a function?

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011#-:-NcZo2G07mYyfPTRnjHh:b-kh9pn)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements.CodeGeneration/test/TestData/NullableAndNonNullableTypesSchema.json` line 62 at r1](https://reviewable.io/reviews/hypar-io/Elements/1011#-NcZnAAc9QZroACzwSfa:-NcZnAAd6omufN1-fT8i:b-pdcw8h) ([raw file](https://github.com/hypar-io/Elements/blob/114872a080b2d50a1cda5524cb3133eecba49616/Elements.CodeGeneration/test/TestData/NullableAndNonNullableTypesSchema.json#L62)):*
> ```JSON
>             "title": "Color"
>         },
>         "NullableColor": {
> ```

Isn't this the`NonNullableColor`, and the one above is the `NullableColor` ?

andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer)

___
*[`Elements.CodeGeneration/test/TestData/NullableAndNonNullableTypesSchema.json` line 62 at r1](https://reviewable.io/reviews/hypar-io/Elements/1011#-NcZnAAc9QZroACzwSfa:-NccpXSB-gaoPvvjLLss:by7keby) ([raw file](https://github.com/hypar-io/Elements/blob/114872a080b2d50a1cda5524cb3133eecba49616/Elements.CodeGeneration/test/TestData/NullableAndNonNullableTypesSchema.json#L62)):*

Isn't this the`NonNullableColor`, and the one above is the `NullableColor` ?
</blockquote></details>

No, `Color.json` is nullable (type ["object", "null"] ), and NonNullableColor is type `"object"`.

andrewheumann (commented):
Objects are always nullable — we don't need to `?` them. This only matters for types which are already `struct`s in Elements, like Color and Vector3

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011#-:-NccpmPOBNNYN3hkKHWX:bfytrgn)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer)

anthonie-kramer (approved):
Right, I am just trying to understand the purpose of nullability, since the `Elements` `Vector3` is a `struct`... I am guessing that the schema-based `Vector3.json` is just used for object creation on the front end and then converted at some point to the `Elements` version for use in functions.

Anyway, I won't hold this up, just curious!

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011#-:-NcheaIaBMJbHqlaBw-C:b-unqxj4)* status: :shipit: complete! 1 of 1 approvals obtained

andrewheumann (commented):
both `Vector3` and `Color` are built-in types in elements which are implemented as `struct`s. We do a thing when doing codegen — for both Elements and inputs — where we check if a type already exists in the `dependencies` project, so that we can avoid generating a conflicting type. It happens that the `Color` schema is already configured to be nullable by default, whereas the `Vector3` schema is not.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1011#-:-NchkPyn9oBIu6AyHonf:b-y520ee)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1013, opened 2023-08-26
Thickened polylines

andrewheumann:
BACKGROUND:
- The Hypar front end supports the concept of a "Thickened Polyline," which is used for wall + corridor drawing. Until now, we've relied on a schema type, and had to ship around a separate library to reproduce the "thickening" behavior that takes place on the client — but now that this type has settled into usefulness, it should get a place in Elements.

DESCRIPTION:
- Adds the `ThickenedPolyline` class. 

TESTING:
- Several tests were added, including an example for docs.
- I have also been using this code extensively in existing projects (https://github.com/hypar-io/function-Circulation/), so trust its behavior.

image

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1013)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 2.0.1
 
+### Added
+- `Elements.Geometry.ThickenedPolyline`
+
 ### Fixed
 
 - `Polygon.Contains3D` passed wrong `out Containment containment` parameter in some cases.
--- a/Elements/src/Geometry/ThickenedPolyline.cs
+++ b/Elements/src/Geometry/ThickenedPolyline.cs
@@ -161,10 +161,11 @@ namespace Elements.Geometry
                 // for each edge, compute all the info we'll need for it: its left
                 // thickness / right thickness, its plane angle to the x axis, and
                 // whether it is pointing towards or away from this point. Sort the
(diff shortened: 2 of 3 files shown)


andrewheumann (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1013)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/test/ThickenedPolylineTests.cs` line 57 at r1](https://reviewable.io/reviews/hypar-io/Elements/1013#-NcxLCoC7a7EIM7Sh3jl:-Nd13Xr71Bz8KBi1Cr2g:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/48312a8b623a45c6bad3d4ba4728cbccd3888773/Elements/test/ThickenedPolylineTests.cs#L57)):*

can we at least put a TODO about some asserts we might make here?  or add a couple?  The code is a bit complex, would be nice to know if we break something compare to intended behavior.
</blockquote></details>

Done.

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1013)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #1007, opened 2023-08-02
Regression Elements.Geometry.Solids.Sweep


pederlh:
**Describe the bug**
From version 1.6.0 to 2.0.0, an exception is thrown when initializing some Sweep objects, in cases where possible earlier. A reproducible example is included below. 

```
Test method HyparDebugging.GeometryTest.Sweepbug threw exception: 
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'The vector could not be created. One or more of the components was NaN.')
    Stack Trace:
        at Elements.Geometry.Vector3..ctor(Double x, Double y, Double z)
   at Elements.Geometry.Solids.Solid.ProjectEdgeAlong(Edge[] loop, Vector3 v, Plane p)
   at Elements.Geometry.Solids.Solid.SweepEdgesBetweenPlanes(Edge[] loop1, Vector3 v, Plane end)
   at Elements.Geometry.Solids.Solid.SweepEdges(Transform[] transforms, Edge[] openEdge)
   at Elements.Geometry.Solids.Solid.SweepFaceAlongCurve(Polygon perimeter, IList`1 holes, IBoundedCurve curve, Double startSetback, Double endSetback, Double profileRotation)
   at Elements.Geometry.Kernel.CreateSweepAlongCurve(Profile profile, BoundedCurve curve, Double startSetback, Double endSetback, Double profileRotation)
   at Elements.Geometry.Solids.Sweep.UpdateGeometry()
   at Elements.Geometry.Solids.Sweep..ctor(Profile profile, BoundedCurve curve, Double startSetback, Double endSetback, Double profileRotation, Boolean isVoid)
```

**To Reproduce**
```C#
// Works in version 1.6.0, not in 2.0.0
Polygon crossSection = new(new List<Vector3>
{
    new Vector3(x:0.2500,  y:0.0000),
    new Vector3(x:0.2378,  y:0.0773),
    new Vector3(x:0.2023,  y:0.1469),
    new Vector3(x:0.1469,  y:0.2023),
    new Vector3(x:0.0773,  y:0.2378),
    new Vector3(x:0.0000,  y:0.2500),
    new Vector3(x:-0.0773, y:0.2378),
    new Vector3(x:-0.1469, y:0.2023),
    new Vector3(x:-0.2023, y:0.1469),
    new Vector3(x:-0.2378, y:0.0773),
    new Vector3(x:-0.2500, y:0.0000),
[...]

PR #1016, opened 2023-09-01
Regression fix for Elements.Geometry.Solids.Sweep

DmytroMuravskyi:
BACKGROUND:
- Where was reported an issue about Sweep: https://github.com/hypar-io/Elements/issues/1007
- The reason code worked before is that setbacks parameters were just ignored.
- Sweep in the issue crashes because setback ends exactly on second vertex. 
- Setbacks need extra handling for special cases as it always assumes that start setback is on first segment, end - on last. This is not the case as they can easily exceed.

DESCRIPTION:
- Changed `GetSubdivisionParameters` to use startSetbackDistance and endSetbackDistance since they were just ignored.
- Changed `Polyline.Frames` and `Polygon.Frames` so they return only frames between setbacks. Since `CreateMiterTransform` and `CreateOrthogonalTransform` are focused on vertices, Transforms on setbacks are created manually.
    
TESTING:
- Added tests for updated functions as well as regression Sweep.
  
FUTURE WORK:
- `Polyline.TransformAt` feels incorrect as it says that u is "between 0.0 and length" but it's treated as 0 to 1 value. Also, since calculation inside is length based, including normal calculation, curve length can possibly influence frame vectors in a wrong way. 

REQUIRED:

COMMENTS:
- I will be away for a while, so feel free to steal this code if necessery.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016)


andrewheumann (changes requested):
> Reviewed 3 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @DmytroMuravskyi)

___
*[`Elements/src/Geometry/Polygon.cs` line 2094 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrLAH99icg0WE1Lrko:-NfrLAH99icg0WE1Lrkp:b-f0cpnk) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/src/Geometry/Polygon.cs#L2094)):*
> ```Smalltalk
>             // Calculate number of frames. 2 frames corresponding to end parameters.
>             // 1 if startIndex == endIndex.
>             var lenght = endIndex - startIndex + 3;
> ```

`lenght` => `length`
___
*[`Elements/src/Geometry/Polyline.cs` line 306 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrLN_G5tmBi9jQXxgT:-NfrLN_G5tmBi9jQXxgU:b-f0cpnk) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/src/Geometry/Polyline.cs#L306)):*
> ```Smalltalk
>             // Calculate number of frames. 2 frames corresponding to end parameters.
>             // 1 if startIndex == endIndex.
>             var lenght =  endIndex - startIndex + 3;
> ```

`lenght` => `length`
___
*[`Elements/test/PolygonTests.cs` line 2234 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrOUag9MnUTLrKx8Hy:-NfrOUag9MnUTLrKx8Hz:b-m8vmln) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/test/PolygonTests.cs#L2234)):*
> ```Smalltalk
>             Assert.True((Vector3.YAxis - Vector3.XAxis).IsParallelTo(frames[4].ZAxis));
> 
>             frames = polygon.Frames(2, 2);
> ```

polyhon.Frames(1, 1) will throw an exception here.

DmytroMuravskyi (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Geometry/Polygon.cs` line 2094 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrLAH99icg0WE1Lrko:-NftjYwh7VvFIa11VVAA:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/src/Geometry/Polygon.cs#L2094)):*

`lenght` => `length`
</blockquote></details>

Done.
___
*[`Elements/src/Geometry/Polyline.cs` line 306 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrLN_G5tmBi9jQXxgT:-NftjZvO6uET-u3pmAVS:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/src/Geometry/Polyline.cs#L306)):*

`lenght` => `length`
</blockquote></details>

Done.
___
*[`Elements/test/PolygonTests.cs` line 2234 at r1](https://reviewable.io/reviews/hypar-io/Elements/1016#-NfrOUag9MnUTLrKx8Hy:-Nftj_i_Cy0kDvIPZjxq:blh570a) ([raw file](https://github.com/hypar-io/Elements/blob/a96dc1875a89964d65d7b99509edb0ebec5fb157/Elements/test/PolygonTests.cs#L2234)):*

polyhon.Frames(1, 1) will throw an exception here.
</blockquote></details>

Good catch. Added index rounding for Polygon.

andrewheumann (changes requested):
The frames being found along the curve are flipped compared to how they used to be — I think this is an unintended regression. This also results in sweeps having "inside-out" normals:

## Elements 1.6.0:


## Elements 2.0.0:


This PR:


All three samples are running this code:
```cs
var polygon = new Polyline((0, 0), (3, 0), (3, 3), (0, 3));
var model = new Model();
model.AddElement(new ModelCurve(polygon, BuiltInMaterials.XAxis));
var frames = polygon.Frames(0.1, 0.1);
model.AddElements(frames.SelectMany(f => f.ToModelCurves()));

var profile = Polygon.L(0.2, 0.2, 0.05);
var ge = new GeometricElement {
    Representation = new Sweep(profile, polygon, 0.1, 0.1, 0, false)
};
model.AddElement(ge);
```


We have been inconsistent about this in the past, but I think the convention in our code is that the z of a curve's frame is _opposite_ the curve's tangent at that point. (Don't ask me why it is this way, I would 100% expect Z+ to be the same as the tangent.)

> Reviewed 3 of 9 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016#-:-Nfu8vplBhee0-GjlVrk:bo8sp9w)* status: 1 change requests, 0 of 1 approvals obtained

DmytroMuravskyi (commented):
Yes, I missed the fact that they are adjusted. Now frames look like this

Note that Frames behaves differently from TrasformAt as first and last frames are not rotated  based on position if they are somewhere in the middle of segment but just . One exception is when setback is coincides with a corner, then Frame is also the same as for corner vertex - rotated. I also made tests more explicit about expected results.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016#-:-NfugZ8i2k3UuNpL05tt:b8argyy)* status: 1 change requests, 0 of 1 approvals obtained

andrewheumann:
Looks good to me, thanks!

andrewheumann (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1016)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1026, opened 2023-09-13
Polygon: delete last vertex if it equals the first one

katehryhorenko:
BACKGROUND:
In drywall function the data exported from Revit and then proceed by transformation/union etc. can cause the creation of polygon where the last vertex duplicated the first one. "Edges" method tries to create a segment from this two points and fails (the line is too short) 

DESCRIPTION:
- Delete the last vertex if it equals the first one as part of polygon validation

TESTING:
- I published Drywall function with this change. It might be worth creating a test

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1026)

### DIFF
--- a/Elements/src/Geometry/Polygon.cs
+++ b/Elements/src/Geometry/Polygon.cs
@@ -77,6 +77,12 @@ namespace Elements.Geometry
 
             this.Vertices = Vector3.RemoveSequentialDuplicates(this.Vertices, true);
             DeleteVerticesForOverlappingEdges();
+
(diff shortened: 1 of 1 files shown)


andrewheumann (commented):
I think this is supposed to already be handled by the `RemoveSequentialDuplicates` method — can we investigate why that's not working? It is being called with `wrap = true`, which should mean it checks if the first / last vertex are identical. Perhaps something is being changed in the `DeleteVerticesForOverlappingEdges` method?

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1026#-:-NeE_GNw648gzyI2GffX:b10alax)* status: 0 of 1 approvals obtained

Issue #1032, opened 2023-09-28
Polygon.TransformAtNormalized returns nonsense result


andrewheumann:
To reproduce:
```cs
var rectangle = Polygon.Rectangle(4, 10);
var transformsNormalized = Enumerable.Range(0, 20).SelectMany((i) => {
    return rectangle.TransformAtNormalized(i / 20.0).ToModelCurves();
});
```
produces:

![image](https://github.com/hypar-io/Elements/assets/31935763/37a0dc90-ce2e-401c-8bbf-220b57c47e2d)

The positions are incorrect; there should be some along the shorter edges. The axes are also incorrect — many Y and Z appear to be zero-length.

Note that the code for `PointAtNormalized` finds the correct positions: 

```cs
var rectLength = rectangle.Length();
var transforms = Enumerable.Range(0, 20).Select((i) => {
    return new ModelCurve(new Line(rectangle.PointAtNormalized(i / 20.0), rectangle.Center())); // rectangle.TransformAt(i * rectLength / 20.0).ToModelCurves();
});
```

![image](https://github.com/hypar-io/Elements/assets/31935763/9c9353e7-cb6e-4a98-8988-873f94ce721d)

DmytroMuravskyi:
TransformAt is working wrong for polyline. At one point it does `var d = this.Length() * u;` which is wrong since u is no longer from 0 to 1. Since it's known that that each segment has domain of 1, much simpler approach can be used. 
Also, you can see that corner points have rotated transformation, but only two. This is because averaging is used on every vertex except first and last vertices. But for Polygon every vertex should be averaged, in my opinion.
Third issue I have found is not related but would be nice to clarify. IndexedPolycurve.CreateVerticesAndCurveIndices produces unique indexes. For two lines and arc in between it would be [0, 1], [2, 3, 4], [5, 6]. It's not only in contrast with CreateCurveIndices but also changes indices when polyline is copies. If the TransformedPolycurve is called on polycurve with indices [0, 1], [1, 2, 3], [3, 4], new polycurve will have indices as described above: [0, 1], [2, 3, 4], [5, 6] indices that out of range of verticies.

DmytroMuravskyi:
![image](https://github.com/hypar-io/Elements/assets/22522160/a1f24ee6-2922-45dd-bb07-625b60296afa)
The result I got

PR #1035, opened 2023-10-04
Fix TransformAt for Polyline

DmytroMuravskyi:
BACKGROUND:
- There is an issue https://github.com/hypar-io/Elements/issues/1032 that shown incorrect work of Polygon.TransformAtNormalized.

DESCRIPTION:
- The issue is tracked down to Polyline.TransformAt where parameter value is treated as domain value in one place but still as 0 to 1 in other place. The function is updated.
- It also treated TransformAt incorrectly for first and last vertex in the list if Polyline is actually a Polygon. Fixed it so all transforms at vertices are averaged.
- I also found that IndexedPolycurve.CreateVerticesAndCurveIndices produced unique indexes. For two lines and arc in between it would be [0, 1], [2, 3, 4], [5, 6]. It contradicted with the fact that vertices in the same function were considered as shared but also changes indices when polyline is copies. If the TransformedPolycurve is called on polycurve with indices [0, 1], [1, 2, 3], [3, 4], new polycurve will have indices as described above: [0, 1], [2, 3, 4], [5, 6] indices that out of range of verticies. Did the fix.

TESTING:
- Added PreservesIndicesTransformed and PolylineTransformAt tests.
  
FUTURE WORK:
- There is discussion about introducing a convenient way to get Points from polyline uniformly distributed by length, since PointAtNormalized is not doing the job.

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1035)


ikeough:
Thanks @DmytroMuravskyi for continuing to dig us out after the curve changes :)

DmytroMuravskyi:
Updated branch and solved conflicts

andrewheumann (approved):
> Reviewed 5 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1035)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ikeough)

PR #1042, opened 2023-10-24
Fix: GeometrisElement.Intersection method

katehryhorenko:
BACKGROUND:
I changed the GeometricElement.Intersection method to support RepresentationInstances. I playing with Stocking Plan Exporter and noticed that sometimes pdf contains wrong geometry:
image

DESCRIPTION:
- The issue was that I have changed prev. this method to collect intersection points and only after that add them to the graph. And for the case below it doesn't work:
image

I changed method to add points like it was before

TESTING:
image

REQUIRED:

COMMENTS:
- Change log contains data that Intersection method was changed

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1042)

### DIFF
--- a/Elements/src/GeometricElement.cs
+++ b/Elements/src/GeometricElement.cs
@@ -213,7 +213,6 @@ namespace Elements
             var graphVertices = new List<Vector3>();
             var graphEdges = new List<List<(int from, int to, int? tag)>>();
 
-            var intersectionPoints = new List<Vector3>();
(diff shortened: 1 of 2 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1042)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1049, opened 2023-10-30
Added optional snapping points generation;

katehryhorenko:
BACKGROUND:
ElementInstancesare not serialized as part of the JSON. Explore uses Representation geometry to create snapping points and now it doesn't have data to do it.

DESCRIPTION:
-  CreateSnappingPoints method is added to the ElementRepresentation
- SnappingPointsSource  property is added to the RepresentationInstance. By default this value is None => no additional geometry will be created
-  Snapping points are saved as HYPAR_snapping_points node extension inside glb. But now I'm saving it to each representation node. But there should be a way to keep it at some place (like mesh) and re-use it.
- related to [this PR](https://github.com/hypar-io/Explore/pull/3863)

TESTING:
- I tested it with Unicorn function and with DebugGeometry flag on to be sure that Explore uses only data from the visible representations
  
FUTURE WORK:
- Add SnappingPointsSource.Glb => Explore will generate snapping points from the Oblect3D edges
- Mode SnappingPoints user data to Mesh Node (or some other reusable place)

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1049)


katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1049)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @andrewheumann)

___
*[`Elements/src/Serialization/JSON/VectorListToByteArrayConverter.cs` line 17 at r3](https://reviewable.io/reviews/hypar-io/Elements/1049#-NiF-Y5h1TJHuOkeewzk:-Ni_zreiArwRiv9gK2uz:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/50d65358649ccd3e7d48133d9c20beaef15e2459/Elements/src/Serialization/JSON/VectorListToByteArrayConverter.cs#L17)):*

I think for safety's sake we should also include  `public override bool CanRead => false;` so we never attempt to call `ReadJson` accidentally.
</blockquote></details>

Done.

PR #1054, opened 2023-11-08
path to generate catalog from catalog obj not uri

wynged:
BACKGROUND:
- We would like to allow other forms of content catalogs to exist, which requires code gen to be more flexible, allowing a catalog cs file to be generated from a Catalog created in some other way rather than just a public URI.

DESCRIPTION:
- factor out a new method for use while generating catalog code files.

TESTING:
- Will be used in other contexts, these tests should remain unchanged.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1054)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1054)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1055, opened 2023-11-10
don't process solid if no csg

wynged:
BACKGROUND:
- I wasn't getting any stuff when making elbows because I had a non-null representation that was empty. this caused errors during the code for getting render data for element.  we previously wouldn't notice these errors because we wouldn't expect to see any geometry, but now with the alternative path of RepresentationInstances we might expect to see geometry in these cases.

DESCRIPTION:
- add a _csg==null check

TESTING:
- no behavior change.
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1055)

### DIFF
--- a/Elements/src/Serialization/glTF/GltfExtensions.cs
+++ b/Elements/src/Serialization/glTF/GltfExtensions.cs
@@ -1623,7 +1623,7 @@ namespace Elements.Serialization.glTF
                 return -1;
             }
 
-            if (geometricElement.Representation != null)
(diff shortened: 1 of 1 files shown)


katehryhorenko (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1055)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1058, opened 2023-11-15
Discriminator warnings and Element concrete class

anthonie-kramer:
BACKGROUND:
- Actually throwing errors during deserialization can make it very hard to debug an issue if the is a deserialization issue, even if that deserialization issue can be safely ignored. We eventually handle these exceptions anyways and surface messages, so this simply does that explicitly.
- This really only affects working with local Elements, but this is frequent enough to be annoying.
- Doesn't allow null values when deserializing

DESCRIPTION:
- Create a _deserializationDiscriminatorWarnings list that saves a list of deserialization issues encountered, rather than throwing for each one encountered.
- Makes `Element` a concrete class so we can instantiate it instead of returning null

TESTING:
- Updated tests in the ModelTests to turn unknown types into the base class of `Element` with `Name` and `Id` assigned and all `AdditionalProperties` are assigned.
  
FUTURE WORK:
- We could do things now when we don't have the right `Element` type for deserialization... we can still get at the properties of the base `Element` through `AdditionalProperties`

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)


andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer and @wynged)

___
*[`Elements/src/Serialization/JSON/JsonInheritanceConverter.cs` line 317 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNe7oq1wG3c4X_sePI:-NjNe7oq1wG3c4X_sePJ:baic7s6) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L317)):*
> ```Smalltalk
>                     Element element = new Element();
> 
>                     foreach (var prop in jObject.Properties())
> ```

I think we don't need to do this. See how the `GeometricElement` fallback was implemented here: 
[https://github.com/hypar-io/Elements/pull/369](https://github.com/hypar-io/Elements/pull/369/files#diff-85226d4946bc80d99014002729083e177c63523aac2be2667efc264b47f10abfR239)

I think we can add another clause there for falling back to `Element` which should work, without having to construct a new one ourselves. AdditionalProperties should be handled automatically. I think it's in those cases (where a fallback was used for a type) that we should add deserialization warnings.

andrewheumann (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer and @wynged)

___
*[`Elements/src/Serialization/JSON/JsonInheritanceConverter.cs` line 317 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNe7oq1wG3c4X_sePI:-NjNewtT24FGHDSZ8g_c:b-v9haf7) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L317)):*

I think we don't need to do this. See how the `GeometricElement` fallback was implemented here: 
[https://github.com/hypar-io/Elements/pull/369](https://github.com/hypar-io/Elements/pull/369/files#diff-85226d4946bc80d99014002729083e177c63523aac2be2667efc264b47f10abfR239)

I think we can add another clause there for falling back to `Element` which should work, without having to construct a new one ourselves. AdditionalProperties should be handled automatically. I think it's in those cases (where a fallback was used for a type) that we should add deserialization warnings.
</blockquote></details>

something like

```cs
// If it's not in the type cache see if it's got a representation.
// Import it as a GeometricElement.
// TODO: this may not be a safe assumption now that we don't always serialize representations. 
if (jObject.TryGetValue("Representation", out _))
{
    return typeof(GeometricElement);
}
// If nothing else has worked, see if it has an ID and treat it as a generic element
if (jObject.TryGetValue("Id", out _))
{
    return typeof(Element);
}

```

wynged (changes requested):
> Reviewed 1 of 4 files at r1, 5 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @anthonie-kramer)

___
*[`Elements/src/Model.cs` line 449 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfedt-urU_5pUsxEt:-NjNfedt-urU_5pUsxEu:b-4swtiu) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Model.cs#L449)):*
> ```Smalltalk
>             JsonInheritanceConverter.Elements.Clear();
>             // Don't leave null elements in the model.
>             var goodElements = new Dictionary<Guid, Element>(model.Elements.Count());
> ```

I think that with this new strategy we won't end up with any nulls right?  so maybe we should remove this bit of code.
___
*[`Elements/test/ModelTests.cs` line 88 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfOYH3mgt-bMDGi3P:-NjNfOYH3mgt-bMDGi3Q:b-iia59b) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/test/ModelTests.cs#L88)):*
> ```Smalltalk
>             // and one base element (Elements.Baz)
>             Assert.Equal(4, model.Elements.Count);
>             Assert.Single(errors);
> ```

can we add a check that the thing that was Elements.Bas has it's discriminator saved as `Elements.Baz` in the additional properties?

anthonie-kramer (commented):
Kind of melding both approaches from @wynged 's original branch, and @andrewheumann 's approach to creating the base `Element`... I think we still want to return null on discriminator errors because of the thrown exception and debugging frustration...

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058#-:-NjO28QR-78kTwBh50sp:b-wkvq20)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[`Elements/src/Model.cs` line 449 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfedt-urU_5pUsxEt:-NjNtpek8GRMNua0mADy:bo5z49) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Model.cs#L449)):*

I think that with this new strategy we won't end up with any nulls right?  so maybe we should remove this bit of code.
</blockquote></details>

I would like to keep this in since we will still be potentially returning `null` when a discriminator exception gets thrown (see updates from Andrew's comment)
___
*[`Elements/src/Serialization/JSON/JsonInheritanceConverter.cs` line 317 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNe7oq1wG3c4X_sePI:-NjNzFzq3E3S9piAjwfU:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs#L317)):*

something like

```cs
// If it's not in the type cache see if it's got a representation.
// Import it as a GeometricElement.
// TODO: this may not be a safe assumption now that we don't always serialize representations. 
if (jObject.TryGetValue("Representation", out _))
{
    return typeof(GeometricElement);
}
// If nothing else has worked, see if it has an ID and treat it as a generic element
if (jObject.TryGetValue("Id", out _))
{
    return typeof(Element);
}

```
</blockquote></details>

Done.
___
*[`Elements/test/ModelTests.cs` line 88 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfOYH3mgt-bMDGi3P:-NjNzDyS9l81rkf9uTuh:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/test/ModelTests.cs#L88)):*

can we add a check that the thing that was Elements.Bas has it's discriminator saved as `Elements.Baz` in the additional properties?
</blockquote></details>

Done.

wynged (changes requested):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @anthonie-kramer)

___
*[`Elements/src/Model.cs` line 449 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfedt-urU_5pUsxEt:-NjOH9FEDBcEwDhwny6p:bv9nlq4) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Model.cs#L449)):*

I would like to keep this in since we will still be potentially returning `null` when a discriminator exception gets thrown (see updates from Andrew's comment)
</blockquote></details>

but I think what you've done has basically made it impossible for a discriminator exception to be thrown except it truly unusual circumstances so this seems just like redundant code executing that we shouldn't need.

anthonie-kramer (commented):
Added a test and updated code to keep the discriminator when the object falls back to a base `GeometricElement` or `Element`

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058#-:-NjPQJygAj34TPx8oz1w:b-ain1pm)* status: 2 change requests, 0 of 1 approvals obtained (waiting on @andrewheumann and @wynged)

___
*[`Elements/src/Model.cs` line 449 at r2](https://reviewable.io/reviews/hypar-io/Elements/1058#-NjNfedt-urU_5pUsxEt:-NjPQ7lJ44j-wa0v7UTB:b-eqqwfs) ([raw file](https://github.com/hypar-io/Elements/blob/2f7a371abb461254ccf92306f1be45a5df7e4f11/Elements/src/Model.cs#L449)):*

but I think what you've done has basically made it impossible for a discriminator exception to be thrown except it truly unusual circumstances so this seems just like redundant code executing that we shouldn't need.
</blockquote></details>

Removed per conversation earlier. We want to know if nulls are assigned.

andrewheumann (approved):
LGTM, tho I haven't tested myself! 

> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058#-:-NjSSFYPDg0dH0ezFnKM:bi6ykog)* status: :shipit: complete! 1 change requests, 1 of 1 approvals obtained (waiting on @wynged)

wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1058)* status: :shipit: complete! 2 of 1 approvals obtained

PR #1062, opened 2023-11-21
fix mesh / topo intersection

andrewheumann:
BACKGROUND:
- Mesh/Ray and Topo/Ray intersection would return the wrong point for closed volumes — it was only returning the first intersection, rather than the nearest, so it was impossible to (e.g.) project a ray down onto a topography and get the location of the intersection. 
- The `RayIntersectTopo` test got blown away by [this PR](https://github.com/hypar-io/Elements/pull/212) with what appears to be some temporary code — it switched from testing `Ray.Intersects(Topography)` to testing `Ray.Intersects(Triangle)`.
image

DESCRIPTION:
- Checks all near triangles, rather than aborting after the first one.
- Fixes the `RayIntersectsTopo` test

REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1062)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 - `Polyline.TransformAt` returns correct transformations when parameter on domain is provided.
 - `IndexedPolycurve` constructor that takes list of `BoundedCurve` now produces `CurveIndices` that share vertices and are withing index range. This means `IndexedPolyline.TransformedPolyline` preserves `CurveIndicies` on new `IndexedPolyline`.
 - `BoundedCurve.ToPolyline` now works correctly for `EllipticalArc` class.
+- `Ray.Intersects(Topography)` and `Ray.Intersects(Mesh)` would sometimes return a different intersection than the closest one.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


jamesbradleym (approved):
Looks good! It probably makes sense for the RayIntersectsTopography test to use ray.Intersects(topo, out var result) rather than ray.Intersects(topo.Mesh, out var result)

One question I have regarding the Intersects method that uses a Topo. Presumably if I'm looking for the intersection of a topo (and MeshElements with a Transform in general) I'd expect that intersection point to be where the actual geometry is rather than where the untransformed mesh is.

I'm not sure it makes sense necessarily to have this method on topo if it takes doesn't adjust to the topo.Transform.Matrix.Translation?

Approving incase it's expected to receive the untransformed intersection but this seems unexpected to me.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1062#-:-NjrNkyjBOQmKmvMO694:b7l0t2q)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1070, opened 2023-12-05
`AdaptiveGrid.AddEdge` no longer delete edges in case where new edge partially overlaps old edges without intersections.

DmytroMuravskyi:
BACKGROUND:
- When creating grid from a set of circulation objects I found that some edges disappear.
- After investigation I found a bug in partial overlap of two edges: when one end is outside and one is inside.  

DESCRIPTION:
- Register missing intersection points in `AddCutEdge`.  
- It's not an issue if some of them will appear in the list several times since the list is later ordered and checked for duplicates.

TESTING:
- Added AdaptiveGridAddCutWithOverlap to cover problematic case.
 
REQUIRED:

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1070)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1070)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1073, opened 2023-12-12
handle folders which look like elemetns but aren't

wynged:
BACKGROUND:
- De-serializing folders in tests broke recently because we started falling back to elements if something has an "Id" property, which folders do, but they are not Elements

DESCRIPTION:
- Check if there is a discriminator before falling back to Element or GeometricElement.  non-elements can't fallback to these types even if they have an Id or Representation property.

TESTING:
- try to make/run a function that as a folder as an input.  You will get a de serialization error on de serializing the input, then try with this change.
  
FUTURE WORK:
- NA

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1073)

### DIFF
--- a/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs
+++ b/Elements/src/Serialization/JSON/JsonInheritanceConverter.cs
@@ -358,7 +358,7 @@ namespace Elements.Serialization.JSON
                 return typeof(GeometricElement);
             }
             // If nothing else has worked, see if it has an ID and treat it as a generic element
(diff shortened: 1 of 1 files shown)


anthonie-kramer (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1073)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1078, opened 2024-03-04
remove all instances of Required.AllowNull

andrewheumann:
BACKGROUND:
- In a number of cases, our NJsonSchema-generated classes reflected an unintentional strange state: that of being `required` but possibly `null`. This is rarely practically useful: in nearly all cases we want a property to either be required or to permit null, but not both.  

DESCRIPTION:
- Adapts all usage of the `Required.AllowNull` attribute to be either `Required.Always`, or not required. Since these values could always be null anyway, the code is generally able to tolerate their "nullness" — the difference is just whether or not we get a serialization failure if a `null` is omitted from the json. For example:
```
{
   Perimeter: [...],
   Voids: null
}
``` 
is a valid `Profile`, but 

```
{
   Perimeter: [...],
}
```
will throw a deserialization error, even though the code can handle it. 

TESTING:
- All tests continue to pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1078)

### DIFF
--- a/Elements/src/CoreModels/RepresentationInstance.cs
+++ b/Elements/src/CoreModels/RepresentationInstance.cs
@@ -43,7 +43,7 @@ namespace Elements
         /// <summary>
         /// The representation's material.
         /// </summary>
-        [JsonProperty("Material", Required = Required.AllowNull)]
+        [JsonProperty("Material")]
         public Material Material { get; set; }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


ikeough (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1078#-:-NsAIVWvDu29c7DsWzbD:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1079, opened 2024-03-06
Fix setting a different catalog

andrewheumann:
BACKGROUND:
- Ancient catalog retrieval code assumed only a single catalog could exist. Setting a second catalog (as the Space Type function does) would fail — we would see we already had a catalog, and just use that, rather than reading the new one.
- This meant that in cases where more than one space type was specified, all but the first one would be missing all properties. 

DESCRIPTION:
- Resets the catalog to null when setting a new path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1079)

### DIFF
--- a/Elements.Components/src/ContentCatalogRetrieval.cs
+++ b/Elements.Components/src/ContentCatalogRetrieval.cs
@@ -25,6 +25,8 @@ namespace Elements.Components
         public static void SetCatalogFilePath(string path)
         {
             catalogFilePath = path;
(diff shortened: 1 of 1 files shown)


katehryhorenko (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1079#-:-NsIqTlx1gdNwoIWRXfD:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1080, opened 2024-03-13
reset space configuration cache when setting catalog path

wynged:
BACKGROUND:
- We noticed that when a space type was updated in the content catalog that old values for the instance parameters on ContentElements would persists across multiple executions.  

DESCRIPTION:
- When a catalog path is set we want to be sure that the `SpaceConfiguration` class will use that new catalog, so we have to clear the catalog that is used for ContentElement retrieval inside `SpaceConfiguration`

TESTING:
- I've tested this with the `function-SpaceType` function and this resolves the stale instance parameter values
  
FUTURE WORK:

REQUIRED:
~- [ ] All changes are up to date in `CHANGELOG.md`.~

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1080)

### DIFF
--- a/Elements.Components/src/ContentCatalogRetrieval.cs
+++ b/Elements.Components/src/ContentCatalogRetrieval.cs
@@ -18,7 +18,7 @@ namespace Elements.Components
         private static string catalogFilePath = null;
 
         /// <summary>
-        /// If you're using a different location for the catalog file, you can 
+        /// If you're using a different location for the catalog file, you can
         /// set it here.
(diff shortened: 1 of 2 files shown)


andrewheumann (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1080)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

Issue #1083, opened 2024-04-16
Strange behaviour with rendering Model (in dotnet.interactive)


daniel-fink:
I am in the early stages of writing a lightweight Elements <-> Speckle converter. At the moment just concentrating on geometry.
To test, I am visualizing results in Polyglot Notebooks (dotnet.interactive)

I am getting some strange behaviour when adding sequential `Element`s to a `Model` and visualizing them (see screenshot attached):

1. I can retrieve the Speckle `Brep` (note the volume & faces)
2. I've written a `Brep` -> `Solid` converter, that somehow gets the faces right but the volume wrong.
3. To see where things might be going wrong, I attempt to visualise the `Solid`, but I note that on trying to sequentially add each face as a `Panel` to a `Model`, while I can render the 4th `Panel` individually, on adding it to the `Model` the rendering fails. 
4. No additional errors are found when searching through console logs in browsers...

I would expect to be able to sequentially add `Panel`s to the `Model` and view all of them.

Running on macOS M1 Sonoma 14.4.1

![image](https://github.com/hypar-io/Elements/assets/32397975/5622937d-91d1-4e29-8262-8a5c99f224fc)

andrewheumann:
Without your speckle geometry this will be hard to debug. Is there a way you can provide a .dib that reproduces this without depending on your speckle account, perhaps by serializing the raw polygons you're getting out of speckle?

daniel-fink:
Thanks @andrewheumann. I'm not sure how to best serialize data in a Polyglot Notebook, but in any case I figured it would be quicker to point you to a repo with the current (very very draft) code -- https://github.com/daniel-fink/speckle-hypar
.dib file is in the /tests dir ('/tests/ConversionTests.dib)

In order to run it, you would need to:
1. Build the SpeckleHypar.Conversion project
2. Point the magic #i nuget source locator to the relative dir on your machine.
3. Replace the argument in the `SpeckleApi()` method with your own Speckle API token.

I believe that should work but happy to coordinate.

daniel-fink:
Hi @andrewheumann, just bumping this. Let me know if you were able to witness the issue, or I should look harder at providing an independent .dib file...

> Thanks @andrewheumann. I'm not sure how to best serialize data in a Polyglot Notebook, but in any case I figured it would be quicker to point you to a repo with the current (very very draft) code -- https://github.com/daniel-fink/speckle-hypar .dib file is in the /tests dir ('/tests/ConversionTests.dib)
> 
[...]

PR #1084, opened 2024-04-28
fix some bugs in the notebook viewer

andrewheumann:
BACKGROUND:
- Fixes https://github.com/hypar-io/Elements/issues/1083

DESCRIPTION:
- There were a couple minor bugs in the 3d viewer code which resulted in the far plane calculation being wrong, which would cause models not centered on the origin to be clipped / invisible. There were also a couple other little syntax issues.

TESTING:
- I loaded this model into a notebook and returned it in a cell. (Scratch.dib is a good place to test this out against the local version of extension.dib)
[bad-model.json](https://github.com/hypar-io/Elements/files/15142356/bad-model.json)

image
  
FUTURE WORK:
- We should find a way to upgrade the version of threejs in our notebook, so we can support setting DEFAULT_UP to `z` instead of `y`, but the recommended CDN approach for three these days requires `importMaps`, which I couldn't get to work in a notebook context.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1084)

### DIFF
--- a/Elements/src/extension.dib
+++ b/Elements/src/extension.dib
@@ -92,10 +92,9 @@ const fitCameraToObject = function (scene, offset = 1.25) {
         camera.position.copy(center.clone().add(size.clone().multiplyScalar(offset)))
     }
 
-    const minZ = boundingBox.min.y;
-    const cameraToFarEdge = (minZ < 0) ? -minZ + cameraZ : cameraZ - minZ;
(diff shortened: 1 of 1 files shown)


wynged (approved):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1084)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1089, opened 2024-12-24
Use representation instances for reducers

jamesbradleym:
BACKGROUND:
- Reducers were not being rendered properly do to (presumably) geometry conflicts from 2 sweeps intersecting.

DESCRIPTION:
- Switches from using `this.Representation` to `this.RepresentationInstances` for Reducer representations.

TESTING:
- Load up a workflow with reducers of various sizes and see them no longer missing.
  
Screenshot 2024-12-24 at 4 40 12 PM

Screenshot 2024-12-24 at 4 39 38 PM

FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)

### DIFF
--- a/Elements.MEP/src/Fittings/FittingRepresentationStorage.cs
+++ b/Elements.MEP/src/Fittings/FittingRepresentationStorage.cs
@@ -11,17 +11,32 @@ namespace Elements.Fittings
         private static readonly Dictionary<string, List<RepresentationInstance>> _fittings = new Dictionary<string, List<RepresentationInstance>>();
         public static Dictionary<string, List<RepresentationInstance>> Fittings => _fittings;
 
-        public static void SetFittingRepresentation(Fitting fitting, Func<IList<SolidOperation>> makeSolids)
+        public static void SetFittingRepresentation(Fitting fitting, Func<IList<SolidOperation>> makeSolids, Boolean unioned = true, Boolean updateTransform = true)
         {
             var hash = fitting.GetRepresentationHash();
             if (!_fittings.ContainsKey(hash))
             {
                 var solids = makeSolids();
-                _fittings.Add(hash, new List<RepresentationInstance> { new RepresentationInstance(new SolidRepresentation(solids), fitting.Material) });
+                var representationInstances = new List<RepresentationInstance>();
+                if (unioned)
+                {
+                    representationInstances = new List<RepresentationInstance> { new RepresentationInstance(new SolidRepresentation(solids), fitting.Material) };
+                }
+                else
+                {
+                    foreach (var solid in solids)
+                    {
+                        representationInstances.Add(new RepresentationInstance(new SolidRepresentation(solid), fitting.Material));
+                    }
+                }
+                _fittings.Add(hash, representationInstances);
             }
             fitting.RepresentationInstances = _fittings[hash];
 
-            fitting.Transform = fitting.GetRotatedTransform().Concatenated(new Transform(fitting.Transform.Origin));
+            if (updateTransform)
+            {
+                fitting.Transform = fitting.GetRotatedTransform().Concatenated(new Transform(fitting.Transform.Origin));
+            }
         }
     }
 }
--- a/Elements.MEP/src/Fittings/FittingRepresentationStorage.cs
+++ b/Elements.MEP/src/Fittings/FittingRepresentationStorage.cs
@@ -11,32 +11,17 @@ namespace Elements.Fittings
         private static readonly Dictionary<string, List<RepresentationInstance>> _fittings = new Dictionary<string, List<RepresentationInstance>>();
         public static Dictionary<string, List<RepresentationInstance>> Fittings => _fittings;
 
-        public static void SetFittingRepresentation(Fitting fitting, Func<IList<SolidOperation>> makeSolids, Boolean unioned = true, Boolean updateTransform = true)
+        public static void SetFittingRepresentation(Fitting fitting, Func<IList<SolidOperation>> makeSolids)
         {
             var hash = fitting.GetRepresentationHash();
             if (!_fittings.ContainsKey(hash))
             {
                 var solids = makeSolids();
-                var representationInstances = new List<RepresentationInstance>();
-                if (unioned)
-                {
-                    representationInstances = new List<RepresentationInstance> { new RepresentationInstance(new SolidRepresentation(solids), fitting.Material) };
-                }
-                else
-                {
-                    foreach (var solid in solids)
-                    {
-                        representationInstances.Add(new RepresentationInstance(new SolidRepresentation(solid), fitting.Material));
-                    }
-                }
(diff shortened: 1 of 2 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @jamesbradleym)

___
*[`Elements.MEP/src/Fittings/Reducer.cs` line 75 at r1](https://reviewable.io/reviews/hypar-io/Elements/1089#-OF2Gn8F1TKHLNq-mzEg:-OF2Gn8F1TKHLNq-mzEh:blvxv4i) ([raw file](https://github.com/hypar-io/Elements/blob/0620af28fa9b139f07ada0d3cee62210a57648d9/Elements.MEP/src/Fittings/Reducer.cs#L75)):*
> ```Smalltalk
>                  .Concat(this.End.GetArrow(endNodeTransform.Origin)).Concat(GetExtensions());
> 
>             this.RepresentationInstances.Add(new RepresentationInstance(new SolidRepresentation(sweep1), this.Material));
> ```

This is not really the pattern we want when we use representation instances.  The goal of representation instances was to make it so we re-use representation, but with how you have done it here we're making new instances every time.  See the `UpdateRepresentation` method in  Elbow.cs 

However, if I recall correctly getting reducer representations to appear correctly was tricky because they both have a direction, large->small vs small->large, and they can be eccentric, which means their transform orientation is that much trickier.  

Do we have a theory about why this use of reprsentation instances works when old style representations doesn't work?

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements.MEP/src/Fittings/Reducer.cs` line 75 at r1](https://reviewable.io/reviews/hypar-io/Elements/1089#-OF2Gn8F1TKHLNq-mzEg:-OF2mzeXBNFnMePsR0Yi:betdrke) ([raw file](https://github.com/hypar-io/Elements/blob/0620af28fa9b139f07ada0d3cee62210a57648d9/Elements.MEP/src/Fittings/Reducer.cs#L75)):*

This is not really the pattern we want when we use representation instances.  The goal of representation instances was to make it so we re-use representation, but with how you have done it here we're making new instances every time.  See the `UpdateRepresentation` method in  Elbow.cs 

However, if I recall correctly getting reducer representations to appear correctly was tricky because they both have a direction, large->small vs small->large, and they can be eccentric, which means their transform orientation is that much trickier.  

Do we have a theory about why this use of reprsentation instances works when old style representations doesn't work?
</blockquote></details>

Using `FittingRepresentationStorage.SetFittingRepresentation` results in bad geometry transforms it looks like, is that the right implementation? I had taken this pattern from https://github.com/hypar-io/AECTech2023/blob/0e6db42d4f20bc85a03d565cdd16a7a1e726a063/ExampleFunctions/Precincts/dependencies/Site.cs#L15-L22 which I thought was on the newer side but maybe that wasn't the right place to look.

wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @jamesbradleym)

___
*[`Elements.MEP/src/Fittings/Reducer.cs` line 75 at r1](https://reviewable.io/reviews/hypar-io/Elements/1089#-OF2Gn8F1TKHLNq-mzEg:-OFNH-r3-Gu8c1u2CWZt:b-mmfvqv) ([raw file](https://github.com/hypar-io/Elements/blob/0620af28fa9b139f07ada0d3cee62210a57648d9/Elements.MEP/src/Fittings/Reducer.cs#L75)):*
<details><summary><i>Previously, jamesbradleym wrote…</i></summary><blockquote>

Using `FittingRepresentationStorage.SetFittingRepresentation` results in bad geometry transforms it looks like, is that the right implementation? I had taken this pattern from https://github.com/hypar-io/AECTech2023/blob/0e6db42d4f20bc85a03d565cdd16a7a1e726a063/ExampleFunctions/Precincts/dependencies/Site.cs#L15-L22 which I thought was on the newer side but maybe that wasn't the right place to look.


</blockquote></details>

hmmm... yea, that rings a bell, other fittings can more easily have a transform that orients them in space, but we never made transforms of reducers make total sense...
ok, it's not important to solve now, let's just add a TODO here that explains that we'd rather use representationInstances with a Factory pattern like other fittings because we want the advantage of instancing geometry, but because reducers have unique orientation requirements we haven't implemented it yet.

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements.MEP/src/Fittings/Reducer.cs` line 75 at r1](https://reviewable.io/reviews/hypar-io/Elements/1089#-OF2Gn8F1TKHLNq-mzEg:-OFSBy939sM66FOY7r_4:b-26myve) ([raw file](https://github.com/hypar-io/Elements/blob/0620af28fa9b139f07ada0d3cee62210a57648d9/Elements.MEP/src/Fittings/Reducer.cs#L75)):*

hmmm... yea, that rings a bell, other fittings can more easily have a transform that orients them in space, but we never made transforms of reducers make total sense...
ok, it's not important to solve now, let's just add a TODO here that explains that we'd rather use representationInstances with a Factory pattern like other fittings because we want the advantage of instancing geometry, but because reducers have unique orientation requirements we haven't implemented it yet.
</blockquote></details>

In that case could we just modify the `SetFittingRepresentation` to work with `Reducer` as well? As implemented: Add a boolean for transforming and a boolean for unioning?

jamesbradleym (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1089)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements.MEP/src/Fittings/Reducer.cs` line 75 at r1](https://reviewable.io/reviews/hypar-io/Elements/1089#-OF2Gn8F1TKHLNq-mzEg:-OFi8HZv5U3ixRjBBzPj:b-896fix) ([raw file](https://github.com/hypar-io/Elements/blob/0620af28fa9b139f07ada0d3cee62210a57648d9/Elements.MEP/src/Fittings/Reducer.cs#L75)):*

oh sorry, since I don't see a representation hash method being implemented, which means we're not actually re-using the representations from the storage, I think we should _just_ leave the TODO.  as it is now we're sort of looking like we're using the storage but we're not really, representations won't ever be re-used, unless i'm mis-reading something.
</blockquote></details>

Done.

PR #1095, opened 2025-05-22
use ubuntu latest

andrewheumann:
our build-bump-alpha-publish action failed. IDK if this will fix it, but all our other actions use ubuntu-latest


andrewheumann:
dupe

PR #1098, opened 2025-05-28
Extending additional transform outside the assembly (in this way we c…

andrzej-rudzki:
…an use reducers in assemblies)

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1098)

### DIFF
--- a/Elements.MEP/src/Fittings/Assembly.cs
+++ b/Elements.MEP/src/Fittings/Assembly.cs
@@ -331,5 +331,16 @@ namespace Elements.Fittings
         {
             throw new NotImplementedException();
         }
+
+        public Transform? GetBranchTransformForConnector(Port connector)
+        {
+            var internalFitting = InternalFittings.FirstOrDefault(f => f.BranchSidePorts().Any(p => p == connector));
+            if (internalFitting?.AdditionalTransform is Transform transform && !new Transform().Equals(transform))
+            {
+                return transform;
+            }
+            return null;
+        }
+
     }
 }
--- a/Elements.MEP/src/Fittings/FittingTree.Builder.cs
+++ b/Elements.MEP/src/Fittings/FittingTree.Builder.cs
@@ -272,6 +272,15 @@ namespace Elements.Fittings
                     var branchSideMatch = OppositeSidePort(connector, branchFitting, out bool hasEnoughSpace);
                     if (branchSideMatch != null && !pipedPorts.Contains(branchSideMatch))
                     {
+                        // Extending additional transform outside the assembly
+                        if (connection is Assembly)
+                        {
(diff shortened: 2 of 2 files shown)


jamesbradleym (changes requested):
Thanks for the PR @andrzej-rudzki - would you mind providing a bit more context regarding the use case for this? It would also be great to get a test for this behavior as well.

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1098#-:-ORMEu-XE7fi85FfTd-m:b-qfaxg6)* status: 1 change requests, 0 of 1 approvals obtained

___
*[`Elements.MEP/src/Fittings/FittingTree.Builder.cs` line 283 at r1](https://reviewable.io/reviews/hypar-io/Elements/1098#-ORMFA594Qeumc3tNUp0:-ORMFA594Qeumc3tNUp1:b-y5y2p3) ([raw file](https://github.com/hypar-io/Elements/blob/6266b1a55d9e116c215d492bafcf0ddf28b5c780/Elements.MEP/src/Fittings/FittingTree.Builder.cs#L283)):*
> ```Smalltalk
>                                 branch.PropagateAdditionalTransform(internalFitting.AdditionalTransform, TransformDirection.TrunkToBranch);
>                             }
>                         }
> ```

This might be useful as a method in `Assembly.cs`, something like:

```
public Transform? GetBranchTransformForConnector(Port connector)
{
    var internalFitting = InternalFittings.FirstOrDefault(f => f.BranchSidePorts().Any(p => p == connector));
    if (internalFitting != null && !internalFitting.AdditionalTransform.Equals(new Transform()))
    {
        return internalFitting.AdditionalTransform;
    }
    return null;
}
```

Then we can call that here with something like:

```
if (connection is Assembly assembly)
{
    var transform = assembly.GetBranchTransformForConnector(connector);
    if (transform != null)
    {
        branch.PropagateAdditionalTransform(transform.Value, TransformDirection.TrunkToBranch);
    }
}
```



_Code quote:_
```Smalltalk
                        // Extending additional transform outside the assembly
                        if (connection is Assembly)
                        {
                            var internalFitting = ((Assembly)connection).InternalFittings.FirstOrDefault(f => f.BranchSidePorts().Any(p => p == connector));
                            if (internalFitting != null && !new Transform().Equals(internalFitting.AdditionalTransform))
                            {
                                branch.PropagateAdditionalTransform(internalFitting.AdditionalTransform, TransformDirection.TrunkToBranch);
                            }
                        }
```

andrzej-rudzki:
I think there should be link between connector and corresponding internal Fittings.

jamesbradleym (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1098#-:-ORSWMHq1fXYLXeYMP-9:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1101, opened 2025-06-02
Prevent curve overwrite during GridLine deserialization

katehryhorenko:
#### 🌌 BACKGROUND:
`GridLine` has three properties that can set its geometry: `Curve`, `Geometry`, and `Line`.
In some workflows (e.g., created by a function and later edited in Pringle), a `GridLine` can enter a corrupt state where:

* `Curve` contains the correct value,
* but it is overwritten during deserialization by `Line` or `Geometry` values,
* leading to incorrect behavior or rendering.

#### 📋 DESCRIPTION:
This PR prevents unwanted overwriting of the `Curve` during deserialization:

* Introduces a `deserializationIsInProgress` flag:

  * Set to `true` inside the `JsonConstructor`.
  * Reset to `false` in the `OnDeserialized` callback.
* Modifies the `Line` and `Geometry` setters to **skip assigning to `Curve`** during deserialization, preserving the original `Curve` if it's already set.

#### 🧪 TESTING:
* Enabled the `suggest-everywhere` flag.
* Added a GridLine via suggestion.
* Modified an existing GridLine.
* Copied the GridLine to clipboard.
* Saved and reloaded the file using deserialization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1101)

### DIFF
--- a/Elements/src/GridLine.cs
+++ b/Elements/src/GridLine.cs
@@ -23,14 +23,16 @@ namespace Elements
         /// <summary>
         /// Initializes a new instance of the GridLine class.
         /// </summary>
-        public GridLine()
+        public GridLine() : base()
         {
             deserializationIsInProgress = false;
         }
 
         // TODO: Remove this constructor once we remove the Line and Geometry properties.
         [JsonConstructor]
-        private GridLine(BoundedCurve curve, Polyline geometry, Line line)
+        private GridLine(BoundedCurve curve, Polyline geometry, Line line,
+        Transform @transform = null, Material @material = null, Representation @representation = null, bool @isElementDefinition = false, System.Guid @id = default, string @name = null)
+            : base(@transform, @material, @representation, @isElementDefinition, @id, @name)
         {
             deserializationIsInProgress = true;
             if (curve != null)
--- a/Elements/src/GridLine.cs
+++ b/Elements/src/GridLine.cs
@@ -20,6 +20,14 @@ namespace Elements
         // TODO: Remove this flag once we remove the Line and Geometry properties.
         private bool deserializationIsInProgress = false;
 
+        /// <summary>
+        /// Initializes a new instance of the GridLine class.
+        /// </summary>
+        public GridLine()
+        {
+            deserializationIsInProgress = false;
+        }
+
         // TODO: Remove this constructor once we remove the Line and Geometry properties.
         [JsonConstructor]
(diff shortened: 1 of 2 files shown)


wynged (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1101)* status: 1 change requests, 0 of 1 approvals obtained

___
*[`Elements/src/GridLine.cs` line 33 at r1](https://reviewable.io/reviews/hypar-io/Elements/1101#-ORmOdMA6807rhL_z2O7:-ORmOdMA6807rhL_z2O8:b63uylt) ([raw file](https://github.com/hypar-io/Elements/blob/b5804a276ad95fb81e8bbc36565c037d2624e73a/Elements/src/GridLine.cs#L33)):*
> ```Smalltalk
>         // TODO: Remove this constructor once we remove the Line and Geometry properties.
>         [JsonConstructor]
>         private GridLine(BoundedCurve curve, Polyline geometry, Line line)
> ```

I'm surprised we don't need to call the geometric element constructor from this line as welll, I think most of our other custom constructors call their inherited constructor.  Will the other properties get set correctly still?

katehryhorenko (commented):
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1101)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @wynged)

___
*[`Elements/src/GridLine.cs` line 33 at r1](https://reviewable.io/reviews/hypar-io/Elements/1101#-ORmOdMA6807rhL_z2O7:-ORmQ5iM9pvnzv8792gu:b-2w7l2r) ([raw file](https://github.com/hypar-io/Elements/blob/b5804a276ad95fb81e8bbc36565c037d2624e73a/Elements/src/GridLine.cs#L33)):*

I'm surprised we don't need to call the geometric element constructor from this line as welll, I think most of our other custom constructors call their inherited constructor.  Will the other properties get set correctly still?
</blockquote></details>

It's a good point. I pushed the change

wynged (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1101)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1106, opened 2026-05-21
Fix LibTess tessellation NRE and post-union vertex welding

jamesbradleym:
BACKGROUND:
Same MEP rendering failure as the [Csg PR](https://github.com/hypar-io/Csg/pull/2). Investigation surfaced two distinct Elements-side bugs in the tessellation → GLB pipeline:

- LibTess NRE. When `LibTess` synthesizes `new ContourVertex`s at contour T-junctions or self-intersections (e.g., for a self-intersecting polygon coming from CSG output), those vertices have `Data == null` by default. The pack code (`Tessellation.PackTessellationsIntoBuffers`) unboxed `v.Data` to a 4-tuple shape without null-handling — instant `NullReferenceException`.

- Post-union vertex welding. After a CSG boolean, polygons can carry duplicate `Shared.Tag` values across unrelated faces. The pack's vertex dedup map keyed on (tag, faceId, solidId) would weld vertices at different 3D positions into one index, producing corrupt mesh output (wrong normals, dropped triangles).

Both bugs fire in the same MEP geometry path (multi-solid Union → tessellate → GLB).

DESCRIPTION:
- New `HyparTessCombine` class providing two `CombineCallback` factories — `HyparDataCombine` (4-tuple data shape for CSG/solid-face adapters) and `CsgTexTagCombine` (`(Csg.Vector2D, int)` for the legacy `AddToMesh` path). Each assigns unique synthetic tags via `Interlocked.Increment` from disjoint starting ranges.
- Both `CsgPolygonTessAdapter` and `SolidFaceTessAdapter` now pass `HyparDataCombine` into `tess.Tessellate(...)`, eliminating the null-Data NRE at the source.
`Tessellation.PackTessellationsIntoBuffers` hardened with:
- Pattern-match fallback for missing/malformed Data (synthetic tag, basis-vector UV — defense-in-depth even with callbacks wired).
- Position-matched vertex reuse: on dedup-map collision, only weld when positions actually match; otherwise allocate a fresh synthetic tag and create a new vertex.
- `AddToMesh` (the >4-vertex polygon branch) wires `CsgTexTagCombine` and null-safely unboxes triangle data via a new `TexTagOrDefault` helper.
- New `CsgVertexData` struct replaces the spread `(UV, uint, uint, uint) ValueTuple` across 6+ producer/consumer sites. 
- Elements/lib/Csg.dll rebuilt from the companion Csg commit (https://github.com/hypar-io/Csg/pull/2).
- 7 new tests in CsgTests.cs: 5 proving tests (each fails when its target change is reverted) and 3 canary probes that assert real CSG Union output doesn't produce vertex tag collisions across positions.
- 2 new MEP smoke tests in Elements.MEP/test/SerializationTests.cs — assert Tree and RoofDrain produce non-empty GLB without needing `SkipCSGUnion = true`.

TESTING:
- dotnet test Elements/test/Elements.Tests.csproj --filter "FullyQualifiedName~Csg" — 22 tests pass.
- dotnet test Elements.MEP/test/Elements.MEP.Tests.csproj --filter "FullyQualifiedName~UnionTessellates" — 2 smoke tests pass.
- run any Hypar function that produces an MEP element (Tree or RoofDrain) at survey coordinates (e.g., (−49256, 12000, 5)). Before this PR: GLB output corrupt or missing geometry. After: renders correctly.
- To verify any of the 5 proving tests catches a real regression: revert its target change and re-run the test — it must fail.

Tested with an Aliaxis workflow + function referencing local Elements:
Example workflow: workflows/edcd98fa-06bd-4a55-8e3d-f9ed103f3b1d
Example function: https://github.com/hypar-io/aliaxis/tree/main/ImportedSystemConfiguration (remember to set stream model to false when using hypar run)
  
FUTURE WORK:

REQUIRED:

COMMENTS:
- Any other notes.

BEFORE:
Screenshot 2026-05-21 at 11 30 19 AM
Screenshot 2026-05-21 at 11 30 01 AM

AFTER:
Screenshot 2026-05-21 at 11 30 58 AM
Screenshot 2026-05-21 at 11 30 48 AM

- - -
This change is [Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1106)


serenayl (approved):
> @serenayl reviewed 13 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1106)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on jamesbradleym).

___
*[`Elements/src/Geometry/Tessellation/HyparTessCombine.cs` line 24 at r1](https://reviewable.io/reviews/hypar-io/Elements/1106#-OtB2qaLALt5wKJdLz7h:-OtB2qaLALt5wKJdLz7i:b-ajnt05) ([raw file](https://github.com/hypar-io/Elements/blob/c246ad5a23244437aad291953ab30c9815f977a4/Elements/src/Geometry/Tessellation/HyparTessCombine.cs#L24)):*
> ```Smalltalk
>     /// collide with the small, sequential tags emitted by the CSG library.
>     /// </summary>
>     internal static class HyparTessCombine
> ```

Nit: Should this just be something like CombineCallbacks instead of sounding extra-hypar-specific?



_Code quote:_
```Smalltalk
HyparTessCombine
```
___
*[`Elements/src/Geometry/Tessellation/HyparTessCombine.cs` line 31 at r1](https://reviewable.io/reviews/hypar-io/Elements/1106#-OtB2z5m6FNrp_Vntg2a:-OtB2z5m6FNrp_Vntg2b:b-22h2ju) ([raw file](https://github.com/hypar-io/Elements/blob/c246ad5a23244437aad291953ab30c9815f977a4/Elements/src/Geometry/Tessellation/HyparTessCombine.cs#L31)):*
> ```Smalltalk
>         private static long _syntheticCsgTagCounter = 0x90000000L;
> 
>         internal static CombineCallback HyparDataCombine { get; } = Combine;
> ```

Nit: Does this maybe just want to be `DataCombine`? Should we name the private field it gets more consistently too, if there are two private combine fields?



_Code quote:_
```Smalltalk
HyparDataCombine
```

jamesbradleym (commented):
> @jamesbradleym resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/hypar-io/Elements/1106)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained.

